PR #9, opened 2024-03-12
A RDAS branch re-organized after the current GDASapp using submodules and mpas related component added

TingLei-NOAA:
Following the current GDASapp which had switched to Git submodules to manage components,  re-organize RDAS repo including updating its' module setup according to the current GDASapp.
The mpas-jedi related components are also added while they are not included in the current GDASapp. 
The building of this RDAS bundle had been successfully done on Hera .
More tests are to be done including a complete/clean "clone" and "build" process and on orion.
But this branch could be used as an starting point for current work on RDAS.


CMakeLists.txt
```diff
@@ -34,136 +34,10 @@ set( ENABLE_MPI ON CACHE BOOL "Compile with MPI" )
...
-    endif()
-
-    if(MPAS_DYCORE)
-      ecbuild_bundle( PROJECT mpas-jedi-data  GIT "https://github.com/jcsda-internal/mpas-jedi-data.git"  BRANCH develop UPDATE )
-    endif()
-
-  endif(CLONE_JCSDADATA)
-
-endif(BUILD_RDASBUNDLE)
+option(WORKFLOW_TESTS "Include global-workflow dependent tests" OFF)
```
CoryMartin-NOAA, line 37: global?

build.sh
```diff
@@ -37,12 +36,15 @@ INSTALL_PREFIX=""
 CMAKE_OPTS=""
 BUILD_TARGET="${MACHINE_ID:-'localhost'}"
 BUILD_VERBOSE="NO"
-CLONE_JCSDADATA="NO"
-CLEAN_BUILD="NO"
-BUILD_JCSDA="NO"
-DYCORE="FV3"
-
-while getopts "p:t:c:m:hvdfa" opt; do
+#cltorg CLONE_JCSDADATA="NO"
```
CoryMartin-NOAA, line 39: These commented out/changed make the build script options worthless

build.sh
```diff
@@ -75,19 +74,21 @@ while getopts "p:t:c:m:hvdfa" opt; do
 done
 
 case ${BUILD_TARGET} in
-  hera | orion)
-    echo "Building RDASApp on $BUILD_TARGET"
+  hera | orion | hercules)
+    echo "Building GDASApp on $BUILD_TARGET"
+    export PS1="[\u@\h \W]\$ " #clt to avoid unbound PS1 error
```
CoryMartin-NOAA, line 79: We just need to move away from Conda. Spack-stack has the python modules we need now. PS1 should not be exported in a non-login shell

build.sh
```diff
@@ -75,19 +74,21 @@ while getopts "p:t:c:m:hvdfa" opt; do
 done
 
 case ${BUILD_TARGET} in
-  hera | orion)
-    echo "Building RDASApp on $BUILD_TARGET"
+  hera | orion | hercules)
+    echo "Building GDASApp on $BUILD_TARGET"
```
CoryMartin-NOAA: GDASApp?

build.sh
```diff
@@ -75,19 +74,21 @@ while getopts "p:t:c:m:hvdfa" opt; do
...
+    export PS1="[\u@\h \W]\$ " #clt to avoid unbound PS1 error
     source $dir_root/ush/module-setup.sh
     module use $dir_root/modulefiles
-    module load RDAS/$BUILD_TARGET
-    CMAKE_OPTS+=" -DMPIEXEC_EXECUTABLE=$MPIEXEC_EXEC -DMPIEXEC_NUMPROC_FLAG=$MPIEXEC_NPROC -DBUILD_GSIBEC=ON"
+    module load GDAS/$BUILD_TARGET.$COMPILER
```
CoryMartin-NOAA: GDAS?

build.sh
```diff
@@ -106,30 +107,22 @@ mkdir -p ${BUILD_DIR} && cd ${BUILD_DIR}
...
-else
-  echo "$DYCORE is not a valid dycore option. Valid options are FV3 or MPAS"
-  exit 1
-fi
-
 # JCSDA changed test data things, need to make a dummy CRTM directory
 if [[ $BUILD_TARGET == 'hera' ]]; then
-  mkdir -p $dir_root/test-data-release/
-  ln -sf $RDASAPP_TESTDATA/crtm $dir_root/test-data-release/crtm
+  if [ -d "$dir_root/bundle/fix/test-data-release/" ]; then rm -rf $dir_root/bundle/fix/test-data-release/; fi
```
CoryMartin-NOAA, line 112: Someone else should probably maintain a test data directory

build.sh
```diff
@@ -138,6 +131,7 @@ set -x
 if [[ $BUILD_JCSDA == 'YES' ]]; then
   make -j ${BUILD_JOBS:-6} VERBOSE=$BUILD_VERBOSE
 else
+  builddirs="fv3-jedi soca iodaconv land-imsproc land-jediincr gdas-utils"
```
CoryMartin-NOAA, line 134: do we even need fv3-jedi anymore? what about mpas? and why gdas-utils?

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
+# ------------------------------------------------------------------------- #
+# JEDI GDAS Bundle             #
```
CoryMartin-NOAA, line 2: RDAS?

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+# ------------------------------------------------------------------------- #
+
+# Check for minimim cmake requirement
+cmake_minimum_required( VERSION 3.20 FATAL_ERROR )
+
+find_package(ecbuild 3.5 REQUIRED HINTS ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/../ecbuild)
+
+project(GDAS-bundle VERSION 1.0.0 LANGUAGES C CXX Fortran )
```
CoryMartin-NOAA: RDAS?

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
+
+# Include ecbuild_bundle macro
+include( ecbuild_bundle )
+
+# Enable MPI
+set( ENABLE_MPI ON CACHE BOOL "Compile with MPI" )
+
+# Handle user options.
+option(BUILD_GDASBUNDLE "Build GDAS Bundle" ON)
```
CoryMartin-NOAA: RDAS

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+option(CLONE_JCSDADATA "Clone JCSDA test data repositories" OFF)
+option(WORKFLOW_TESTS "Include global-workflow dependent tests" OFF)
```
CoryMartin-NOAA, line 37: regional

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+
+# Initialize bundle
+# -----------------
+ecbuild_bundle_initialize()
+
+# Build bundle source code.
+if(BUILD_GDASBUNDLE)
```
CoryMartin-NOAA: RDASBUNDLE

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+  ecbuild_bundle( PROJECT atlas GIT "https://github.com/ecmwf/atlas.git" TAG 0.35.0 )
+
+# External (required) observation operators
+  # TODO remove the CRTM from here and use it as a library
+  #option("BUNDLE_SKIP_CRTM" "Don't build CRTM" "OFF") # Don't build crtm unless user passes -DBUNDLE_SKIP_CRTM=OFF
+  #ecbuild_bundle( PROJECT crtm GIT "https://github.com/JCSDA/crtm.git"   TAG v2.4.1-jedi.1 )
+  ecbuild_bundle( PROJECT crtm SOURCE "../sorc/crtm" )
+
+# Build GSI-B
+  option(BUILD_GSIBEC "Build GSI-B" OFF)
```
CoryMartin-NOAA, line 70: Not sure this works for regional apps

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+  if(BUILD_GSIBEC)
+    ecbuild_bundle( PROJECT gsibec SOURCE "../sorc/gsibec" )
+  endif()
+
+# Gibbs seawater
+#clt  ecbuild_bundle( PROJECT gsw SOURCE "../sorc/gsw" )
```
CoryMartin-NOAA, line 76: don't need this, no soca

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+  ecbuild_bundle( PROJECT vader SOURCE "../sorc/vader" )
+  ecbuild_bundle( PROJECT saber SOURCE "../sorc/saber" )
+  option(ENABLE_IODA_DATA "Obtain ioda test data from ioda-data repository (vs tarball)" ON)
+  ecbuild_bundle( PROJECT ioda SOURCE "../sorc/ioda" )
+  option(ENABLE_UFO_DATA "Obtain ufo test data from ufo-data repository (vs tarball)" ON)
+  ecbuild_bundle( PROJECT ufo SOURCE "../sorc/ufo" )
+
+# FMS and FV3 dynamical core
+  ecbuild_bundle( PROJECT fms SOURCE "../sorc/fms" )
+  ecbuild_bundle( PROJECT fv3 SOURCE "../sorc/fv3" )
```
CoryMartin-NOAA, line 89: no need for fv3 anymore ,right?

bundle/CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+  ecbuild_bundle( PROJECT femps SOURCE "../sorc/femps" )
+  ecbuild_bundle( PROJECT fv3-jedi-lm SOURCE "../sorc/fv3-jedi-lm" )
+  option(ENABLE_FV3_JEDI_DATA "Obtain fv3-jedi test data from fv3-jedi-data repository (vs tarball)" ON)
+  ecbuild_bundle( PROJECT fv3-jedi SOURCE "../sorc/fv3-jedi" )
+#build mpas-jedi
+  ecbuild_bundle( PROJECT mpas SOURCE "../sorc/mpas" )
+  ecbuild_bundle( PROJECT mpas-jedi-data SOURCE "../sorc/mpas-jedi-data" )
+  ecbuild_bundle( PROJECT mpas-jedi SOURCE "../sorc/mpas-jedi" )
+
+# SOCA associated repositories
```
CoryMartin-NOAA, line 101: no soca in regional

ush/check_yaml_keys.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python3
```
CoryMartin-NOAA, line 1: we shouldn't use this anymore, global is migrating to `wxflow` YAML tools. Ask @aerorahul for details

ush/convert_yaml_ewok2gdas.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
```
CoryMartin-NOAA: @danholdaway hopes to make this redundant with JCB, I suggest this script is not even needed anyways as of now

CoryMartin-NOAA (changes requested):
It looks like a bunch of things were just copied from GDASApp to here without 1) thinking if they were relevant to the regional system or 2) renaming variables, scripts, etc. 

I think this needs to be cleaned up before it gets merged or else these things will persist.

TingLei-NOAA:
> It looks like a bunch of things were just copied from GDASApp to here without 1) thinking if they were relevant to the regional system or 2) renaming variables, scripts, etc.
> 
> I think this needs to be cleaned up before it gets merged or else these things will persist.

Yes, in this rdas re-organizing, I preferred  keeping/copying components in gdasapp into rdas if they didn't cause much troubles for radas building.   I (or other RDAS developers) would clean them through this PR process and future use/developments of radas.

bundle/CMakeLists.txt
TingLei-NOAA: For being now, I still hope to keep fv3/fv3-jedi still in rdas. At least, it includes rich examples for our jedi work.

build.sh
TingLei-NOAA: This line was added for building on herea, because, otherwise, the building process would abort complain unbound PSI. 
Seems conda was invoked somewhere.  I will see if a better fix could be found

build.sh
TingLei-NOAA: I take them as a simplified version at the beginning. I am still not sure if components like gdas-utils should be included in rdas.

CMakeLists.txt
TingLei-NOAA: I kept them there , which would be a good place holder for future regional workflow if it is decided to be included in rdas.

build.sh
TingLei-NOAA: I think we could still keep fv3-jedi as mentioned in the previous comments.   gdas-utils couldn't be used by rdas at all? 
Currently, as you pointed out previously, this option branch wouldn't not be used.

bundle/CMakeLists.txt
TingLei-NOAA: Yes. seems GSIBEC building failed in my tries.

bundle/CMakeLists.txt
TingLei-NOAA: Currently, soca building failed in my tries and building of them were turned off or commented out. 
But,  I am wondering if they could be included as an optional components when some developments in it are interesting to regional jedi too.  But this might be done without keeping them in rdas.  I would follow the decision made during this reviewing process.

ush/check_yaml_keys.py
TingLei-NOAA: @CoryMartin-NOAA  Thanks a lot for your comments/suggestions in such a prompt way.  
I would dig more on this and talk to @aerorahul on this .

CoryMartin-NOAA:
@TingLei-NOAA I think we (@ShunLiu-NOAA @danholdaway ) should chat soon. GDASApp is containing the GDAS-specific implementation of JEDI, and will be where releases are tagged, and the source code is controlled with hashes/tags. I am nervous about blindly copying things for RDAS, as these two systems will be on a different upgrade cycle, and thus need different modulefiles, different hashes, and because of MPAS vs FV3 and other reasons, different utilities (like blending would not need to be in the global)

ShunLiu-NOAA:
Ting, Thank you for working on this. 
@CoryMartin-NOAA @danholdaway, I have the same concerns about how to keep RDASApp and GDASApp on the same page.

TingLei-NOAA:
> @TingLei-NOAA I think we (@ShunLiu-NOAA @danholdaway ) should chat soon. GDASApp is containing the GDAS-specific implementation of JEDI, and will be where releases are tagged, and the source code is controlled with hashes/tags. I am nervous about blindly copying things for RDAS, as these two systems will be on a different upgrade cycle, and thus need different modulefiles, different hashes, and because of MPAS vs FV3 and other reasons, different utilities (like blending would not need to be in the global)

Sure. a clear plan for how to use rdas is needed.  The current PR is supposed to make rdas could at least work (be able to be built on hera/orion) by, at least, EMC developers for being now. 
To use the current heads of submodules rather than using hashed commits in gdasapp (or rdas), a submodule updating script is added : checkout_submodules_branches.sh.

modulefiles/RDAS/hera.intel.lua
delippi, line 1: @TingLei-NOAA , This is bit-for-bit the same file as the GDAS/hera.intel.lua. Do we need the GDAS directory at all? Also, being bit-for-bit there are mentions to GDAS instead of RDAS. The same goes for the hercules and orion *lua files under the RDAS/ directory.

build.sh
```diff
@@ -24,7 +24,6 @@ usage() {
   echo "  -f  force a clean build             DEFAULT: NO"
   echo "  -d  include JCSDA ctest data        DEFAULT: NO"
   echo "  -a  build everything in bundle      DEFAULT: NO"
-  echo "  -m  select dycore                   DEFAULT: FV3"
```
delippi, line 27: @TingLei-NOAA , why remove this option and not just change to default to MPAS?

modulefiles/RDAS/hera.intel.lua
TingLei-NOAA: @delippi  Yes. now it is just a mirror of GDAS modules.  To have a RDAS dir, it is more readable for like the build.sh. We could make it a link pointing to the GDAS.  However, at the same time, I am wondering if we need the flexibilities to use different setting up modules from GDAS and, hence, use the current practice.

build.sh
TingLei-NOAA: I need to re-sort out on this part of this part.  Thanks.

aerorahul:
> @TingLei-NOAA I think we (@ShunLiu-NOAA @danholdaway ) should chat soon. GDASApp is containing the GDAS-specific implementation of JEDI, and will be where releases are tagged, and the source code is controlled with hashes/tags. I am nervous about blindly copying things for RDAS, as these two systems will be on a different upgrade cycle, and thus need different modulefiles, different hashes, and because of MPAS vs FV3 and other reasons, different utilities (like blending would not need to be in the global)

FWIW, I am with @CoryMartin-NOAA here.  The RDASApp should use GDASApp as a guide and add features that are needed and build on them, rather than import everything and then spend time removing them later.

ShunLiu-NOAA (dismissed):

ShunLiu-NOAA:
@TingLei-NOAA Can we close this PR and open an new PR to merge other changes that you want to do to the current "develop"

TingLei-NOAA:
This PR is closed for https://github.com/NOAA-EMC/RDASApp/pull/11 has been merged into develop branch.
Thanks to all who had contributed to the discussions on this PR.

Issue #14, opened 2024-03-25
The failure to launch ctest jobs on Rocky Hera "srun: error: cli_filter/lua: Unable to stat /apps/slurm/23.11.1/etc/cli_filter.lua: No such file or directory"


TingLei-NOAA:
It was first pointed out by @ShunLiu-NOAA 
After successful building of the current RDAS develop branch 
git log show: 
```
commit 9fc953a15cfa340aa6d43f35213d65e5c1282aee (HEAD -> develop, origin/develop, origin/HEAD)
Author: Cory Martin <cory.r.martin@noaa.gov>
Date:   Thu Mar 21 15:14:35 2024 +0000
```
, first , ran ctest as 
```
module load RDAS/hera.intel
cd ../build
ctest -VV -R fv3jedi_test_tier1_geometry_gfs127 > & ctest-fv3jedi_test_tier1_geometry_gfs127.out &
```
We would see error message in the output as :
```
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 1167
    Start 1167: fv3jedi_test_tier1_geometry_gfs127

1167: Test command: /apps/slurm/default/bin/srun "-n" "6" "/scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-emc-rdas/RDASApp/build/bin/test_fv3jedi_geometry.x" "testinput/geometry_gfs127.yaml"
1167: Environment variables:
1167:  OMP_NUM_THREADS=1
1167: Test timeout computed to be: 1500
1167: srun: error: cli_filter/lua: Unable to stat /apps/slurm/23.11.1/etc/cli_filter.lua: No such file or directory
1167: srun: error: Couldn't load specified plugin name for cli_filter/lua: Plugin init() callback failed
1167: srun: error: cannot create cli_filter context for cli_filter/lua
```
When we have no problems running ctests with jedi-bundle (used jedi's "standard" module setup to build and be loaded before running ctests), we suspect there are some setup issues in the current RDASapp.

delippi:
Hi @TingLei-NOAA and @ShunLiu-NOAA, I think we should not get this error if in modulefiles/RDAS/hera.intel.lua we change:
`local mpiexec = '/apps/slurm/default/bin/srun'` --> `local mpiexec = '/apps/slurm_hera/default/bin/srun'`

TingLei-NOAA:
[...]

PR #15, opened 2024-03-25
fix srun error when running ctests

delippi:
This PR addresses https://github.com/NOAA-EMC/RDASApp/issues/14. We were just pointing to the wrong version of srun.

Tested on Hera and using this before running the ctest: 
`source /scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-jedi-bundle/tlei-jedi-rocky.csh`

### DIFF
--- a/modulefiles/RDAS/hera.intel.lua
+++ b/modulefiles/RDAS/hera.intel.lua
@@ -78,7 +78,7 @@ setenv("CC","mpiicc")
 setenv("FC","mpiifort")
 setenv("CXX","mpiicpc")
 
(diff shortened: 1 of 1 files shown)


ShunLiu-NOAA (approved):

TingLei-NOAA (approved):

TingLei-NOAA:
@delippi  Thanks for this fix.   It seems the loaded module points to the right srun ( by "which srun").  In the future,  that will be desirable to have this variable to be "dynamically " defined.

PR #17, opened 2024-03-29
Add basic on-prem CI to RDASApp

CoryMartin-NOAA:
Adding the capability to run some basic CI on Hera and Orion like what GDASApp can do. 

NOTE: this runs as my user on these machines. We should find someone else to (eventually) take over these crons.

### DIFF
--- a/ci/driver.sh
+++ b/ci/driver.sh
@@ -1,0 +1,107 @@
+#!/bin/bash --login
+
+my_dir="$( cd "$( dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd )"
+
+# ==============================================================================
+usage() {
+  set +x
+  echo
+  echo "Usage: $0 -t <target> -h"
+  echo
+  echo "  -t  target/machine script is running on    DEFAULT: $(hostname)"
+  echo "  -h  display this message and quit"
+  echo
+  exit 1
+}
+
+# ==============================================================================
+# First, set up runtime environment
+
+export TARGET="$(hostname)"
+
+while getopts "t:h" opt; do
+  case $opt in
+    t)
+      TARGET=$OPTARG
+      ;;
+    h|\?|:)
+      usage
+      ;;
+  esac
+done
+
+case ${TARGET} in
+  hera | orion)
+    echo "Running Automated Testing on $TARGET"
+    source $MODULESHOME/init/sh
+    source $my_dir/${TARGET}.sh
+    module purge
+    module use $RDAS_MODULE_USE
+    module load RDAS/$TARGET
+    module list
+    ;;
+  *)
+    echo "Unsupported platform. Exiting with error."
+    exit 1
+    ;;
+esac
+
+# ==============================================================================
+# pull on the repo and get list of open PRs
+cd $RDAS_CI_ROOT/repo
+CI_LABEL="${RDAS_CI_HOST}-RT"
+gh pr list --label "$CI_LABEL" --state "open" | awk '{print $1;}' > $RDAS_CI_ROOT/open_pr_list
+open_pr_list=$(cat $RDAS_CI_ROOT/open_pr_list)
+
+# ==============================================================================
+# clone, checkout, build, test, etc.
+repo_url="https://github.com/NOAA-EMC/RDASApp.git"
+# loop through all open PRs
+for pr in $open_pr_list; do
+  gh pr edit $pr --remove-label $CI_LABEL --add-label ${CI_LABEL}-Running
+  echo "Processing Pull Request #${pr}"
+  mkdir -p $RDAS_CI_ROOT/PR/$pr
+  cd $RDAS_CI_ROOT/PR/$pr
+
+  # clone copy of repo
+  git clone --recursive $repo_url
+  cd RDASApp
+
+  # checkout pull request
+  git pull
+  gh pr checkout $pr
+
+  # get commit hash
+  commit=$(git log --pretty=format:'%h' -n 1)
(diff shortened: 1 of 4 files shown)


CoryMartin-NOAA:
Do not approve/merge this yet, seems that the missing piece is the GitHub CLI that I forgot was needed and is not in the module env

ShunLiu-NOAA (dismissed):

delippi:
Hi @CoryMartin-NOAA, what are the cron entries you are using to run this? I want to give it a try.

CoryMartin-NOAA:
@delippi 
```
*/15 * * * * /scratch1/NCEPDEV/da/Cory.R.Martin/CI/RDASApp/repo/ci/driver.sh -t hera
```

delippi (dismissed):
@CoryMartin-NOAA, Thanks for sharing your cron entry. I was able to get it to run up until the point of
`gh pr list --label "$CI_LABEL" --state "open"`
but it is saying "no pull requests match your search in NOAA-EMC/RDASApp". I'm pretty sure that is because the CI testing has already completed and the labels were updated to hera-RT-Passed. I think everything looks good from what you've shown.

CoryMartin-NOAA:
@delippi thanks, and you are correct!

ci/driver.sh
```diff
@@ -0,0 +1,107 @@
...
+repo_url="https://github.com/NOAA-EMC/RDASApp.git"
+# loop through all open PRs
+for pr in $open_pr_list; do
+  gh pr edit $pr --remove-label $CI_LABEL --add-label ${CI_LABEL}-Running
+  echo "Processing Pull Request #${pr}"
+  mkdir -p $RDAS_CI_ROOT/PR/$pr
+  cd $RDAS_CI_ROOT/PR/$pr
+
+  # clone copy of repo
+  git clone --recursive $repo_url
```
TingLei-NOAA, line 67: @CoryMartin-NOAA  Will use "--recurse-submodules"  better?   This form of git clone I just learned from Donnie , which combines "recursive " and following "git submot init and update"

CoryMartin-NOAA: You don't necessarily want to do that, that won't give you the expected versions/tags of the submodules but rather the newest, right?

ci/driver.sh
```diff
@@ -0,0 +1,107 @@
...
+cd $RDAS_CI_ROOT/repo
+CI_LABEL="${RDAS_CI_HOST}-RT"
+gh pr list --label "$CI_LABEL" --state "open" | awk '{print $1;}' > $RDAS_CI_ROOT/open_pr_list
+open_pr_list=$(cat $RDAS_CI_ROOT/open_pr_list)
+
+# ==============================================================================
+# clone, checkout, build, test, etc.
+repo_url="https://github.com/NOAA-EMC/RDASApp.git"
+# loop through all open PRs
+for pr in $open_pr_list; do
```
TingLei-NOAA, line 60: Are there any controls to let only those PR with new commits  after last CI test (and successful)) to be tested again?

CoryMartin-NOAA: hmm, maybe, just remove the passed label and re-add the queue label is what I've used in the past.

ci/driver.sh
TingLei-NOAA: "Git submodule --update" will get the submodules the same as what is in the  cloned super project (please correct me if I am wrong).  Is this what we expect? 
Otherwise, the submodules are maybe different from what are in the branch/PR we are cloning out.  
So if I understand right, we should always run those submodule updates following a clone .

CoryMartin-NOAA: I think it should be superfluous, you're right, there should be no harm in running it, but I think it shouldn't do anything if it is a fresh clone. Perhaps we should test this to confirm?

TingLei-NOAA: In an extreme hypothetical case,  A PR only includes some changes in the submodules , will a simple "clone" fail to get those changes?

CoryMartin-NOAA: Ting, I think I see your point. But the submodule update line may need to go after the `gh pr checkout` command, to update the submodules for that PR.

TingLei-NOAA (dismissed):

ci/run_ci.sh
```diff
@@ -0,0 +1,79 @@
...
+echo '```' >> $outfile
+echo "Start: $(date) on $(hostname)" >> $outfile
+echo "---------------------------------------------------" >> $outfile
+# ==============================================================================
+# run build script
+cd $repodir
+module purge
+export BUILD_JOBS=8
+rm -rf log.build
+./build.sh -t $TARGET &>> log.build
```
CoryMartin-NOAA: @TingLei-NOAA do we/you want this to build both FV3 and MPAS when it runs?

TingLei-NOAA: @CoryMartin-NOAA  Yes. that is what I think for being  now and I feel, in quite some time in the near future, we may need to work on both like UFO-focused work still using fv3-jedi cases. But if they should be tested on a regular basis, that is a different choice.  Can the CI be enhanced to determine if the PR is related to mpas or fv3 and hence decide what building option -m to be used?   Thanks

CoryMartin-NOAA: We could add an extra label to check which needs to compile. But is there a harm (besides compute time) in always compiling both?

TingLei-NOAA:
I haven't seen the disadvantage building both (except for computing time) . The gdasapp/rdasapp seems to be able to manage both well. For developers , they can also use existing -m option to select what they want.  Maybe, in the future, when more submodules are added depending on one of mpas or fv3 ,  this combined option should be looked at .

TingLei-NOAA (approved):

PR #20, opened 2024-04-03
Add cron to update stable branch

CoryMartin-NOAA:
Using submodules means that `develop` has to be manually updated. But we can create a cron that will ensure that `feature/stable-nightly` has the most recent versions of JEDI repositories that build and tests pass.

To do this, the default branches had to be updated/added for the repositories. This is because the public JCSDA repos have `master` branches as default but we usually need `develop`. Running `git submodule update` on the public repos would have resulted in grabbing very old commits rather than recent ones.

Note: this script will not run properly until develop is updated with those submodule changes in `.gitmodules`. To test, one can add `git checkout feature/update-stable` in the `stable_driver.sh` script right after `git checkout develop` but this was tested manually yesterday on Orion. See: https://github.com/NOAA-EMC/RDASApp/tree/feature/stable-nightly

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,48 +1,64 @@
 [submodule "sorc/crtm"]
 	path = sorc/crtm
 	url = https://github.com/jcsda/crtm
+	branch = release/crtm_jedi_v2.4.1
 [submodule "sorc/femps"]
 	path = sorc/femps
 	url = https://github.com/jcsda/femps
+	branch = develop
 [submodule "sorc/fms"]
 	path = sorc/fms
 	url = https://github.com/jcsda/fms
+	branch = release-stable
 [submodule "sorc/fv3"]
 	path = sorc/fv3
 	url = https://github.com/jcsda/GFDL_atmos_cubed_sphere
+	branch = release-stable
 [submodule "sorc/fv3-jedi"]
 	path = sorc/fv3-jedi
 	url = https://github.com/jcsda/fv3-jedi
+	branch = develop
 [submodule "sorc/fv3-jedi-km"]
 	path = sorc/fv3-jedi-lm
 	url = https://github.com/JCSDA/fv3-jedi-linearmodel
+	branch = develop
 [submodule "sorc/gsibec"]
 	path = sorc/gsibec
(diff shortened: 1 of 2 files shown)


TingLei-NOAA:
@CoryMartin-NOAA   Thanks for taking care of this, which we think is the reason why some ctests failed with the current rdas. 
After the changes in the .gitmodules and run your script to change branches to right branches. 
In the future , unless there are specific requirements for versions of various submodules, we could run  "git submodule update --remote --merge" rather than "git submodule update" to make sure all submodules are updated to their latest develop/targeted branches.

TingLei-NOAA (approved):

.gitmodules
```diff
@@ -1,48 +1,64 @@
...
 	url = https://github.com/jcsda/ioda
+	branch = develop
 [submodule "sorc/iodaconv"]
 	path = sorc/iodaconv
 	url = https://github.com/jcsda-internal/ioda-converters
+	branch = develop
 [submodule "sorc/jedicmake"]
 	path = sorc/jedicmake
 	url = https://github.com/jcsda/jedi-cmake
+	branch = develop
```
TingLei-NOAA, line 40: This one is a little confusing to me. Which is the newer one?  master or develop?

CoryMartin-NOAA: `master` was the default GitHub repo branch name, it has since been replaced by `main`. But the branches that are synced are `develop` so that is the one you should always use.

CoryMartin-NOAA:
> @CoryMartin-NOAA Thanks for taking care of this, which we think is the reason why some ctests failed with the current rdas. After the changes in the .gitmodules and run your script to change branches to right branches. In the future , unless there are specific requirements for versions of various submodules, we could run "git submodule update --remote --merge" rather than "git submodule update" to make sure all submodules are updated to their latest develop/targeted branches.

@TingLei-NOAA yes, in fact this script: https://github.com/NOAA-EMC/RDASApp/blob/develop/ush/submodules/update_develop.sh does exactly that, it wasn't working correctly but it does now.

ShunLiu-NOAA (approved):

PR #21, opened 2024-04-08
meet the data ownership requirement by the mpas-jedi tests with change on mpas-jedi-data to be copied into RDASapp/bundle dir

TingLei-NOAA:
MPAS-jedi apps require write permission to some template files, which are from mpas-jedi-data.  This PR changes the symlink of mpas-jedi-data from a specified machine-dependent location (which might not be owned by the user and cause all mpas-jedi tests fail) to "copy" of that dir to the current RDASapp/bundle dir to resolve this issue. 
With this PR, all mpas-jedi tests passed except for :
```
The following tests FAILED:
    1375 - test_mpasjedi_3denvar_amsua_allsky (Failed)
    1376 - test_mpasjedi_3denvar_amsua_bc (Failed)
    1381 - test_mpasjedi_4denvar_VarBC (Failed)
    1382 - test_mpasjedi_4denvar_VarBC_nonpar (Failed)
```
which were attributed to the use of different crtm data .

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -57,7 +57,7 @@
 [submodule "sorc/mpas"]
 	path = sorc/mpas
 	url = https://github.com/jcsda-internal/MPAS-Model
-	branch = develop
+	branch = release-stable 
 [submodule "sorc/mpas-jedi"]
 	path = sorc/mpas-jedi
 	url = https://github.com/JCSDA/mpas-jedi
--- a/bundle/CMakeLists.txt
+++ b/bundle/CMakeLists.txt
@@ -109,6 +109,7 @@ endif()
 
 # ioda, ufo, fv3-jedi, and saber test data
 #---------------------------------
+  set(myown_JCSDA_DATA_ROOT "/scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-jedi-new/jedi-bundle") #clt
   if(CLONE_JCSDADATA)
 
    set(JCSDA_DATA_ROOT "$ENV{RDASAPP_TESTDATA}/jcsda")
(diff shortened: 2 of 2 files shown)


ShunLiu-NOAA (dismissed):

ShunLiu-NOAA (approved):

delippi:
Hi @TingLei-NOAA, so before mpas-jedi-data/ would have been under build/ and then symlinked to bundle/. What you are proposing is to just put mpas-jedi-data/ directly in bundle/ due to some instance of failures due to file permissions if the user doesn't own the copy of build/mpas-jedi-data/. Is that correct? Shouldn't each user just download their own test data and wouldn't that fix the problem? If the problem is different from that, then I still think we should keep the "build/mpas-jedi-data/ and symlinked to bundle/" configuration for consistency and find a cleaner solution for this particular use case. Perhaps just manually make the copy yourself?

TingLei-NOAA:
@delippi  before, the mpas-jedi-data in bundle dir would be the  symlink of the one in Cory's dir, which we (except for Cory) have no write permission and this PR would copy Cory's mpas-jedi-data to under bundle ( it would still be linked to that under build). 
Yes,  allowing users to clone mpas-jedi-data would also be a solution.  but now, I 'd like to stick to the practice of using local version to reduce the downloading burden.

delippi:
@TingLei-NOAA Okay, I see what the issue is now. I think that seems reasonable. Should we do the same with the other *-data directories?

TingLei-NOAA:
@delippi Seems only mpas-jedi app needs write permission to its "input" data (some template files, i think).  For being now, this PR would allow all ctests using various data  work.

PR #22, opened 2024-04-12
initial version with a rrfs test add

TingLei-NOAA:
Begin creating a rdasapp component to store/help organize rrfs jedi tests.
The structure in rdasapp and strategies for managing rrfs jedi test  could be continually adjusted according to RDASapp developers/users's feedbacks. 
Co-authors:  @ShunLiu-NOAA and @delippi

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,8 @@ project(RDASApp VERSION 1.0.0 LANGUAGES C CXX Fortran )
 
 include(GNUInstallDirs)
 enable_testing()
+#clt for rrfs test
+set(MPIEXEC_MAX_NUMPROCS 120 CACHE STRING "Maximum number of processors")
 
 # Build type.
 if(NOT CMAKE_BUILD_TYPE MATCHES "^(Debug|Release|RelWithDebInfo|MinSizeRel)$")
--- a/bundle/CMakeLists.txt
+++ b/bundle/CMakeLists.txt
@@ -123,7 +123,12 @@ endif()
    endif()
 
   endif(CLONE_JCSDADATA)
+     set(RDAS_RRFS_DATA_ROOT "$ENV{RDASAPP_RRFS_TESTDATA}/jcsda")
+     set(RDAS_RRFS_DATA_ROOT "/scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-emc-rdas/rrfs-test-data") #clttodo
+     set(DESTINATION_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
+     file( COPY "${RDAS_RRFS_DATA_ROOT}" DESTINATION ${DESTINATION_DIR} )
 
+  ecbuild_bundle( PROJECT rrfs-test SOURCE "../rrfs-test" )
 endif(BUILD_RDASBUNDLE)
 
 # Finalize bundle
--- a/rrfs-test/CMakeLists.txt
+++ b/rrfs-test/CMakeLists.txt
@@ -1,9 +1,8 @@
-# (C) Copyright 2017-2023 UCAR
-#
-# This software is licensed under the terms of the Apache Licence Version 2.0
-# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#!-------------------------------------------------------------------------
+#!    NOAA/NCEP, National Centers for Environmental Prediction  RDASapp        !
+#!-------------------------------------------------------------------------
 
-# FV3-JEDI test data - from fv3-jedi-data repo if found, from local directory, or from tarball
+# rrfs data - from rrfs-test-data repo if found, from local directory, or from tarball
 set(rrfs-test-data_FOUND 1)  #clttodo
 if (rrfs-test-data_FOUND)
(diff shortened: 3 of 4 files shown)


ShunLiu-NOAA:
@TingLei-NOAA could you give me a short instruction about how to test this case?

TingLei-NOAA:
@ShunLiu-NOAA  the added rrfs ctest is rrfs_fv3jedi_hyb_2022052619.  So, after building, just run this rrfs ctest to see if it pass.  Right now, the Jedi's "in-app"  test block (in the yaml ) is not added, so, it would pass just if it will finish smoothly.

ShunLiu-NOAA (dismissed):

bundle/CMakeLists.txt
```diff
@@ -123,7 +123,12 @@ endif()
    endif()
 
   endif(CLONE_JCSDADATA)
+     set(RDAS_RRFS_DATA_ROOT "$ENV{RDASAPP_RRFS_TESTDATA}/jcsda")
+     set(RDAS_RRFS_DATA_ROOT "/scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-emc-rdas/rrfs-test-data") #clttodo
```
CoryMartin-NOAA, line 127: Don't set this here, this is hard coded for Hera, so it should not be in the bundle CMakeLists.txt. Instead, define this data root in the modulefile?

CoryMartin-NOAA: Also the default data directory is being overwritten with this path?

rrfs-test/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+# (C) Copyright 2017-2023 UCAR
```
CoryMartin-NOAA: we are not UCAR, just remove this whole copyright

CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ project(RDASApp VERSION 1.0.0 LANGUAGES C CXX Fortran )
 
 include(GNUInstallDirs)
 enable_testing()
+#clt for rrfs test
+set(MPIEXEC_MAX_NUMPROCS 120 CACHE STRING "Maximum number of processors")
```
CoryMartin-NOAA, line 15: why is this set to 120 but then 80 is used in the test?

bundle/CMakeLists.txt
TingLei-NOAA: @CoryMartin-NOAA  Yes. the RDAS_RRFS_DATA_ROOT is supposed to be defined outside of RDASapp itself while being defined through env variables as the line 126 indicated.   The current Line 127 is a temporary setup and I will see if there could be a location on hera /orion shared by EMC RDAS developers.

CMakeLists.txt
TingLei-NOAA: It is the maximum allowed MPI numbers and hence this larger number is used now.

CoryMartin-NOAA: I guess I'm asking why is this defined, I'm nervous about putting something like this in at build time unless absolutely needed.

rrfs-test/CMakeLists.txt
TingLei-NOAA: @CoryMartin-NOAA Thanks for pointing this out. Any template on this for EMC ( or an appropriate one) copyrights?

CoryMartin-NOAA: No copyright needed, if we have a single license at the top level, that's more than sufficient. Plus fed employees technically can't copyright things

CMakeLists.txt
TingLei-NOAA: @CoryMartin-NOAA   It is because this number I speculate had been defined in the MPI module provided by RDAS/GDAS setup modules and that was a smaller number  (40).  While RDAS developers/myself don't have access to those modules,     It should be overridden with a larger number.  This location (in build.sh) to do this  is currently where I choose most transparent to RDAS developers/users.    Any suggestions on a better choice  is welcome.

CoryMartin-NOAA: No, that's not defined at all in GDASApp. 40 is the number of processors per node on Hera, and that is what CMake sets it to by default.

Just so I understand, are you saying that without this variable set, CMake will complain if you try to make a ctest with more than 40 processors?

TingLei-NOAA: Interesting to know 40 is chosen because it is the number of processors /node.  I think it is done in findMPI , right ? and findMPI could be set up while MPI related modules are provided . If that is right, I think its' behavior could be tuned in those modules.  
And yes, if it was 40,  the ctest requesting mpi task numbers larger than 40 would be disabled.

CoryMartin-NOAA: Got it, thanks. Let's leave it as you have it for now then.

rrfs-test/CMakeLists.txt
TingLei-NOAA: I changed to the current words.

CoryMartin-NOAA (approved):

ShunLiu-NOAA (approved):

PR #23, opened 2024-04-16
A rrfs Ens3Dvar test for mpasjedi is added

TingLei-NOAA:
Following the similar procedure by which a rrfs test for fv3jedi was added,  a rrfs Ens3Dvar test is added for mpasjedi. 
The ctest name is  rrfs_mpasjedi_2022052619_Ens3Dvar. 
Also, now the env variable pointing to the location of the local dir containing the rrfs data is set in the machine specific module file.

### DIFF
--- a/rrfs-test/CMakeLists.txt
+++ b/rrfs-test/CMakeLists.txt
@@ -16,36 +16,76 @@ if (rrfs-test-data_FOUND)
 else()
   message(FATAL_ERROR  "rrfs data would only be avaiable from local sourcerfs-fv3-jedi-data repository")
 endif()
-list( APPEND rrfs_fv3jedi_test_cases
-       rrfs_fv3jedi_hyb_2022052619
+if(FV3_DYCORE)
+   list( APPEND rrfs_fv3jedi_test_cases
+          rrfs_fv3jedi_hyb_2022052619
+       )
+   # JEDI config files
+   # ----------------
+   list( APPEND rrfs_fv3jedi_test_yamf_files 
+   rrfs_fv3jedi_hyb_2022052619.yaml
+   )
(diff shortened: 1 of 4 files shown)


rrfs-test/CMakeLists.txt
```diff
@@ -47,5 +76,11 @@ ecbuild_add_test( TARGET   ${target_test}
                   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/rundir-${target_test}
                   ARGS     ${target_test}.yaml 
                   COMMAND  fv3jedi_var.x )
+set(target_test rrfs_mpasjedi_2022052619_Ens3Dvar)
```
CoryMartin-NOAA: the build script has the option for either or both dycore right? should we add CMake to only build these tests if that dycore is built?

TingLei-NOAA: @CoryMartin-NOAA  Good suggestion. Will add this condition for them.

ShunLiu-NOAA (dismissed):
Looks good to me.

xyzemc (approved):

CoryMartin-NOAA (approved):
Thanks Ting!

CoryMartin-NOAA:
@TingLei-NOAA I noticed you `force pushed`, If you do that it makes reviews harder because I can't view "changes since your last review". I know that was the old GSI way of doing things, but now we don't have to force push, we can just push each commit and squash them when merging in to develop.

TingLei-NOAA:
@CoryMartin-NOAA Thanks for this reminder.  I will avoid using "push --force"  unless I am more strongly confident that some trivial changes are not interesting to the reviewers.

PR #28, opened 2024-04-22
changes for rrfs-mpas jedi rrfs test on orion 

TingLei-NOAA:
It is found a few changes needed for rdasapp building and running on orion.
One is as discussed in https://github.com/ecmwf/ecbuild/issues/61 when mpasjedi rrfs test case requires two nodes to get enough memory avoid the job fail.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -81,7 +81,7 @@ case ${BUILD_TARGET} in
     source $dir_root/ush/module-setup.sh
     module use $dir_root/modulefiles
     module load RDAS/$BUILD_TARGET.$COMPILER
-    CMAKE_OPTS+=" -DMPIEXEC_EXECUTABLE=$MPIEXEC_EXEC -DMPIEXEC_NUMPROC_FLAG=$MPIEXEC_NPROC -DBUILD_GSIBEC=ON"
+    CMAKE_OPTS+=" -DMPIEXEC_EXECUTABLE=$MPIEXEC_EXEC -DMPIEXEC_NUMPROC_FLAG=$MPIEXEC_NPROC -DBUILD_GSIBEC=ON -DMACHINE_ID=$MACHINE_ID"
     module list
     ;;
   $(hostname))
--- a/modulefiles/RDAS/orion.intel.lua
+++ b/modulefiles/RDAS/orion.intel.lua
@@ -83,7 +83,7 @@ setenv("RDASAPP_TESTDATA","/work2/noaa/da/cmartin/CI/RDASApp/data")
 setenv("RDAS_RRFS_DATA_ROOT","/work/noaa/fv3-cam/tlei/dr-emc-rdas/")
 prepend_path("PATH","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/prod_util/1.2.2/bin")
 
-execute{cmd="ulimit -s unlimited",modeA={"load"}}
+--cltorg execute{cmd="ulimit -s unlimited",modeA={"load"}}
 
 whatis("Name: ".. pkgName)
 whatis("Version: ".. pkgVersion)
--- a/rrfs-test/CMakeLists.txt
+++ b/rrfs-test/CMakeLists.txt
@@ -81,7 +81,7 @@ if(MPAS_DYCORE)
 
    endforeach()
   message(STATUS "MACHINCE_ID is " ${MACHINE_ID})
-   if("${MACHINE_ID}" STREQUAL "orion")
+   if("$ENV{MACHINE_ID}" STREQUAL "orion")
      message(STATUS "MACHINCE_ID ok " )
      if (MPI_ARGS)
        set(RESTORE_MPI_ARGS ${MPI_ARGS})
--- a/rrfs-test/CMakeLists.txt
+++ b/rrfs-test/CMakeLists.txt
@@ -80,12 +80,23 @@ if(MPAS_DYCORE)
       file(COPY ${src_casedir}/namelist.atmosphere_15km  DESTINATION ${casedir})
 
    endforeach()
-
-   set(target_test rrfs_mpasjedi_2022052619_Ens3Dvar)
-   ecbuild_add_test( TARGET   ${target_test} 
+  message(STATUS "MACHINCE_ID is " ${MACHINE_ID})
+   if("${MACHINE_ID}" STREQUAL "orion")
+     message(STATUS "MACHINCE_ID ok " )
+     if (MPI_ARGS)
+       set(RESTORE_MPI_ARGS ${MPI_ARGS})
+     endif()
+     set( MPI_ARGS "${MPI_ARGS} -N 2 --ntasks-per-node=18")
+   endif()
+     set(target_test rrfs_mpasjedi_2022052619_Ens3Dvar)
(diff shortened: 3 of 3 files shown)


modulefiles/RDAS/orion.intel.lua
```diff
@@ -83,7 +83,7 @@ setenv("RDASAPP_TESTDATA","/work2/noaa/da/cmartin/CI/RDASApp/data")
 setenv("RDAS_RRFS_DATA_ROOT","/work/noaa/fv3-cam/tlei/dr-emc-rdas/")
 prepend_path("PATH","/apps/contrib/NCEP/libs/hpc-stack/intel-2018.4/prod_util/1.2.2/bin")
 
-execute{cmd="ulimit -s unlimited",modeA={"load"}}
+--cltorg execute{cmd="ulimit -s unlimited",modeA={"load"}}
```
CoryMartin-NOAA, line 86: is this correct to remove this?

TingLei-NOAA, line 86: In my  rdasapp building, this line would cause building fail with message saying such changes are not allowed. 
I am wondering if it is a general issue or it was caused by my local env including the login nodes.

build.sh
```diff
@@ -81,7 +81,7 @@ case ${BUILD_TARGET} in
     source $dir_root/ush/module-setup.sh
     module use $dir_root/modulefiles
     module load RDAS/$BUILD_TARGET.$COMPILER
-    CMAKE_OPTS+=" -DMPIEXEC_EXECUTABLE=$MPIEXEC_EXEC -DMPIEXEC_NUMPROC_FLAG=$MPIEXEC_NPROC -DBUILD_GSIBEC=ON"
+    CMAKE_OPTS+=" -DMPIEXEC_EXECUTABLE=$MPIEXEC_EXEC -DMPIEXEC_NUMPROC_FLAG=$MPIEXEC_NPROC -DBUILD_GSIBEC=ON -DMACHINE_ID=$MACHINE_ID"
```
CoryMartin-NOAA, line 84: rather than pass it as a CMAKE option, we can grab the env var from the CMake

rrfs-test/CMakeLists.txt
```diff
@@ -80,12 +80,23 @@ if(MPAS_DYCORE)
       file(COPY ${src_casedir}/namelist.atmosphere_15km  DESTINATION ${casedir})
 
    endforeach()
-
-   set(target_test rrfs_mpasjedi_2022052619_Ens3Dvar)
-   ecbuild_add_test( TARGET   ${target_test} 
+  message(STATUS "MACHINCE_ID is " ${MACHINE_ID})
+   if("${MACHINE_ID}" STREQUAL "orion")
```
CoryMartin-NOAA:
```suggestion
   if("$ENV{MACHINE_ID}" STREQUAL "orion")
```

modulefiles/RDAS/orion.intel.lua
CoryMartin-NOAA: do you already set the stack size in your login environment?

TingLei-NOAA: @CoryMartin-NOAA , that is the tricky one.  I typed the above explanation of mine before seeing your question. 
If GDASapp has this command and no issue.  it shouldn't occur to rdasapp.

CoryMartin-NOAA: Yeah it's been there for a while with no complaints: https://github.com/NOAA-EMC/GDASApp/blob/3bb46b61e710a77e85916d9b20388df1e1f390f3/modulefiles/GDAS/orion.intel.lua#L86 

Probably good for other reviewers to chime in before deciding to keep/remove

rrfs-test/CMakeLists.txt
TingLei-NOAA: Good.  But, then, we need to export it in the build.sh , right?  otherwise, env in the child script will not get it by env. 
I will try it.

build.sh
TingLei-NOAA: I will first try your suggestion

rrfs-test/CMakeLists.txt
CoryMartin-NOAA: probably but not 100% sure

modulefiles/RDAS/orion.intel.lua
ShunLiu-NOAA: It is preferable to be consistent with GDASApp.

TingLei-NOAA: @CoryMartin-NOAA  It seems a newly cloned gdasapp has the same behavior on orion.  I need to comment out that line for building of gdasapp to continue.

CoryMartin-NOAA: Do others have this issue? It sounds like something specific to your environment

CoryMartin-NOAA:
```
cat ~tlei/.bashrc
...
ulimit -s 1900000
...
```
You set the stack size at login, it probably doesn't let you re-set it

TingLei-NOAA: @CoryMartin-NOAA  To make sure if that line in the  .bashrc does have any impact, I commented out it and re-ran the building and it was the same behavior.  But interestingly, if I directly ran "module load " on the front end, there is no problem.

CoryMartin-NOAA: @TingLei-NOAA this is very strange

TingLei-NOAA: If no one else has the same problem I will do some digging when I have a chance :).

SamuelDegelia-NOAA: I was able to successfully run the mpasjedi case on Orion by sourcing the default orion module file (i.e., keeping the unlimited stack size command). This was done using a basic job file and not through the ctest, though.

TingLei-NOAA: @SamuelDegelia-NOAA  so your running of the build.sh (which would use that module ) has no problem?

SamuelDegelia-NOAA: @TingLei-NOAA Correct, I had no problem building using the default build.sh. I just followed the wiki instructions line for line and everything worked for me.

TingLei-NOAA: @SamuelDegelia-NOAA  Thanks. then seems some set on my side caused that issue.

SamuelDegelia-NOAA: Is there a specific error message that occurs when you build?

TingLei-NOAA: The error message said those re-set are not allowed.

TingLei-NOAA:
Dear All, I will close this PR because 1) seems the failure "ulimit -s... " only happen to me 2) Sam has successfully run mpas-jedi rrfs case without requesting 2 nodes, that means maybe the failure of my run using one node also came from 1.  Hence, the major change in this PR is maybe not needed.
Thanks to all your comments/suggestions in this PR.

Issue #24, opened 2024-04-17
ENKF test case


hu5970:

hu5970:
run LETKF for FV3

SamuelDegelia-NOAA:
Closed due to PR #35.

PR #35, opened 2024-05-08
Added LETKF ctest and experiment template

SamuelDegelia-NOAA:
Following #24, a yaml file for an LETKF test case is added here for FV3-JEDI (`rrfs-test/testinput/rrfs_fv3jedi_letkf_2022052619.yaml`). The ctest is also added through edits to `RDASApp/rrfs-test/CMakeLists.txt`. This LETKF case uses the same ensemble data as the hybrid test case for FV3-JEDI so that no additional data need to be copied over when building. Additional changes in this PR include: 

- Modifications to `rrfs-test/scripts/setup_experiment.sh` to set up the experiment directory for running LETKF. This script copies the test data from the hybrid case into the LETKF directory and then removes the old hybrid yaml and replaces it with the LETKF yaml
- Modifications to the `rrfs-test/scripts/templates/run*sh` scripts to now also pass in a specific yaml file during `setup_experiments.sh`.
- Renaming Donnie's original python plotting scripts to add the `singleob` identifier since those are designed for plotting the mesonet single observation tests.
- A new `fv3jedi_increment_fulldom.py` plotting script that can be used for tests other than the single observation tests. Without any editing, this script serves as an example of plotting the LETKF case. 

As before, I also made edits to the Wiki page to describe this new ctest and test experiments. Those edits are given on my fork here: https://github.com/SamuelDegelia-NOAA/RDASApp/wiki/Running-FV3%E2%80%90JEDI-or-MPAS%E2%80%90JEDI-ctest-and-setting-up-an-experiment


rrfs-test/scripts/setup_experiment.sh
```diff
@@ -26,12 +28,19 @@ if [[ ! -d $YOUR_PATH_TO_RDASAPP && ! -d `dirname $YOUR_EXPERIMENT_DIR` ]]; then
   exit 1
 fi
 
+# Error check if choosing LETKF for MPAS 
+if [[ $DYCORE == "MPAS" && $DA_METHOD == "LETKF" ]]; then
+   echo "ERROR: LETKF is not et implemented for MPAS"
```
CoryMartin-NOAA:
```suggestion
   echo "ERROR: LETKF is not yet implemented for MPAS"
```

CoryMartin-NOAA (commented):
minor issue otherwise looks good to me

SamuelDegelia-NOAA:
Whoops, thanks for catching the typo!

CoryMartin-NOAA (dismissed):

TingLei-NOAA:
@SamuelDegelia-NOAA  a problem i have been hoping to have a chance to discuss with RDASapp developers is the use of Data dir.  For efficiency consideration, I would recommend all runs use the same Data dir if possible (like using the same background /ensembles, and all obs could be put into the same location ) So, in that setup,  only yaml (and maybe the reference output if used) would need to be copied ,or even we just copy yaml in one location like testinput,  the developer adding new ctests would only need to add a ecbuild_add_test block.   And, by doing so, also no need to add special logic treatment as you are doing now. 
We could make the casedir name is more generic like rundir-rrfs_fv3jedi_2022052619 rather than rrfs_fv3jedi_**hyb**_2022052619

SamuelDegelia-NOAA:
> @SamuelDegelia-NOAA a problem i have been hoping to have a chance to discuss with RDASapp developers is the use of Data dir. For efficiency consideration, I would recommend all runs use the same Data dir if possible (like using the same background /ensembles, and all obs could be put into the same location ) So, in that setup, only yaml (and maybe the reference output if used) would need to be copied ,or even we just copy yaml in one location like testinput, the developer adding new ctests would only need to add a ecbuild_add_test block. And, by doing so, also no need to add special logic treatment as you are doing now. We could make the casedir name is more generic like rundir-rrfs_fv3jedi_2022052619 rather than rrfs_fv3jedi_**hyb**_2022052619

Ting, I agree that using a data directory cleaner solution rather than coping over entire experiment directories. We would still likely need two data directories (one for fv3jedi and one for mpasjedi), but it would make it a lot easier to add additional ctests or example yaml files. 

Should we go ahead and approve this PR just to get the LETKF yaml in RDASApp per Issue #24? Then I could work on setting up a clean data directory and other related changes.

ShunLiu-NOAA (dismissed):

TingLei-NOAA:
@Samuel Degelia - NOAA Affiliate ***@***.***>  we could
discuss more in our meeting.

______________________________

Ting Lei

Physical Scientist, Contractor with Lynker in support of

EMC/NCEP/NWS/NOAA

5830 University Research Ct., Cubicle 2765

College Park, MD 20740

***@***.***

301-683-3624

SamuelDegelia-NOAA:
Following Ting's suggestions, this PR has been updated with a new commit that reorganizes the staged input data from what was previously directories for each ctest, to now more general data directories for fv3jedi or mpasjedi. This change allows users to more easily add ctests or experiments without adding special logic to `CMakeLists.txt`. It also allows us to now link data from these input data for each ctest (saving build time and disk space) instead of copying the input data individually into each ctest directory in `build/rrfs-test/rundir-*`. 

The new input data that I reorganized can be found in these locations: 

Orion: `/work2/noaa/fv3-cam/sdegelia/RRFS/rrfs-test-data`
Hera: `/scratch1/NCEPDEV/da/Samuel.Degelia/RRFS/rrfs-test-data`

More specific changes in this new commit include: 

- Made an effort to unify the directory structures between `rrfs-data_fv3jedi_2022052619` and `rrfs-data_mpasjedi_2022052619` (most of the changes were for mpasjedi) 
- Edited `rrfs-test/CMakeLists.txt` to now link in the ctest input data, and removed special logic for LETKF from last commit. This cmake file features the most changes here so it might be easier to just look at the new file then compare line changes 
- Modified the yaml files to now output experiment or ctest results into the main directory instead of adding them into the linked subdirectories 
- We now retrieve the yaml for the ctests from the repository itself in `RDASApp/rrfs-test/testinput`. This allows a user to easily add a new ctest into `RDASApp/rrfs-test/CMakeLists.txt` following the instructions there or in the last section of my [forked Wiki](https://github.com/SamuelDegelia-NOAA/RDASApp/wiki/Running-FV3%E2%80%90JEDI-or-MPAS%E2%80%90JEDI-ctest-and-setting-up-an-experiment-reorganized). Additionally, other collaborators can now easily add ctests or experiments by committing to the main repo instead of needing to modify the input data
- Modified `rrfs-test/scripts/setup_experiment.sh` and the python scripts to support this reorganized input/output data structure

Additionally, I understand that reorganizing the staged input data structure like this is a bit of a big change, so we can feel free to discuss here whether these changes are worth it. @TingLei-NOAA, I modified the module files to link to the staged data in my directories, but if we agree with this PR then you can feel free to copy back onto your directory.

ShunLiu-NOAA (approved):

CoryMartin-NOAA (approved):

bundle/CMakeLists.txt
```diff
@@ -127,16 +127,15 @@ endif()
...
   if(CLONE_RRFSDATA)
      set(RDAS_RRFS_DATA_ROOT "$ENV{RDAS_RRFS_DATA_ROOT}/rrfs-test-data")
-#    set(RDAS_RRFS_DATA_ROOT "/scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-emc-rdas/rrfs-test-data")
      set(DESTINATION_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
      if(NOT EXISTS "${DESTINATION_DIR}/rrfs-test-data")
         file(MAKE_DIRECTORY "${DESTINATION_DIR}/rrfs-test-data")
      endif()
      if(FV3_DYCORE)
-     file( COPY "${RDAS_RRFS_DATA_ROOT}/rundir-rrfs_fv3jedi_hyb_2022052619" DESTINATION "${DESTINATION_DIR}/rrfs-test-data" )
+     file( COPY "${RDAS_RRFS_DATA_ROOT}/rrfs-data_fv3jedi_2022052619" DESTINATION "${DESTINATION_DIR}/rrfs-test-data" )
```
TingLei-NOAA, line 135: for easy to understand, I would suggest use names like "base_run_dir-rrfs_fv3jedi_hyb_2022052619" or something like that.

SamuelDegelia-NOAA: My thought was that we would have experiment-agnostic data directories. That way the fv3jedi directory here could also be used for the LETKF ctest. Would including "hyb" here and in `RDASApp/bundle/rrfs-test-data` make it more confusing?

TingLei-NOAA (approved):

delippi:
Hi everyone, sorry about this. I should've worked faster to review this PR before it got merged ha!

I was going through these updates today and I think some of this is good and some of this might not really be necessary. I'm specifically talking about the setup_experiment.sh script, at least. Basically all this setup experiment script is meant to do is copy the data and some generic scripts for a user to start working. It doesn't need to reference the input yaml, the executable, or DA_METHOD. A user should just add their own yaml to the testinput directory and can then just pass it as an argument in the jedi run script for their own use case.

I also kind of prefer the rsync vs linking things. It really isn't that much data (18G for the fv3-jedi case--the MPAS case is a bit larger but still not that bad).

If it is okay, I'm going to revert some of the changes in the setup experiments script in my GSI test data PR?

There were parts that were okay, such as renaming the directory to not have "hyb" because the data doesn't care what you are running. It is the same for hyb as it is for LETKF etc. Those differences are all in the yaml file.

TingLei-NOAA:
@delippi  That is good your have ideas to further improve this PR.  I think it is better/simpler that you could open an separate PR rather than combining with your another PR.

SamuelDegelia-NOAA:
@delippi Thank you for reviewing the changes. I agree that `setup_experiment.sh` could be cleaned up. I also realize now that `$DA_METHOD` could especially be removed since that is a remnant from the first version of this PR. 

For the rest of the arguments I added: I included the `EXE` input in `setup_experiment.sh` because the executable will differ if you want to run an experiment with LETKF or a variational method. Previously, the template scripts always copied in the variational executable. So if we want to remove `EXE` then the template scripts will also need to change. We could just copy in both the letkf and variational executables to make it easier. And regarding the yaml input, I see now that you intend this script to be more of an informal setup for a user to get started running experiments, so I think the yaml input (and associated code that creates a run directory based on the yaml file) could be dropped too. 

For rsync vs linking, I guess I don't really understand why rsync would be better. Could you elaborate on that? We are no longer outputting any results to those Data directories, so why not save time and space by linking? I added linking for the ctests since we are creating individual directories for each test and creating so many copies of those data directories could add up overtime. But I will also lean towards whatever your decision is here since I understand that I am much newer to the system. I'll make sure that I run these changes by you in the future @delippi.

delippi:
Hey @SamuelDegelia-NOAA,  no worries. I should’ve provided a review sooner. I was planning to get to it this week, and I didn’t expect the PR to be merged so soon.

Yep, the script is really just meant to be a basic user setup to “get going” running their own experiments. I see what your intention is with the EXE option though. I wouldn’t mind having something there, but I’m not sure what yet.

Rsync vs link: If you linked the data for ctests, that is probably fine. I think you actually have more experience with ctests that I do! I’m just talking about the links in the setup_experiment script. My ideas are by no means the best, but this one is sort of a personal preference for me. I like having a copy of the data knowing that I have a safe copy to go back to and it allows the directory to be a little cleaner. The copy is really only done once and for me only takes about 80s (for fv3jedi data) so it isn’t really saving much time. You also don’t have to worry about if the directory housing those files gets deleted like if you are cleaning up old versions of RDASApp or something happens to it while recompiling. If that happens you might end up using more time to recompile and copy the staged data. You can also use the same data directory with a new version of the code and just modify where the executables are and such in the script. I guess my thinking was to keep it completely untethered from a specific RDASApp install since the data is reusable and you can just point the run script to a new install manually. I’m probably just being overly picky here, but I just prefer having a copy instead of a link hahaha. I also think a user might only have at the very most a handful of these data directories for their experiments and at 18GB per that is really not an issue. Fv3-cam scratch2 has like 45TB of free space currently.


> I'll make sure that I run these changes by you in the future @delippi.

That’s what the review process is for and I didn’t make mine quick enough this time. So my apologies.

One last thing I see at the moment is that there are yamls in the staged directory. Those shouldn’t need to be there. Just in rrfs-test/testinput.

@TingLei-NOAA, I'll put these changes in a new PR if you think that is best. I will need these changes for my GSI test data PR so the new one will need to go first.

TingLei-NOAA:
@delippi  Sure. to create a separate PR to show the change you proposed would make the discussion/reviewing on them much more simpler.  After they go to the develop branch, it will also be easy to merge them to your current PR. 
Thanks.

SamuelDegelia-NOAA:
@delippi I understand why using rsync would be better for `setup_experiments.sh` now - that makes sense to me and I am okay with reverting the changes there. Also, I will go ahead and remove the yaml files from the staged directory too.

PR #36, opened 2024-05-10
Add GSI test data and scripts for FV3-JEDI validation

delippi:
This is the initial PR for adding GSI test data for validation experiments of the FV3-JEDI 2022052619 case. The point of this is to allow users to set up a basic GSI case to compare against their FV3-JEDI case for validation purposes. We need to stage the GSI data somewhere on Hera and Orion--probably with the JEDI test data? As for now, I have it located here:
`/scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2/staged-data/gsi_2022052619` and currently only on hera.

The GSI data is retrieved using the rrfs-test/scripts/setup_experiment.sh script by setting two new user defined settings `GSI_TEST_DATA` and `YOUR_PATH_TO_GSI`. The setup experiment will then copy template scripts to the gsi run directory and ready for a user to start their work. There are two main gsi run scripts, the first runs `run_gsi_hera.sh` the gsi itself and the second `run_gsi_ncdiag_hera.sh` creates the gsi-diag derived ioda obs file. I also added a script to plot a scatter plot of the gsi and jedi hofx ratios.

The next PR should include EVA verification tool and support for Orion.

Wiki was updated in the main repo [Running FV3‐JEDI Hofx Validation](https://github.com/NOAA-EMC/RDASApp/wiki/Running-FV3%E2%80%90JEDI-Hofx-Validation)


delippi:
This is kind of the idea I was thinking to get us started with a GSI test case available. It's still draft PR because there are obvious things that need fixed. I'm getting off a little early and I wanted to get this saved before leaving.

TingLei-NOAA:
@delippi   Basically,  this test is supposed to run a gsi as the reference?  Will it be more "efficient" to just use the results from GSI (either diag files or/and final results).  What is the standard practice for ufo ctests? 
And, i would prefer those one observation tests from the prebufr only with one obs (different files correspond to different observation types).

SamuelDegelia-NOAA:
I think this is a good idea to include. My only thought so far is to include the GSI verification as a separate script from `setup_experiment.sh` (maybe called `setup_gsi_verif.ksh`). This way we keep the focus of the main setup script for running JEDI experiments, but give users the option to do verification separately if they want. 

Also, does this even need to be added as a ctest or would it okay to just leave them as tools available in the repository? I'm still not super familiar with ctests but would the goal of the test be to compare the Hofx between GSI and JEDI?

delippi:
@TingLei-daprediction and @SamuelDegelia-NOAA, the point of this is to make the test data available for other developers to run hofx validation, validation of the solvers, and other aspects that might need validation for the JEDI transition. It could be used for helping create additional ctests but that is not really my main intention with this PR. @ShunLiu-NOAA has a whole list of ob platforms that need tested and we are supposed to be coordinating that effort with developers from GSL too. There's also the list of obs in the RRFS JEDI transition gantt smartsheet.

> I think this is a good idea to include. My only thought so far is to include the GSI verification as a separate script from `setup_experiment.sh` (maybe called `setup_gsi_verif.ksh`). This way we keep the focus of the main setup script for running JEDI experiments, but give users the option to do verification separately if they want.

The point of having the FV3-JEDI case and the set up script is to do validation experiments whether that be for hofx validation, ob error validation, validation of the solvers, or to make sure any other aspect of the system performs as expected and if not bit-for-bit identical is at least comparable. We can't transition from GSI to JEDI without doing the validation work. And we can't do the validation work without the GSI case. Furthermore, users can just not copy the GSI case by setting one of those options to "NO" at the top of `setup_experiments.sh`, so no extra script is needed and users have the option to copy the data or not.

ShunLiu-NOAA:
@delippi, @SamuelDegelia-NOAA and @TingLei-NOAA Thank you for the good discussion here. I think we have to setup a templet for UFO validation when we integrate observations to RDAS-JEDI. Let's discuss how to prepare data, utilities and the appropriate validation procedure for this purpose.

delippi:
Hi everyone, I think this is ready for review now. I decided to not add the EVA tools yet to help keep the PR simple. I will add that in the next one. We can/should add another PR to expand on this PR for Orion support which I haven't done here. Again, just to start as simple as possible. You should also be able to follow my [instructions](https://github.com/delippi/RDASApp/wiki/Running-FV3%E2%80%90JEDI-Hofx-Validation) on my wiki page if you wanted to test setting up a GSI case and doing some initial validation (at least with my mesonet obs).

Oh, I also wasn't sure how to compile the GSI to get `nc_diag_cat_serial.x`. I'm not sure how to do that anymore. That is needed for this.

rrfs-test/scripts/templates/run_gsi_hera_template.sh
```diff
@@ -0,0 +1,69 @@
...
+  # omi_aura ssmi_f13 ssmi_f14 ssmi_f15 hirs4_n18 amsua_n18 mhs_n18 \
+  # amsre_low_aqua amsre_mid_aqua amsre_hig_aqua ssmis_las_f16 \
+  # ssmis_uas_f16 ssmis_img_f16 ssmis_env_f16 mhs_metop_b \
+  # hirs4_metop_b hirs4_n19 amusa_n19 mhs_n19"
+  listall=`ls pe* | cut -f2 -d"." | awk '{print substr($0, 0, length($0)-3)}' | sort | uniq`
+  for type in $listall; do
+    count=`ls pe*${type}_${loop}* | wc -l`
+    if [[ $count -gt 0 ]]; then
+      #cat pe*${type}_${loop}*.nc4 > diag_${type}_${string}.${ANAL_TIME}
+      nc_diag_cat_serial.x -o diag_${type}_${string}.${ANAL_TIME} pe*${type}_${loop}*.nc4
```
SamuelDegelia-NOAA, line 66: This job failed for me here during my first attempt at running it. Should this be a `./nc_diag_cat_serial.x -o diag_${type}_${string}.${ANAL_TIME} pe*${type}_${loop}*.nc4` since nc_diag_cat_serial.x isn't added to $PATH (at least for me)?

delippi: Hi @SamuelDegelia-NOAA, First of all, I really appreciate you stress testing this!

So far, I've only added a precompiled  `nc_diag_cat_serial.x` which should just be in my GSI staged data directory (probably bad practice). It should get copied to your run directory, so you can try `./` before. This was fairly old exe that I found in an old directory, I'm not sure how to compile GSI to get this anymore, so this is something I still need to figure out.

Or is the problem that you don't have the executable? I also tried running this with a single ob test, and if you do that it fails because it needs more than one `pe*${type}_${loop}*.nc4`. If that doesn't help, point me to your directory so I can see what you have.

SamuelDegelia-NOAA: The executable works and was copied into the GSI run directory. I just needed to add the `./` before it in `run_gsi_hera_template.sh` to solve my problem. And I don't have any problem including a precompiled version of the executable but others my differ on that

SamuelDegelia-NOAA:
Donnie, this was really helpful for me to get an idea of how to generate the IODA files from GSI. After making the change from my above comment, everything worked as described in your wiki. The only minor comment I have pertains to your wiki. Like was noted in the comments for `run_gsi_ncdiag_hera.sh`, I had to modify `RDASApp/build/lib/python3.10/pyiodaconv/gsi_ncdiag.py` to add obtypes 188 and 195 for surface obs: 

```
conv_bufrtypes = {
...
    "sfc": [181, 187, 188, 195], 
...
}

```
It might be worth mentioning this in the wiki page directly since running `run_gsi_ncdiag_hera.sh` will not be successful without this change.

SamuelDegelia-NOAA (approved):

delippi:
> Donnie, this was really helpful for me to get an idea of how to generate the IODA files from GSI. After making the change from my above comment, everything worked as described in your wiki. The only minor comment I have pertains to your wiki. Like was noted in the comments for `run_gsi_ncdiag_hera.sh`, I had to modify `RDASApp/build/lib/python3.10/pyiodaconv/gsi_ncdiag.py` to add obtypes 188 and 195 for surface obs:
> 
> ```
> conv_bufrtypes = {
> ...
>     "sfc": [181, 187, 188, 195], 
> ...
> }
> ```
> 
> It might be worth mentioning this in the wiki page directly since running `run_gsi_ncdiag_hera.sh` will not be successful without this change.

Yeah, I made a note in that ncdiag script for needed to add ob types of *88 and *95, but I wasn't really clear what needed changed. It might be better for me to just make a PR to iodaconv and update that particular script. For now, I'll add something to the wiki to better indicate that that change is needed as you have suggested.

Thank you very much for running through the instructions! I'm glad it was helpful. It is my intention that it should be easy for users to get started working on things with good, clear instructions and good documentation.

Edit: I just updated my wiki with that information.

TingLei-NOAA (approved):

ShunLiu-NOAA (approved):

Issue #38, opened 2024-05-14
Ioda-converters YAML for the RRFS MSONET prepBUFR data


PraveenKumar-NOAA:
**Description**

Write an example YAML for the MSONET prepBUFR RRFS data that converts a BUFR file into an IODA file using the bufr2ioda.x for the following subset:

| MSONET   | A48117 | MESONET SURFACE REPORTS (COOPERATIVE NETWORKS)           |

- [x]  Extract the BUFR table, queries, and important Mnemonics information
- [x]  Use this YAML to convert into an IODA-formatted netCDF file
- [x]  Ioda obs validation using ioda-validate.x
- [x]  Perform validation for the observations using the input BUFR and output IODA file
- [x]  Add output(s) to the testoutput directory
- [ ]  Add tests to the CMakeLists.txt files
- [ ]  Have the tests pass

**Requirements**

- All important mnemonics required by the RRFS users need to be known.

**Acceptance Criteria**

- It will be finished when all files have been checked and tests have passed.

**Dependencies**

- _bufr2ioda.x_

PraveenKumar-NOAA:
**Validation of the output ioda-obs using the ioda-validate.x:** 

**Final results:**
_errors: 0
warnings: 0_

PraveenKumar-NOAA:
Validation of the output IODA results, obtained using the MSONET prepBUFR YAML and bufr2ioda.x, and input prepBUFR file for the following two observations:

**pressure (POB):**
BUFR plot:
![bufr_pressure (2)](https://github.com/NOAA-EMC/RDASApp/assets/63739712/36660692-a000-44f1-9910-41bdeaf7519d)

IODA plot: 
![ioda_pressure (1)](https://github.com/NOAA-EMC/RDASApp/assets/63739712/55aad2a4-413e-46b4-ad9b-a7e90d7dcf8f)


**windEastward (UOB):**
BUFR plot:
![bufr_windEastward (2)](https://github.com/NOAA-EMC/RDASApp/assets/63739712/781b446a-9188-4314-b1b5-b16b72a6bad0)


IODA plot: 
![ioda_windEastward](https://github.com/NOAA-EMC/RDASApp/assets/63739712/b344b291-2ca3-4c03-8d0f-ef66e4edca77)

PraveenKumar-NOAA:
In the the IODA wind plot, there are more observations compaired to the BUFR plot, which are due to the missing values. There are a total of 181121 missing values.

[...]

PR #45, opened 2024-05-22
MSONET prepBUFR YAML and test data

PraveenKumar-NOAA:
This PR add an ioda-converters YAML for the MSONET prepBUFR data and test data for the following date 2022052619. 

**The following three files are new:** 

- test/testinput/bufr_ncep_prepbufr_msonet.yaml
- test/testinput/rap.t19z.prepbufr.MSONET.tm00 (prepBUFR data only for MSONET)
- test/testoutput/rap.t19z.prepbufr.MSONET.tm00.nc

The validation for the output obs is performed  in #38.


delippi (changes requested):
Hi @PraveenKumar-NOAA, thanks for putting this together it will be nice to be able to use an "official" and more thoroughly tested configuration to convert mesonet data than what I put together.

I wonder if these should go under rrfs-test? 

Also, I want to try to use the output from these yamls. In just comparing against my old yaml (which is not necessarily correct) I saw a few differences that I wanted to double check.

PraveenKumar-NOAA:
> Hi @PraveenKumar-NOAA, thanks for putting this together it will be nice to be able to use an "official" and more thoroughly tested configuration to convert mesonet data than what I put together.
> 
> I wonder if these should go under rrfs-test?
> 
> Also, I want to try to use the output from these yamls. In just comparing against my old yaml (which is not necessarily correct) I saw a few differences that I wanted to double check.

@delippi thanks. Its up to you where you want the test data, but I always prefer it to be with the converter. Let me know if you have any question/ concern about the converter.

delippi:
> Hi @PraveenKumar-NOAA, thanks for putting this together it will be nice to be able to use an "official" and more thoroughly tested configuration to convert mesonet data than what I put together.
> 
> I wonder if these should go under rrfs-test?
> 
> Also, I want to try to use the output from these yamls. In just comparing against my old yaml (which is not necessarily correct) I saw a few differences that I wanted to double check.

Hi @PraveenKumar-NOAA, I ran the converter using your yaml and compared it against the result from my yaml. I have a few questions to start some discussion with the rest of the group:

1. I don't think we need ObsError in the yaml because I think that should be defined at the JEDI yaml level since that should be application specific.
2. It looks like an ObsValue for stationPressure (and also under QualityMarker) is missing. I do see "pressure" in QualityMarker. I've seen it references as stationPressure in other yamls, so that was what I was using. But I also had it called pressure in the MetaData. I'm not really sure what the best thing to do here.
3. I have a feeling there is some disconnect between the converters and the jedi yamls on specifying the QC info. In all of the converter yamls I've seen they use "QualityMarker" and in jedi it is often referenced as "PreQC". Again, I'm not sure what the best thing to do here.
4. I didn't see a MetaData entry for DHR parameter (obsTimeMinusCycleTime). I'm sure that will be needed at some point to do time filtering on the data.
5. I didn't see an ObsType group which is something used in certain filters to apply only to certain bufr types.

@CoryMartin-NOAA, are you able to give some guidance on any of these points?

delippi:
> > Hi @PraveenKumar-NOAA, thanks for putting this together it will be nice to be able to use an "official" and more thoroughly tested configuration to convert mesonet data than what I put together.
> > I wonder if these should go under rrfs-test?
> > Also, I want to try to use the output from these yamls. In just comparing against my old yaml (which is not necessarily correct) I saw a few differences that I wanted to double check.
> 
> @delippi thanks. Its up to you where you want the test data, but I always prefer it to be with the converter. Let me know if you have any question/ concern about the converter.

The `test/` is at the same level of `rrfs-test/`. It seems it could be all in one? @ShunLiu-NOAA @TingLei-NOAA what do you suggest?

TingLei-NOAA:
@PraveenKumar-NOAA and @delippi  Thanks for setting up such  good "questions/PR" for discussions. 
I have two cents 
1) about giving obs errors in yaml ,   I think obs errors given in the ioda convertor level is good, which could represent the best estimate of the observation errors from the observations themselves from the obs-processing experts , for example, they could specify them using GSI 's stats.   In the downstream,  like in variational app or letkf,  if the users/developers want to change them, the changes should be validated/explained. 
2) the testoutput normally contain text output from various tests.  For the converted data, maybe, we could create a new dir to hold those nc files, like obs-output ?

ShunLiu-NOAA:
@PraveenKumar-NOAA @delippi I think it is better to host YAML file in ./rrfs-test/IODA/yaml, and prepbufr/bufr and IODA data should be in a local directory on HERA and ORION so that we do not have to worry about permission issues.

CoryMartin-NOAA:
please @PraveenKumar-NOAA DO NOT store any binary/netCDF files in a GitHub repository. We do not have git-lfs (like JCSDA does) and thus this will balloon the size of the repository.

PraveenKumar-NOAA:
> please @PraveenKumar-NOAA DO NOT store any binary/netCDF files in a GitHub repository. We do not have git-lfs (like JCSDA does) and thus this will balloon the size of the repository.

@CoryMartin-NOAA thanks, I removed the binary/netCDF files and moved converter to the following directory ~/rrfs-test/IODA/yaml.

delippi:
> > please @PraveenKumar-NOAA DO NOT store any binary/netCDF files in a GitHub repository. We do not have git-lfs (like JCSDA does) and thus this will balloon the size of the repository.
> 
> @CoryMartin-NOAA thanks, I removed the binary/netCDF files and moved converter to the following directory ~/rrfs-test/IODA/yaml.

@CoryMartin-NOAA, I'm curious if this will still be a problem since the binaries were committed? Will they still be in the commit history similar to the problem we are trying to resolve now?

CoryMartin-NOAA:
@delippi because we do squash commits, it won't be a problem when it's merged in, only the final commit gets actually saved in the repository.

delippi:
> @delippi because we do squash commits, it won't be a problem when it's merged in, only the final commit gets actually saved in the repository.

Thanks @CoryMartin-NOAA, that is really good to know.

PraveenKumar-NOAA:
> prepbufr/bufr and IODA data should be in a local directory on HERA and ORION

@ShunLiu-NOAA thanks. Any preference for a data directory path on HERA and ORION for the prepbufr/bufr and IODA data files, otherwise I will create one in my space?

ShunLiu-NOAA:
@PraveenKumar-NOAA Could you please work with @SamuelDegelia-NOAA so that your data can be stored in rrfs-test-data/rrfs-data_fv3jedi_2022052619/Data/bufr2ioda?

SamuelDegelia-NOAA:
I agree with Donnie's comments. At the very least, stationPressure should be added into ObsValue before we can merge this. That should be a fairly easy change. For my own knowledge of just trying to understand how bufr2ioda works, I also found that the DHR variable can be included first by adding a new export as: 

```
        timeOffset:
           query: "*/DHR"
```

And then later on, we can output it to MetaData through: 

```
        - name: "MetaData/timeOffset"
          coordinates: "longitude latitude"
          source: variables/timeOffset
          longName: "Difference from cycle time"
          units: "hours"
```

I am not sure how to include the missing ObsType group, but if it is used for QC filters then it probably should also be included.  

As an aside, for anyone running `bufr2ioda.x` on Orion, I originally experienced an error about missing `libingester.so` in my environment. On Hera, this seems to be taken care of by loading the RDAS modules. But on Orion, I had to do `export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/RDASApp/build/lib` to solve the problem.

delippi:
> 1. about giving obs errors in yaml ,   I think obs errors given in the ioda convertor level is good, which could represent the best estimate of the observation errors from the observations themselves from the obs-processing experts , for example, they could specify them using GSI 's stats.   In the downstream,  like in variational app or letkf,  if the users/developers want to change them, the changes should be validated/explained.

Hi @TingLei-NOAA to one of your points, I really don't think we want ob errors in the IODA file. I seem to recall @JacobCarley-NOAA saying this before. First, this would assume that those errors even exist in the bufr/prepbufr file. Which I was just looping through some and I wasn't seeing them in any of the couple messages that I looked at (using `readmp`). Second, obs-processing folk shouldn't have to specify the ob errors from the GSI or to how we might tune them in the future. The converters is meant just to take the information from the bufr format into netcdf IODA. We can and should just specify the ob errors ourselves in the jedi yaml. I feel like this would be a lot easier to do anyway. Third, if the ob errors are specified in the IODA then if we wanted to change ob errors we would have to reprocess the data. I really think the best thing to do is just not have it there.

TingLei-NOAA:
@delippi   I think we don't need to assume observation errors variables exist in the prepbufr. I just want to the ioda converted from bufr contains the observation error values which represent the best estimates from obs-process experts, which are the "initial" observation errors (and, hence, not relevant to how other treatment on obs error would be realized).  Maybe, finally, using some  yaml /fix files to contains those initial obs errors are what we will do as GSI does. For being now, and a long time in the near future, I think putting those initial obs errors in the ioda data is still the more convenient way to communicate this "initial" obs errors quantities.

ShunLiu-NOAA:
@TingLei-NOAA I think at the moment we can follow GDAS IODA infrastructure.  If there is no initial obs errors in the GDAS ioda data for some variables, we may just follow this infrastructure at the moment. We can further discuss this with obsproc team about the initial obs error in bufr/IODA later. But I guess it will take time and extra work to get the initial obs error.

PraveenKumar-NOAA:
@delippi @ShunLiu-NOAA @TingLei-NOAA @SamuelDegelia-NOAA I made changes to the YAML. Please check it out.

SamuelDegelia-NOAA:
Thanks @PraveenKumar-NOAA! I see the ObsType group was added and includes the stationElevation and stationPressure variables. Do you know if all the variables need to be in there for the filters you were referencing @delippi?

delippi:
@PraveenKumar-NOAA, thanks for making these changes. As @SamuelDegelia-NOAA was saying there should be an ObsType for all the variables in addition to only the stationElevation and stationPressure. All the groups like QualityMarker, ObsValue, ObsType should have the same number of variables.

I still maintain that ObsError does not need to be in here and that ob errors should only be defined in the jedi yamls. I'm happy to further discuss.

PraveenKumar-NOAA:
@delippi @ShunLiu-NOAA @SamuelDegelia-NOAA @TingLei-NOAA I have added a few more ObsType variables and removed all ObsError variables from the YAML. @delippi 'QualityMarker' is fine as we are using this in global converters and we can rename that to PreQC in UFO YAML easily.

PraveenKumar-NOAA:
If you review/ approve this today, I can provide test data to @SamuelDegelia-NOAA.

TingLei-NOAA (approved):

ShunLiu-NOAA (approved):

SamuelDegelia-NOAA (approved):

delippi (approved):
@PraveenKumar-NOAA, thanks for making those changes. I think it looks good for me now.

PR #48, opened 2024-05-23
GSI test support for Orion

delippi:
This PR is meant to help with setting up GSI experiment from staged data. Data is now staged on Orion in addition to Hera. The helper scripts were also simplified to not have one for hera and one for orion. Instead, some logic was added to each script to decide which machine it is being run on. This also means that updates are needed to the wiki to not specify ${platform} in the helper scripts names.


delippi:
I went back through the wiki to make sure everything still works with all these renamed files. I have that ready to go once we are ready to merge.

SamuelDegelia-NOAA:
Did you need to change cpus-per-task from 2 to 4 for `run_fv3jedi.sh` on Orion? I see that change for the GSI script but not the fv3jedi script that you mentioned via email. 

Also, there is now a `rrfs-test/ush/mpasjedi_increment_fulldom.py` script so we can include that in `setup_experiment.sh` like we do for the fv3jedi case. 

Other than that, things look good to go for me.

delippi:
> Did you need to change cpus-per-task from 2 to 4 for `run_fv3jedi.sh` on Orion? I see that change for the GSI script but not the fv3jedi script that you mentioned via email.
> 
> Also, there is now a `rrfs-test/ush/mpasjedi_increment_fulldom.py` script so we can include that in `setup_experiment.sh` like we do for the fv3jedi case.
> 
> Other than that, things look good to go for me.

Hi @SamuelDegelia-NOAA, thanks for the suggestions. I just changed the cpus-per-task count to 4 and added the copy of that mpas python script to the setup experiment script.

SamuelDegelia-NOAA (approved):

ShunLiu-NOAA (approved):

Issue #46, opened 2024-05-23
Assimilation of METAR observations: GSI vs JEDI


spanNOAA:
I am working on assimilating METAR observations 187 and 287 in FV3 JEDI EnVar and am observing some unusual results when comparing them to the results from the GSI EnVar analysis.

Below are the model level 64 and level 1 wind, temperature and specific humidity increments comparing between GSI (left) and JEDI EnVar using the same background and the same ensemble. The GSI is configured with horizontal localization of 110 km and vertical localization of 3 levels. And the JEDI is configured with horizontal localization of 110 km and vertical localization of 3 (but I'm not pretty sure the unit of the vertical localization, cause the JEDI documentation says "the vertical lengthscale is in vertical units of model's choice").

The major surface increment differences (@ level 64) between GSI and JEDI are in the observation locations. It is evident that most of the METAR observations used in JEDI are within CONUS, whereas many METAR observations over the marine areas are assimilated in GSI. Another notable difference is that, even with the same horizontal localization, the JEDI increment spreads less horizontally compared to the GSI increment. To determine the difference in length scales between the two systems, a single observation experiment will be necessary in the future.

![U_Increment_Level64](https://github.com/NOAA-EMC/RDASApp/assets/142531472/937f1ff0-ec0a-48df-9233-ea8d3aaa8406)
![V_Increment_Level64](https://github.com/NOAA-EMC/RDASApp/assets/142531472/839ae376-8d32-424d-9b0a-49bf785eefe5)
![T_Increment_Level64](https://github.com/NOAA-EMC/RDASApp/assets/142531472/dcdaaa16-8a05-4038-bb56-b47bddc0c27c)
![Sphum_Increment_Level64](https://github.com/NOAA-EMC/RDASApp/assets/142531472/0306a0fe-9900-4034-9172-67998b62165a)

The third issue concerns vertical localization. Below is a comparison of wind, temperature, and specific humidity increments at model level 1 (top of the model). Humidity has reasonable increments at the top of the model in both systems, which are less than 1‰ of the surface increments. However, wind and temperature increments have similar magnitudes to their surface increments in both GSI and JEDI. This issue likely requires further investigation through a single observation experiment. Additionally, more results from HofX will be posted in the near future.

![U_Increment_Level1](https://github.com/NOAA-EMC/RDASApp/assets/142531472/bfcf9204-df18-4638-91c2-1899c123ac42)
[...]

PR #53, opened 2024-06-05
add yaml rrfs_fv3jedi_bumploc_2022052619.yaml for creatiing bumploc 

TingLei-NOAA:
Added the yaml : rrfs_fv3jedi_bumploc_2022052619.yaml, which was used to create bumploc files used in the corresponding fv3-jedi hybrid analysis case.

### DIFF
--- a/rrfs-test/testinput/rrfs_fv3jedi_bumploc_2022052619.yaml
+++ b/rrfs-test/testinput/rrfs_fv3jedi_bumploc_2022052619.yaml
@@ -1,0 +1,86 @@
+geometry:
+ fms initialization:
+   namelist filename: DataFix/fmsmpp.nml
+   field table filename: DataFix/field_table
+ namelist filename: DataFix/input_lam_C775.nml
+ akbk: DataFix/fix/akbk61.nc
+ field metadata override: Data/fieldmetadata/tlei-gfs-restart.yaml
+ # input.nml
+ npz: 60
+ layout: [8,10]
+ io_layout: [1,1]
+ ntiles: 1 
+background:
+  datetime: 2022-05-26T19:00:00Z
+  filetype: fms restart
+  datapath: Data/bkg
+  filename_core: fv3_dynvars.nc
+  filename_trcr: fv3_tracer.nc
+  filename_sfcd: fv3_sfcdata.nc
+  filename_cplr: coupler.res
+  state variables: [air_temperature,surface_pressure] 
+background error:
+  covariance model: SABER
+  saber central block:
+    saber block name: BUMP_NICAS
+    calibration:
+      general:
+        universe length-scale: 250.0e3
+      io:
+        files prefix: Data/bump_new/fv3jedi_bumpparameters_nicas_lam_atm
+        alias:
+        - in code: air_temperature
+          in file: fixed_250km_0.3
+        - in code: surface_pressure
+          in file: fixed_250km
+
+      drivers:
+        multivariate strategy: univariate 
+        compute nicas: true
+        write local nicas: true
+      nicas:
+        resolution: 6
+        explicit length-scales: true
+        horizontal length-scale:
+        - groups:
+          - air_temperature
+          value: 250000.0
+        - groups:
+          - surface_pressure
+          value: 250000.0
+
+        vertical length-scale:
+        - groups:
+          - air_temperature
+          value: 0.3
+        - groups:
+          - surface_pressure
+          value: 0.0
+
+      grids:
+      - model:
+          variables:
+          - air_temperature
+      - model:
+          variables:
+          - surface_pressure
+      output model files:
+      - parameter: cor_rh
+        file:
+          filetype: fms restart
+          datapath: Data/bump_new/
+          filename_core: bumpparameters_nicas_lam_atm.cor_rh.fv_core.res.nc
+          filename_trcr: bumpparameters_nicas_lam_atm.cor_rh.fv_tracer.res.nc
+          filename_sfcd: bumpparameters_nicas_lam_atm.cor_rh.sfc_data.nc
+          filename_sfcw: bumpparameters_nicas_lam_atm.cor_rh.fv_srf_wnd.res.nc
+          filename_cplr: bumpparameters_nicas_lam_atm.cor_rh.coupler.res
+      - parameter: cor_rv
+        file:
+          filetype: fms restart
+          datapath: Data/bump_new/
+          filename_core: bumpparameters_nicas_lam_atm.cor_rv.fv_core.res.nc
+          filename_trcr: bumpparameters_nicas_lam_atm.cor_rv.fv_tracer.res.nc
+          filename_sfcd: bumpparameters_nicas_lam_atm.cor_rv.sfc_data.nc
+          filename_sfcw: bumpparameters_nicas_lam_atm.cor_rv.fv_srf_wnd.res.nc
+          filename_cplr: bumpparameters_nicas_lam_atm.cor_rv.coupler.res
+
--- a/rrfs-test/testinput/rrfs_mpasjedi_bumploc_2022052619.yaml
+++ b/rrfs-test/testinput/rrfs_mpasjedi_bumploc_2022052619.yaml
@@ -1,0 +1,40 @@
+_output config: &outputConfig
+  date: &validDate '2022-05-26T19:00:00Z'
+  stream name: control
+geometry:
+  nml_file: "./namelist.atmosphere_15km"
+  streams_file: "./streams.atmosphere_15km"
+  deallocate non-da fields: true
+  bump vunit: "height" # modellevel, height
+background:
+  state variables: &vars [spechum,surface_pressure,temperature,uReconstructMeridional,uReconstructZonal]
+  filename: "bg.2022-05-26_19.00.00.nc"
+  date: *validDate
+
+background error:
+  covariance model: SABER
+
+  saber central block:
+    saber block name: BUMP_NICAS
+    calibration:
+      io:
+        files prefix: bumploc_1000km6km
+
+      drivers:
+        multivariate strategy: duplicated
+        compute nicas: true
+        write local nicas: true
+        #write nicas grids: true
+        internal dirac test: false 
+      model:
+        level for 2d variables: last
+      nicas:
+        resolution: 8.0
+        explicit length-scales: true
+        horizontal length-scale:
+          - groups: [common]
+            value: 1000.0e3
+        vertical length-scale:
+          - groups: [common]
+            value: 6.0e3
+


ShunLiu-NOAA (dismissed):
Looks good to me.

SamuelDegelia-NOAA:
@TingLei-NOAA Are the localization radii set in this file, or does it use some default values?

delippi:
Does this run without any ob error assignment?

TingLei-NOAA:
> @TingLei-NOAA Are the localization radii set in this file, or does it use some default values?

I had replaced it with the right one:(.

TingLei-NOAA:
> Does this run without any ob error assignment?

The current one is used by  fv3jedi_error_covariance_toolbox.x to create nicas correlation files.

SamuelDegelia-NOAA:
That makes sense! Looks good to me now.

SamuelDegelia-NOAA (dismissed):

rrfs-test/testinput/rrfs_fv3jedi_bumploc_2022052619.yaml
```diff
@@ -0,0 +1,86 @@
...
+          - surface_pressure
+          value: 250000.0
+
+        vertical length-scale:
+        - groups:
+          - air_temperature
+          value: 0.3
+        - groups:
+          - surface_pressure
+          value: 0.0
```
guoqing-noaa, line 58: @TingLei-NOAA For line 58, does it mean no any vertical correlation for surface_pressure?

TingLei-NOAA: @guoqing-noaa   Good question.  I think it means no vertical correlation for surface pressure --- the 2D control variable/analysis variables (since it is a 2D variable).   So for all 3d analysis variables, if not specified specifically like air_temperature in this case, it would use the length scales for the air_temperature (3D ) group.  All other 2d control variables like sst would use ones in the surface_pressure group.

guoqing-noaa: @TingLei-NOAA Thanks for the information. 
(1) if we don't specify "air_temperature", but use humidity, will those not-specified 3D variables use the length scale for humidity? 
(2) For some 2D variables, I understand that we only want to do 2DVAR. But for surface pressure, we may want to spread vertically to some extent. When we change settings for this, will all other 2D variables change accordingly?  Thanks!

TingLei-NOAA: Guoqing,  1) Yes.   2)  here, the surface_pressure is the analysis variable , I am wondering How you would want to "spread vertically "?   For analysis using surface pressure observation,  the surface observation would spread its impact vertically to those 3d anallysis variables (since non-zero vertical localization is set up).   But, specific to your question, if you change this set up for surface pressure (analysis variable) , yes, all other 2d variable will use the changed setup ( if only one 2d group exist).  For multiple 2d group situation, we need to dig more to see what the behavior would be).

guoqing-noaa: Thanks, Ting! I will need to learn more about BUMP and NICAS. So the vertical distribution of obs innovation is through ensemble BEC, right?

If we don't have the surface_pressure group, will the 2D variables also use a default 0.0 vertical length scale?

TingLei-NOAA: > Thanks, Ting! I will need to learn more about BUMP and NICAS.
Same here!
> So the vertical distribution of obs innovation is through ensemble BEC, right?
Yes!
> 
> If we don't have the surface_pressure group, will the 2D variables also use a default 0.0 vertical length scale?
I am not sure. Will the jedi Var app abort complaining about missing group?   I need to learn about them:).

rrfs-test/testinput/rrfs_fv3jedi_bumploc_2022052619.yaml
```diff
@@ -0,0 +1,86 @@
...
+          - air_temperature
+          value: 250000.0
+        - groups:
+          - surface_pressure
+          value: 250000.0
+
+        vertical length-scale:
+        - groups:
+          - air_temperature
+          value: 0.3
```
guoqing-noaa, line 55: Line 55, @TingLei-NOAA Do you know how to set this value if we want to set the vertical length-scale to be 3 vertical levels?

TingLei-NOAA: @guoqing-noaa  I don't know if bump allows other specification method, like in the grid unit,  for the vertical scale. Maybe, you could ask the question in saber.

SamuelDegelia-NOAA: For MPAS-JEDI, the units for vertical localization can be set in the geometry section of the yaml you use for BUMP. However, I am not seeing this option in any of the code for FV3-JEDI, so I am not sure if the same functionality is available there. 

Example for MPAS-JEDI: 

```
geometry:
  nml_file: "./namelist.atmosphere_15km"
  streams_file: "./streams.atmosphere_15km"
  deallocate non-da fields: true
  bump vunit: "height" # vertical unit for vertical localization [modellevel, height]
```

guoqing-noaa: Thanks @TingLei-NOAA @SamuelDegelia-NOAA !

Can we also add a bumploc yaml file for MPAS-JEDI?

SamuelDegelia-NOAA: The bump yaml for MPAS-JEDI is currently included in the staged directory with the ensemble localization files. But it's a good idea to include it in the repository. I can add that commit later tonight.

SamuelDegelia-NOAA: I just added a bumploc yaml for mpas-jedi to this PR. Hope that worked - I haven't added a commit to another user's PR before.

TingLei-NOAA: @SamuelDegelia-NOAA  Thanks for adding this mpas-jedi bump yaml.  
@guoqing-noaa   i would like to highlight that a rather large horizontal scale 1000km is used in this yaml (and in this mpas-jedi test case in rdasapp) because a realistic length scale like 200km (with 36 mpi ranks used by this test case)  would make the bump tool fail to finish in 8 hours clock time limit.     Yongzuo is testing a new bump function ( https://github.com/JCSDA-internal/saber/pull/833.)  and hopefully to resolve this issue.

SamuelDegelia-NOAA (approved):

ShunLiu-NOAA (approved):

rrfs-test/testinput/rrfs_fv3jedi_bumploc_2022052619.yaml
```diff
@@ -0,0 +1,86 @@
+geometry:
+ fms initialization:
+   namelist filename: DataFix/fmsmpp.nml
+   field table filename: DataFix/field_table
+ namelist filename: DataFix/input_lam_C775.nml
+ akbk: DataFix/fix/akbk61.nc
+ field metadata override: Data/fieldmetadata/tlei-gfs-restart.yaml
```
guoqing-noaa, line 7: Per @spanNOAA, the default vertical coordinate for FV3 BUMP is sigma.
We can add the following line

`vert coordinate: sigma #logp, orograhpy
`
before line 8 to change to use a different type of vertical coordinate.

The relevant code is at:
https://github.com/JCSDA/fv3-jedi/blob/e9175cbf268f82c9b276601b4ea4e0061df81426/src/fv3jedi/Geometry/GeometryParameters.h#L64

guoqing-noaa: @TingLei @SamuelDegelia-NOAA 
I think we may want to add this line to this file. Thanks!

SamuelDegelia-NOAA: Thanks for finding this @guoqing-noaa and @spanNOAA! Though, if the default vertical coordinate is sigma, then I wonder why the profile I plotted [here](https://github.com/NOAA-EMC/RDASApp/issues/46#issuecomment-2148391801) shows increments up to ~800 hPa despite the yaml assigning a vertical length scale of 0.3.

TingLei-NOAA: @guoqing-noaa  I suspect this part of parameters are specifically  for the session defined in the fv3 namelist input as :"namelist filename: DataFix/input_lam_C775.nml"  .    not about the vertical coord for the localization scales.    You can try adding that parameter in the saber session to see what will happen.

guoqing-noaa: > Thanks for finding this @guoqing-noaa and @spanNOAA! Though, if the default vertical coordinate is sigma, then I wonder why the profile I plotted [here](https://github.com/NOAA-EMC/RDASApp/issues/46#issuecomment-2148391801) shows increments up to ~800 hPa despite the yaml assigning a vertical length scale of 0.3.

The figure showed that there are still some small increments around 700 hPa which corresponds to 0.3 sigma cutoff radius.

guoqing-noaa: @TingLei-NOAA Thanks for the input. We will try it.

SamuelDegelia-NOAA: > > Thanks for finding this @guoqing-noaa and @spanNOAA! Though, if the default vertical coordinate is sigma, then I wonder why the profile I plotted [here](https://github.com/NOAA-EMC/RDASApp/issues/46#issuecomment-2148391801) shows increments up to ~800 hPa despite the yaml assigning a vertical length scale of 0.3.
> 
> 
> 
> The figure showed that there are still some small increments around 700 hPa which corresponds to 0.3 sigma cutoff radius.
> 
> 

Ah, okay I see. I was thinking in terms of "sigma levels" previously.

spanNOAA: > @guoqing-noaa I suspect this part of parameters are specifically for the session defined in the fv3 namelist input as :"namelist filename: DataFix/input_lam_C775.nml" . not about the vertical coord for the localization scales. You can try adding that parameter in the saber session to see what will happen.

The vertical coordinate parameter should be applicable to the bump feature as well. This is supported by the fact that altering the vertical coordinate from sigma to logp also results in a change in the sublevels defined within the bump, as evidenced by the differences observed in the log file. 
if I set the vertical length scale as 0.2 and the vertical coordinate as sigma, I have a total of 33 effective levels. However, upon switching the vertical coordinate to logp while maintaining the same length scale, the total effective levels increases to 47.
I'll share some figures once I've got more results.

TingLei-NOAA: @spanNOAA  Thanks for you and @guoqing-noaa 's exploration in this aspect which help me have a better understanding of this issue. 
The below is what I understand now.
1) the "vert coordinate" is the parameter in fv3-jedi geometry session. It could be set up as in 3dvar_snow.yaml. 
but itself  is not used for saber/bump and would also not be passed to saber. (and , hence, we add "vert coordinate"  in saber's "vertical lengthscale" session would not cause  any impacts if not causing any unrecognized parameter errors.  
2)  The vert coordinate in geometry parameter would cause the vertical coordinates different (though I am not sure if it could be setup freely by the users),) and ,hence, the model vertical coordinates will be inherited/used by saber .  At the same time, the "vertical scale" value set up in the saber session will be used in bump in the same way. Hence, whether the model is in sigma or logp would cause different results, even though the vertical length scale is used with the same value.    
3)  I thought , by default, the "vertical length scale"  value is for logp (by default).  But following what you and Guoqing had pointed to,  the right way to describe it is that it is in the model vertical coordinates, for fv3-jedi,  which would be decided by "vert coordinates)" parameters (by default, it is sigma, this needs to be double checked). 
4) Similarly, we can further explore how units of  model vertical coordinates in mpas jedi is decided.

PR #61, opened 2024-06-07
update setup_experiment and related scripts for jet

delippi:
This PR is just meant to update the setup_experiment and related scripts in rrfs-test. This script is a good place for developers to start setting up their test case.

### DIFF
--- a/rrfs-test/scripts/templates/run_mpasjedi_template.sh
+++ b/rrfs-test/scripts/templates/run_mpasjedi_template.sh
@@ -18,6 +18,8 @@ if [[ $hostname == "h" ]]; then
   platform="hera"
 elif [[ $hostname == "o" ]]; then
   platform="orion"
+elif [[ $hostname == "f" ]]; then
+  platform="jet"
 fi
 
 module use @YOUR_PATH_TO_RDASAPP@/modulefiles
--- a/rrfs-test/scripts/templates/run_mpasjedi_template.sh
+++ b/rrfs-test/scripts/templates/run_mpasjedi_template.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-#SBATCH --account=fv3-cam
+#SBATCH --account=@SLURM_ACCOUNT@
 #SBATCH --qos=debug
 #SBATCH --ntasks=36
 #SBATCH -t 00:30:00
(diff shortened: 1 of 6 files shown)


ShunLiu-NOAA (dismissed):

rrfs-test/scripts/templates/run_mpasjedi_template.sh
```diff
@@ -18,6 +18,8 @@ if [[ $hostname == "h" ]]; then
   platform="hera"
 elif [[ $hostname == "o" ]]; then
   platform="orion"
+elif [[ $hostname == "f" ]]; then
+  platform="jet"
```
guoqing-noaa: FYI, at the computing nodes, the hostname does not necessarily return strings leading with "f", "h" etc.

guoqing-noaa (dismissed):
LGTM. Thanks!

rrfs-test/scripts/templates/run_mpasjedi_template.sh
delippi: What do you suggest this should be?

guoqing-noaa: At this moment, we may use a similar method as what is used by global workflow and rrfs-workflow, i.e. to check specific folders. Here is an example:
https://github.com/NOAA-EMC/rrfs-workflow/blob/dev-sci/ush/fix_rrfs_locations.sh

Hercules and Orion share the same file system but I recalled that `hostname` at their computing nodes returns distinguishable strings between them.

guoqing-noaa: ok, we already have a powerful script in RDASApp:
`RDASApp/ush/detect_machine.sh`

delippi: Yep, I just found that and was going to use it. Thank you!

delippi: That makes a lot of these scripts a lot simpler. Thanks for the suggestion.

ShunLiu-NOAA: @delippi Have you completed the test on all three machines?

delippi: Give me a moment and I will confirm on all three.

delippi: @ShunLiu-NOAA, everything still works on Orion and Hera. I had a strange failure on Jet that I need to figure out.

delippi: @ShunLiu-NOAA, it seems to be working now on Jet as well. I ***think*** the issue was one of Jet's files systems (lsf4) was having issues and something in module load was trying to reach something staged there.

guoqing-noaa: @delippi  You are right. /lfs4 had performance issues in the past few days and Admin took Jet down to fix it. Fortunately, it was successfully fixed yesterday.

guoqing-noaa (approved):

ShunLiu-NOAA:
@delippi Is there any other update needed for this PR?

delippi:
Hi @ShunLiu-NOAA, it should be all set.

ShunLiu-NOAA (approved):

Issue #62, opened 2024-06-07
how to update wiki?


guoqing-noaa:
Is there a way to update the wiki by the community? Can we track the modification history?

If NOT, we may want to create a `doc/` subdirectory to hold RDASApp-related documents, including a quick Readme.md.
This way, everyone may contribute to the documentation and a review process can be enforced. Thanks!

SamuelDegelia-NOAA:
You can see the modification history for the wiki by clicking the "XX revisions" link at the top of any of the pages. But as far as I know there is no way to have PRs for those changes. I think a quick README.md file would be helpful, at least with build instructions. But I think some of the more detailed instructions, like for the one-ob test, are better left for the wiki. 

To update the wiki, you should just be able to hit the edit page at the top. But maybe you need permissions for this?

guoqing-noaa:
Thanks! @SamuelDegelia-NOAA 
I think some contents in the documentation are tied with specific commits of the repo. Those documents will be better to include as one part of the repo. Wiki may hold some general things regardless of the changed versions.

No, I cannot see any "edit" buttons/links.

SamuelDegelia-NOAA:
@CoryMartin-NOAA Could you add Guoqing as a repository collaborator so that he can edit the wiki? 

Also, I agree that the wiki documentation changes frequently so it would be nice to track/approve those changes as part of the repo. But I also think that the layout on the wiki pages makes it much easier to follow a tutorial like for the one-ob test. So I'll do a little more digging on if there is a way to commit to the wiki.

CoryMartin-NOAA:
@guoqing-noaa see if you have access now

guoqing-noaa:
Thanks @CoryMartin-NOAA! I saw the "edit" button now.

SamuelDegelia-NOAA:
FYI I found out about this GitHub Action (https://github.com/marketplace/actions/github-wiki-action) that can sync files between the Wiki and repository (e.g., files in `docs/`). This allows you to commit and have PRs for the wiki, and could be an option in the future if we wanted.

guoqing-noaa:
That's great news! Thanks, @SamuelDegelia-NOAA 
We should try this.

SamuelDegelia-NOAA:
I can try and get it working on my fork and report back.

PR #63, opened 2024-06-08
Simplify the build.sh command line for normal use

guoqing-noaa:
Currently, to build RDASApp, one will need to enter the command similar as follows:
`./build.sh -f -a -d -r -t $platform`

The PR is to change the default behavior so that one may only need to enter a much-simplified command:
`./build.sh`

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -69,9 +69,8 @@ while getopts "p:t:c:m:hvfs-:" opt; do
       ;;
     -)
       if [[ "${OPTARG}" == "notestdata" ]]; then
-        LINK_NEST=ON
-        CLONE_RRFSDATA=NO
         CLONE_JCSDADATA=NO
+        CLONE_RRFSDATA=NO
       fi 
       ;;
     h|\?|:)
--- a/build.sh
+++ b/build.sh
@@ -22,10 +22,9 @@ usage() {
   echo "  -c  additional CMake options        DEFAULT: <none>"
   echo "  -v  build with verbose output       DEFAULT: NO"
   echo "  -f  force a clean build             DEFAULT: NO"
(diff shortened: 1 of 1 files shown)


SamuelDegelia-NOAA (commented):
Overall looks good to me. Since we are getting `$MACHINE_ID` from the `ush/detech_machine.sh` script, should we just go ahead and remove the `-t` build option?

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for your review!

I agree with you that it will be safe to remove the -t option. However, when we were working on a similar issue for rrfs-workflow (issue https://github.com/NOAA-EMC/rrfs-workflow/issues/81), chan-hoo would like to keep the possibility to use the -t option. I did not know that underlying reason but it does no harm to keep -t either.

SamuelDegelia-NOAA:
Okay, if there is a reason to keep it then its no problem!

SamuelDegelia-NOAA (approved):

delippi (approved):
I think these changes will be nice. @guoqing-noaa, when this PR is merged, make sure to update the wiki with the correct build command.

guoqing-noaa:
> I think these changes will be nice. @guoqing-noaa, when this PR is merged, make sure to update the wiki with the correct build command.

Thanks, @delippi ! As we discussed in issue #62 , the build instruction will be tied with repo versions. So I plan to add docs/ subdirectories for holding commit-specific documents while wiki for general ones. I can create a PR to continue our discussions there.

ShunLiu-NOAA:
@guoqing-noaa We can merge this PR to _develop_ first. But updating wiki to match this version is also necessary at the moment.

Issue #65, opened 2024-06-09
git status --ignore-submodules


guoqing-noaa:
I am not sure whether this is an issue for others. Or it might already be documented.

`git status` is substantially slower sometimes for this repo as it contains 17 submodules. 

By default, when you run `git status` in a repo with submodules, it'll recursively run `git status `in every submodule (sequentially). `git diff` behaves similarly to `git status`. Pre-commit appears to run` git status` once at the beginning and then runs `git diff` before and after each hook (to detect any changes made by the hook). All these git operations will sometimes become substantially slower for a repo with many submodules.

Since we most likely will not update submodule contents while working on RDASApp, we can add a command line option `--ignore-submodules` to make normal git operations much faster.  However, it will be sweeter if there is a way to make this the default option.

TingLei-NOAA:
@guoqing-noaa  Good point!  However, will this be done by developers themselves like by setting up their own alias after this issue is documented ,say, in the wiki?     
Another way is to have a setup in rdas module file to add an alias like status_v1 , but to let the developers know it is not straightforward  and maybe it also need to be done by the wiki.   But, then, I will prefer the users/developers add the alias by themselves.

TingLei-NOAA:
And, another way is not to run "git status" in the root dir and just run it in the component the developer is interested in. I think in that case, only "status" in the current project/submodule would be checked out.

SamuelDegelia-NOAA:
The behavior @guoqing-noaa describes can be made default by adding an `ignore = dirty` line in the `.gitmodules` file like so: 

```
[submodule "sorc/crtm"]
        path = sorc/crtm
        url = https://github.com/jcsda/crtm
        branch = release/crtm_jedi_v2.4.1
        ignore = dirty
```

This speeds up `git status` significantly. You could also choose `ignore = all` to suppress all output and comparisons for the submodules altogether. See the difference of all and dirty [here](https://stackoverflow.com/a/41598706). 

Whether we want to do this is up for discussion though. I think a few people might be using the base RDASApp to also make code changes to some of the submodules like UFO, fv3-jedi, mpas-jedi etc. This change might make it harder to track those developments and in that case just recommending an alias via the wiki could be better. But I welcome any input there.

TingLei-NOAA:
@SamuelDegelia-NOAA  According to chatgpt (:)) :
```
[...]

PR #75, opened 2024-06-12
update build_and_test.md

guoqing-noaa:


ShunLiu-NOAA (approved):

docs/build_and_test.md
```diff
@@ -1,32 +1,23 @@
 ## 1.Clone RDASApp
-In the directory that you want to install RDASApp:  
 ```
 git clone --recurse-submodules https://github.com/NOAA-EMC/RDASApp.git
```
delippi, line 3: I wonder if we should add this line here? If not just mention this command and suggest it *could* be used here.
```
git submodule update --init --remote
```
This will update all the submodules to the most recent instead of using the versions that haven't been updated in awhile. Although, I think we will start updating submodules more frequently in the future. Otherwise it looks good to me.

delippi (approved):

TingLei-NOAA (approved):

docs/build_and_test.md
guoqing-noaa: @delippi Thanks for your review! 
As discussed in the issue https://github.com/NOAA-EMC/RDASApp/issues/65 and agreed in a small group tag up, we will use snapshots of submodules (instead of moving heads) for each RDASApp commit. If a submodule needs a new version, a PR should be created in RDASApp to update it.  @ShunLiu-NOAA and @hu5970 have been working on a draft for the code management strategy and will share it with us soon.

Issue #79, opened 2024-06-13
Clean up executable assignment in RDASApp/rrfs-test/CMakeLists.txt


SamuelDegelia-NOAA:
I am planning to come up with a compact way to deal with the increasingly long list of executables in `RDASApp/rrfs-test/CMakeLists.txt`.

SamuelDegelia-NOAA:
CMake doesn't directly support dictionary data structures but you can get around it by using a list and then splitting elements via regex. @TingLei-NOAA, is the this the sort of solution you were looking for? 

Instead of: 

```
      # JEDI test names (yaml files)
      list( APPEND rrfs_fv3jedi_tests
             rrfs_fv3jedi_hyb_2022052619
             rrfs_fv3jedi_letkf_2022052619
          )

      # JEDI executables for each test case above
      list ( APPEND rrfs_fv3jedi_exes
             fv3jedi_var.x
             fv3jedi_letkf.x
      )
```

We can combine both lists into a more compact "dict-like" structure as in: 

      set( rrfs_fv3jedi_tests
           "rrfs_fv3jedi_hyb_2022052619=fv3jedi_var.x"
           "rrfs_fv3jedi_letkf_2022052619=fv3jedi_letkf.x"
      )

And then we can retrieve the keys/pairs through: 

```
      foreach(keypair ${rrfs_fv3jedi_tests})
         string(REGEX MATCH "([^=]+)=([^=]*)" _ ${keypair})
         set(case ${CMAKE_MATCH_1})
         set(exe ${CMAKE_MATCH_2})
      endforeach()
```

TingLei-NOAA:
@SamuelDegelia-NOAA  Sure. this version  would make this step more readable and less prone to errors.   But, i would prefer a more "generic"/friendlier interface with a cost of a few additional functions. 
Would you please have a look at chatgpt "product" :  chatgpt-dictionary-mimic.sh  in hera:/home/Ting.Lei/dr-out/.  And let you know what your opinions are.

SamuelDegelia-NOAA:
@TingLei-NOAA Yes that way looks good too and more user-friendly. BTW, what kind of message do you type in to ChatGPT to get something like this?

TingLei-NOAA:
@SamuelDegelia-NOAA  Interesting question.  I checked the records:).   My prompts were as 
[...]

PR #81, opened 2024-06-14
Update rrfs-test/CMakeLists.txt to be more compact for adding a CTest

SamuelDegelia-NOAA:
This PR updates `RDASApp/rrfs-test/CMakeLists.txt` to be a little more user-friendly by making it easier to add CTests. Two new functions are added to allow for a “dictionary-like” data structure in CMake. To add a new CTest now, you just need to add a single line to either the fv3-jedi or mpas-jedi lists at the top of the file: 

```
### Define CTests here ###

# FV3-JEDI tests
set(rrfs_fv3jedi_tests "")
add_to_dictionary(rrfs_fv3jedi_tests "rrfs_fv3jedi_hyb_2022052619"     "fv3jedi_var.x")
add_to_dictionary(rrfs_fv3jedi_tests "rrfs_fv3jedi_letkf_2022052619"   "fv3jedi_letkf.x")

# MPAS-JEDI tests
set(rrfs_mpasjedi_tests "")
add_to_dictionary(rrfs_mpasjedi_tests "rrfs_mpasjedi_2022052619_Ens3Dvar"     "mpasjedi_variational.x")
add_to_dictionary(rrfs_mpasjedi_tests "rrfs_mpasjedi_2022052619_atms_npp_qc"  "mpasjedi_variational.x")
add_to_dictionary(rrfs_mpasjedi_tests "rrfs_mpasjedi_2022052619_letkf"        "mpasjedi_enkf.x")
```

All CTests in `RDASApp/build/rrfs-test` pass after making this change. 

Also, per discussion in #79, we will likely move to one mega executable in the near future (i.e., just a single `rdas_jedi.x` instead of `fv3jedi_var.x`, `fv3jedi_letkf.x`, `mpasjedi_enkf.x`, etc.). `CMakeLists.txt` will need to be updated again at that point, but we should be able to keep the same new code structure that is added in this PR. It will eventually look something like: 

`add_to_dictionary(rrfs_tests “rrfs_fv3jedi_hyb_2022052619” “fv3jedi” “variational”)
`


rrfs-test/CMakeLists.txt
```diff
@@ -2,6 +2,72 @@
 #!    NOAA/NCEP, National Centers for Environmental Prediction  RDASapp        !
 #!-------------------------------------------------------------------------
 
+###########################
+### Begin function list ###
```
TingLei-NOAA: @SamuelDegelia-NOAA   Could  we put those functions definition part to a separate include file ( I guess there would be more functions to be added in the future) and then, to be "included" into this file?

SamuelDegelia-NOAA: Good idea, I can do that.

SamuelDegelia-NOAA: @TingLei-NOAA How about this? I just put the `dict_funcs.cmake` inside `RDASApp/rrfs-test/` but I'd be fine with it going somewhere else.

guoqing-noaa: Thank @SamuelDegelia-NOAA and @TingLei-NOAA for working on this.

If I understand correctly, I think Ting means a file named like "funcs.cmake" instead of "dict_funcs.cmake" which is limited to dict functions only.

rrfs-test/CMakeLists.txt
```diff
@@ -73,25 +77,12 @@ if (CLONE_RRFSDATA)
...
-      # JEDI executables for each test case above
-      list ( APPEND rrfs_mpasjedi_exes
-             mpasjedi_variational.x
-             mpasjedi_enkf.x
-             mpasjedi_variational.x
-      )
-
-      foreach(case IN LISTS rrfs_mpasjedi_tests)
-
+      foreach(keypair ${rrfs_mpasjedi_tests})
```
guoqing-noaa: Should we use `get_from_dictionary `here?

rrfs-test/CMakeLists.txt
```diff
@@ -22,22 +38,12 @@ if (CLONE_RRFSDATA)
...
-
-      # JEDI executables for each test case above
-      list ( APPEND rrfs_fv3jedi_exes
-             fv3jedi_var.x
-             fv3jedi_letkf.x
-      )
-
-      foreach(case IN LISTS rrfs_fv3jedi_tests)
+      foreach(keypair ${rrfs_fv3jedi_tests})
+	 string(REGEX MATCH "([^=]+)=([^=]*)" match ${keypair})
```
guoqing-noaa: Should we use get_from_dictionary here?

rrfs-test/CMakeLists.txt
SamuelDegelia-NOAA: We could, but the `get_from_dictionary` function requires you to pass in the key name (test case name in this case). And in order to get the key name from the dict that is being looped over, I needed to use the `string(REGEX MATCH "([^=]+)=([^=]*)" _ ${keypair})` line. This command returns both the key and the value as `${CMAKE_MATCH_1}` and `${CMAKE_MATCH_2}` so I just went ahead and used it instead of calling the `get_from_dictionary` function. But I left that function in `funcs.cmake` for completeness and in case we need it later.

rrfs-test/CMakeLists.txt
SamuelDegelia-NOAA: Good point, `funcs.cmake` is more generic if we need to add more functions later on. I added new commits that renamed the function file.

rrfs-test/CMakeLists.txt
TingLei-NOAA: I think we could try a little different use of this dictionary as I would add in the comment with a label #dic1

rrfs-test/CMakeLists.txt
```diff
@@ -2,6 +2,22 @@
 #!    NOAA/NCEP, National Centers for Environmental Prediction  RDASapp        !
 #!-------------------------------------------------------------------------
 
+include(funcs.cmake)
+
+### Define CTests here ###
+
+# FV3-JEDI tests
+set(rrfs_fv3jedi_tests "")
```
TingLei-NOAA: #dic1
set (rrfs_tests_execs)
add_to_dictionary(rrfs_tests_execs  "rrfs_fv3jedi_hyb_2022052619"  "fv3jedi_var.x") 

and all the similar to the same. 
In fact, to add corresponding "add_to_dictionary(rrfs_tests_execs, <test_name>, <test_exec>) " should be done by the developer who is adding a new test

rrfs-test/CMakeLists.txt
```diff
@@ -22,22 +38,10 @@ if (CLONE_RRFSDATA)
          set(MPI_ARGS "${MPI_ARGS} --exclusive")
       endif()   
 
-      # JEDI test cases
-      # ----------------
-      # NOTE: To add a new ctest for FV3, simple add the new yaml file here (without the extension)
-      #       Also need to add the associated executable to the rrfs_fv3jedi_exes list below 
-      list( APPEND rrfs_fv3jedi_tests
```
TingLei-NOAA, line 29: #dic1,
This part can be the same, no need to change.

rrfs-test/CMakeLists.txt
```diff
@@ -22,22 +38,10 @@ if (CLONE_RRFSDATA)
...
-          )
-
-      # JEDI executables for each test case above
-      list ( APPEND rrfs_fv3jedi_exes
-             fv3jedi_var.x
-             fv3jedi_letkf.x
-      )
-
-      foreach(case IN LISTS rrfs_fv3jedi_tests)
+      foreach(keypair ${rrfs_fv3jedi_tests})
```
TingLei-NOAA: #dic1
Then, in this part, all are the same as the version prior to this PR.  Only need to change the line 44 as
```
set (exe $get_from_dictionary(rrfs_tests_execs($case)) ! the exact syntax needs check. 

```
Will this way be good to make use of the dictionary-like structure you added?

rrfs-test/CMakeLists.txt
SamuelDegelia-NOAA: So just to confirm @TingLei-NOAA, you are recommending replacing: 

`set(exe ${CMAKE_MATCH_2})`

with

`get_from_dictionary(rrfs_fv3jedi_tests case exe)` 

There is no `set` call needed here since the `get_from_dictionary` function already assigns the output to `exe`. Was there another change there or is that it? I suppose that is good since it shows how to use the new get_from_dictionary function even if there is no functional difference in the result. But either way works for me.

TingLei-NOAA: Sam,   Basically, I suggest we don't need to use " foreach(keypair ${rrfs_mpasjedi_tests})"  with rrfs_mpasjedi/fv3jedi_tests being converted to dic-like structure. 
The added dic-like structure could be like  rrfs_tests_execs as I used in #dic1 comment.  we just need set it up with the key is the ctest name and the value is the corresponding executable.   By doing this, we don't need extra processing as the line 41, 42. I feel this way has a better readability by using the dic-like structure introduced in your PR.

SamuelDegelia-NOAA: Sorry Ting, I don't think I understand. How do we iterate through the dict if we do not use `foreach(keypair ${rrfs_mpasjedi_tests})`? And isn't the code I added to `CMakeLists.txt` setting up the dict just as you describe where the key is the ctest name and the value is the executable? I might be missing something here. 

```
set(rrfs_fv3jedi_tests)
add_to_dictionary(rrfs_fv3jedi_tests "rrfs_fv3jedi_hyb_2022052619"     "fv3jedi_var.x")
add_to_dictionary(rrfs_fv3jedi_tests "rrfs_fv3jedi_letkf_2022052619"   "fv3jedi_letkf.x")
```

TingLei-NOAA: Sam, one thing is that in current version, the rrfs_fv3/mpas_jedi_tests  are converted to dic-like structures, which make the following line 80 is needed since the former are not a list again.  In my opinion, by using a new dic-like variable would make the changes simpler.     I could make an example on top of your codes as soon as possible.

SamuelDegelia-NOAA: Yes maybe it would be better for you to show me an example so I understand a little better.

TingLei-NOAA: @SamuelDegelia-NOAA I created a draft PR to show what I referred to. https://github.com/NOAA-EMC/RDASApp/pull/82.    Hopefully, this is simpler/easier to understand.  
You could have a different version following your original plan by let the dict-like rrfs_fv3jedi_test to have more methods like the one to  iterate over elements .  The key is that 1), it is better we don't need extra pattern processing (on pair) in the CMakefile.txt, it is better to be hidden in the function level, 2) the structure like rrfs_fv3jedi_tests is better to be able to provide a method to give a list of ctest.   That is my two cents.  Thanks.

SamuelDegelia-NOAA: @TingLei-NOAA so in your version you use a combination of lists and dicts? I see you loop through rrfs_fv3jedi_tests (containing the ctest names), and then use the ctest name as the key to retrieve the value (exe name) in the dict? If someone were adding a new ctest, wouldn't this mean that they need to add the new test both to the list and to the dict? 

```
     list( APPEND rrfs_fv3jedi_tests
             rrfs_fv3jedi_hyb_2022052619
             rrfs_fv3jedi_letkf_2022052619
             NEW_CTEST_NAME
          )
      add_to_dictionary(rrfs_test_execs "rrfs_fv3jedi_hyb_2022052619"     "fv3jedi_var.x")
      add_to_dictionary(rrfs_test_execs "rrfs_fv3jedi_letkf_2022052619"   "fv3jedi_letkf.x")
      add_to_dictionary(rrfs_test_execs "NEW_CTEST_NAME"                    "NEW_EXE.x")
```

I thought the goal of this PR was to be more concise on adding a new CTest so that you wouldn't need to add to both the test name lists and the exe list. In my version, all you need to do to add a new ctest is add one line at the top of the file: 

`add_to_dictionary(rrfs_fv3jedi_tests "NEW_CTEST_NAME"     "NEW_EXE.x")`

I understand not wanting to include the `string(REGEX MATCH "([^=]+)=([^=]*)" _ ${keypair})` and instead keep that at the function level, but it still seems simpler from my perspective.  If this is the route we want to go though, I am okay with it.

guoqing-noaa:
I would think we probably do "over-engineering" here. 
@SamuelDegelia-NOAA 's first solution in the issue https://github.com/NOAA-EMC/RDASApp/issues/79 without a function is enough to handle the problem and is easy to read by any developers.  More importantly, if we decide to use one single executable, this change will be replaced soon. I would think investing efforts in one single executable will be more valuable. My two cents.

rrfs-test/CMakeLists.txt
TingLei-NOAA: @SamuelDegelia-NOAA  Yes, in my version,  the developer need to add their test to the test lists first.  Normally, I prefer readability the first.  As I said, your method with only one dic-like structure would be more advantageous , from my point of view, if  an iterator method could be added to your dic-like structure data  while the pattern processing on "pair" is not needed.  Basically, such a pattern processing over a pair from the dic make some loss of the expected readability and simplicity introducing the dic-like structure (when the pattern processing is difficult to understand)

TingLei-NOAA:
> I would think we probably do "over-engineering" here. @SamuelDegelia-NOAA 's first solution in the issue #79 without a function is enough to handle the problem and is easy to read by any developers. More importantly, if we decide to use one single executable, this change will be replaced soon. I would think investing efforts in one single executable will be more valuable. My two cents.

I am really wondering when there would be a super executable to be realized ,and I also have some different opinions on such a practice.  Second, even with such one exec being realized,  I expect  it needs an command  option to tell what run it is . While that kind of options are , as Sam mentioned, treated with the help of the introduced dic-like data structure.

rrfs-test/CMakeLists.txt
SamuelDegelia-NOAA: Okay, so the primary issue then is with the REGEX statement. I can see if I can move that into `funcs.cmake` somehow.

TingLei-NOAA: Sam, Yes,  with your strategy, only a method to get all the keys ( rather than we have to get those pair and do processing again) would be enough .   From the chatgpt:).  Such a function could be like 
```
function(get_all_keys dict result)
    set(keys "")
    foreach(entry IN LISTS ${dict})
        string(FIND "${entry}" "=" EQUAL_POS)
        if(EQUAL_POS GREATER -1)
            string(SUBSTRING "${entry}" 0 ${EQUAL_POS} ENTRY_KEY)
            list(APPEND keys "${ENTRY_KEY}")
        endif()
    endforeach()
    set(${result} "${keys}" PARENT_SCOPE)
endfunction()
```.
Now, i think the most perferable way is your current one with an enhancement similar to the above one.

SamuelDegelia-NOAA: Okay I'll work on adding that in on Monday. I agree with @guoqing-noaa that we are likely over-engineering this problem, so hopefully things will be good to go after this change.

SamuelDegelia-NOAA:
I updated this PR to now use the function `get_all_keys` instead of regex. Hopefully this is good to go now.

TingLei-NOAA (approved):
@SamuelDegelia-NOAA  the current version looks working cleanly with much improved readability by using the dic-like structure.    Thanks.

guoqing-noaa (approved):

ShunLiu-NOAA (approved):

Issue #51, opened 2024-05-30
Problems with halo observation distribution for regional LETKF


SamuelDegelia-NOAA:
I am experiencing an issue with the regional LETKF crashing when using the Halo observation distribution. This first post is just background on the issue, then I will discuss potential causes/solutions in a follow-up post. Here is the error message: 

```
 0: MSONET airTemperature nlocs = 312314, nobs = 298824, min = 0, max = 19, avg = 7
 0: MSONET dewPointTemperature nlocs = 312314, nobs = 298824, min = 29, max = 29, avg = 29
 0: MSONET specificHumidity nlocs = 312314, nobs = 298824, min = 29, max = 29, avg = 29
 0: MSONET stationPressure nlocs = 312314, nobs = 298824, min = 29, max = 29, avg = 29
 0: MSONET virtualTemperature nlocs = 312314, nobs = 298824, min = 29, max = 29, avg = 29
 0: MSONET windEastward nlocs = 312314, nobs = 298824, min = 29, max = 29, avg = 29
 0: MSONET windNorthward nlocs = 312314, nobs = 298824, min = 29, max = 29, avg = 29
 1: Assertion failed: numRecognizedFlags == gnlocs in print, line 195 of /work/noaa/wrfruc/sdegelia/RDASApp/bundle/ufo/src/ufo/filters/QCmanager.cc
 3: Assertion failed: numRecognizedFlags == gnlocs in print, line 195 of /work/noaa/wrfruc/sdegelia/RDASApp/bundle/ufo/src/ufo/filters/QCmanager.cc
22: Assertion failed: numRecognizedFlags == gnlocs in print, line 195 of /work/noaa/wrfruc/sdegelia/RDASApp/bundle/ufo/src/ufo/filters/QCmanager.cc
```

This error occurs when using the full set of RAP mesonet observations created from bufr2ioda. We did not experience this crash using the smaller set of observations that are included as part of the LETKF ctest in RDASApp. Also, this was tested using FV3-JEDI, but I would suspect that the error also occurs with MPAS-JEDI. 

[...]

PR #85, opened 2024-06-18
Add a simple offline domain check for IODA observations to work with halo obs. distribution

SamuelDegelia-NOAA:
This PR adds a simple tool under `RDASApp/rrfs-test/IODA/offline_domaincheck.py` to allow for the assimilation of the RAP observations created from `bufr2ioda.x` with halo observation distribution. The tool takes in a list of IODA files, creates a new file with the `_dc` prefix standing for "domain check", and then fills the new file in with any observations inside the specified region. This allows the regional LETKF to work with the `bufr2ioda.x` observations. 

As discussed previous, this is just a temporary workaround for #51 until the bug in IODA or UFO can be resolved, or until the `ReduceObsSpace` filter action is fully working for our test case.  

I did not include any scripts for running this, but all you need to do is edit the top of `offline_domaincheck.py` to give a list of the input IODA files and then the latitude/longitude ranges for the simple domain check. The example `lat_range` and `lon_range` in the script work for the fv3-jedi test case in RDASApp and allow LETKF to successfully run with the IODA observations.  These ranges would need to change whenever we get the MPAS test domain configured, but hopefully the underlying issue will be resolved by then.

### DIFF
--- a/rrfs-test/IODA/offline_domaincheck.py
+++ b/rrfs-test/IODA/offline_domaincheck.py
@@ -1,0 +1,74 @@
+import numpy as np 
+from netCDF4 import Dataset
+import os 
+
+###########################################################
+# This code performs a very simple domain check for IODA 
+# observations output from bufr2ioda.x based on a range 
+# of latitude and longitudes. Currently, the online domain
+# check in JEDI/UFO keeps obs. in memory which causes various
+# issues including when using halo obs. distribution for 
+# a regional LETKF. 
+# 
+# Note: the output file from this code will be appended with
+# the "_dc" filter standing for "domain-check". 
+###########################################################
+
+### Settings ###
+filelist = ['rap.t19z.prepbufr.MSONET.tm00.nc']
+lat_range = [21.0, 53.0]
+lon_range = [-133.0, -60.0]
+
+### Begin code ###
+for infile in filelist: 
+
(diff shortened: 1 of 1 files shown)


TingLei-NOAA (approved):

guoqing-noaa (approved):
Nice tool! Looks good to me.

Did you check the obs distribution before and after this domain check?

SamuelDegelia-NOAA:
> Nice tool! Looks good to me.
> 
> Did you check the obs distribution before and after this domain check?

Good question - I should have shown that in the PR. Here are the observations (1) before and (2) after running `offline_domaincheck.py`. 

1. Before 
dist_obs

2. After
dist_obs_dc

The tool is working as intended by removing obs far away from the domain. The obs just outside the domain (e.g., southwest Canada or Mexico) are not a problem since they are included in halo regions as long as the halo radius isn't very small.

ShunLiu-NOAA (approved):

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for the figures! Nice!

It looks like the halo radius is not small in Fig. 2. Do you know what the radius is? Thanks!

SamuelDegelia-NOAA:
The halo radius is set to 1250km which is sufficient to capture those obs just outside the domain

Issue #80, opened 2024-06-13
Incorporate Mesonet DA into JEDI for RRFS


delippi:
The UFO in JEDI is the component that not only computes model-simulated observations, but also houses filters and methods for observation QC, ob errors, and bias correction. The GSI observer is the equivalent. Many forward operators for various observations have been developed for the UFO. These operators can be utilized in RDAS. However these operators still must be tested for RDAS. The steps for transition by observation platform are as follows:

1. Establish IODA processing
2. Establish operator in UFO
3. Establish YAML input file
4. Validate with test assimilation experiments
5. Create ctest, if necessary

Mesonet data is bufr obtype=188/288 (and also 195/295 but not sure if that data is/will be used).

Edit:
A link to my GSI fork/branch based on Andrew and Emily's ufo_geovals branch. I added some extra geovals for output.
https://github.com/delippi/GSI/tree/feature/ufo_geovals_rdas

delippi:
I've finished hofx validation for mesonet airTemperature. In this example, I'm using a version of GSI (from Andrew) that writes out GSI geovals and a modified version of the ufo-vertInterp code (based on a branch from Dan who modified the gnssro operator to be able to read in GSI geovals using the fv3jedi executable--I did the same thing but for the vertInterp operator). I also had to write a short python script to update my GSI-geovals files to have saturation_specific_humidity and surface_geometric_height. The former was calculated from Rogers and Yau eq 2.19. The latter just needed to grab only the lowest model level for surface_geometric_height from the geometric_height field. The following is the result for airTemperature hofx validation. More work will be done on other variables. Then once all hofx validation is done I will move to observation error and other QC checks.
![gsi_hofx_vs_jedi_hofx_MSONET_hofxs_airTemperature_2022052619_mesonet_airTemperature](https://github.com/NOAA-EMC/RDASApp/assets/31075260/e1611d8c-3085-4bd3-ba48-85e4194ff15f)

If the GSI-geovals are not read in this is the following result. I'm thinking that this could be considered as "close enough" when ***NOT*** using GSI-geovals. I believe the differences are due to how GSI and JEDI compute geovals (i.e., 2D interpolation using 3 vs 4 points for example). I have an [issue ](https://github.com/NOAA-EMC/RDASApp/issues/54)looking into this.
[...]

Issue #54, opened 2024-06-05
Understanding the differences between GSI and JEDI geovals


delippi:
In hofx testing, most testing has been done using GSI-IODA file derived from the GSI diagnostic files and forcing JEDI to use GSI derived geovals. When using the fv3jedi executable it is not [currently] possible to force JEDI to use the GSI geovals. The purpose of this issue is to document findings in understanding the differences in how GSI and JEDI compute geovals which lead to slight differences in hofx validation. Those slight differences in hofx validation should be considered "okay" so long as we can show where those differences come from.

delippi:
For `fv3_regional` in GSI:`guess_grids.F90`, the following method is used for computing the layer midpoint pressure (`ges_prsl`).:
```
       if (fv3_regional) then
          do jj=1,nfldsig
             do k=1,nsig
                 kp=k+1
                do j=1,lon2
                   do i=1,lat2
                      ges_prsl(i,j,k,jj)=(ges_prsi(i,j,k,jj)+ges_prsi(i,j,kp,jj))*half
                      ges_lnprsl(i,j,k,jj)=log(ges_prsl(i,j,k,jj))

                   end do
                end do
             end do
          end do
```
There only seems to be [currently] one recipe in VADER for computing air_pressure and that is `src/vader/recipes/AirPressure_A.cc`. In this recipe the method for computing the layer midpoint pressure (`airPressure`) is the Phillips method:
```
        airPressure(hh, vv) = pow(((pow(airPressureLevels(hh, vv+1), kap1) -
                                    pow(airPressureLevels(hh, vv), kap1)) /
                                    (kap1*(airPressureLevels(hh, vv+1) -
                                    airPressureLevels(hh, vv)))), kapr);

```
A new recipe for air_pressure could be added to match the GSI method for `fv3_regional`. For now, I modified the current recipe as follows to test my theory:
```
airPressure(hh, vv) = (airPressureLevels(hh, vv+1) + airPressureLevels(hh, vv)) / 2;
```

[...]

PR #90, opened 2024-06-25
Add validated mesonet yamls. And made minor changes to validation tools.

delippi:
This PR will merge: 
* mesonet (obtype 188) yamls that have been validated for hofx and oberror for airTemperature, specificHumidity, windEastward, windNortward, and stationPressure.
*  Some minor updates to the non-eva validation tools.

Issues:
This PR covers Issue #80
In addition there is a supplemental Issue #54 which adds justification for accepting some slight differences in final hofx and oberror.


CoryMartin-NOAA (commented):
Until we have proper YAML generation set up with JCB here, can we at least split apart the obs sections from the redundant background, background error, etc. sections. Like use `cat` + one YAML and then the `validated_yamls` to create the 'real YAML' that way we don't have the same few hundred lines duplicated in several places?

delippi:
> Until we have proper YAML generation set up with JCB here, can we at least split apart the obs sections from the redundant background, background error, etc. sections. Like use `cat` + one YAML and then the `validated_yamls` to create the 'real YAML' that way we don't have the same few hundred lines duplicated in several places?

Thanks @CoryMartin-NOAA. I had considered that, but I wasn't sure which would be best. I think it is a great suggestion. I will work on that.

ShunLiu-NOAA:
@CoryMartin-NOAA, Good suggestion. We are currently preparing a new MPAS test case. Once the new test case is ready, we will organize the related functionality testing, for example, reducing redundant session in YAML files and unifying YAML format to align with VAR and EnKF test case.

delippi:
@CoryMartin-NOAA, I was able to come up with a quick solution. One of my favorite tools to use is the stream editor (sed). I put a placeholder variable `@OBSERVATIONS@` in the yaml under basic_config and then that gets replaced with the contents of which ever obtype config yaml is chosen when you run gen_yaml.sh.

CoryMartin-NOAA:
>  One of my favorite tools to use is the stream editor (sed). 

@delippi you're an old soul :-)

SamuelDegelia-NOAA:
This is great! The only comment I have is in regard to the split yamls. If we cat these individual yamls together then the 

```
  observations:
     observers:
```

lines will repeat in the merged file. I'm not sure if that will throw an error or not. Should we remove the first two lines of the individual yamls to instead start with `- obs space:`?

delippi:
@SamuelDegelia-NOAA, good suggestion. I was only thinking of combining one of the obtype yamls with the basic configuration yaml at the moment. I think this is all a temporary storage solution anyway since we will move to JCB in a few months or so. I kind of don't have a preference on this. I will go ahead and make the change unless anyone objects.

SamuelDegelia-NOAA:
I just ask because it looks like GDASApp does something similar (e.g., https://github.com/NOAA-EMC/jcb-gdas/blob/develop/observations/atmosphere/sfc.yaml.j2). But I agree that this is just a temporary storage solution so either way works.

delippi:
@CoryMartin-NOAA do you guys have any tools for enforcing certain formatting standards for yamls? That would be great to start doing here.

CoryMartin-NOAA:
@delippi it's hard because we are using Jinja2 templates so they're non-standard YAML. We mainly just test to make sure they are valid YAML at the end

SamuelDegelia-NOAA (approved):

TingLei-NOAA (approved):

ShunLiu-NOAA (approved):

Issue #89, opened 2024-06-24
mpasjedi crashed for the new 2024052700 case


guoqing-noaa:
@Junjun-NOAA and I have been creating a new `mpasjedi_2024052700 `case following @chunhuazhou setting but using the community physics.

We have successfully created the bump localization files.

But mpasjedi crashed when running a pure 3DEnVar single sonde observation case (similar as in issue #83 )

The run directory is at:
`/scratch1/BMC/wrfruc/gge/RDASApp/expr/jedi-assim_test/rrfs-data_mpasjedi_2024052700`

We found that `mpasjedi_2022052619 `case modified all ensemble restart files with a `dchg.py` program. Is this key to successfully running a mpasjedi case? If yes, why would we need to make this change?

Any suggestions? @TingLei-NOAA @xyzemc @SamuelDegelia-NOAA 

Thanks!

guoqing-noaa:
I ran dchg.py on each restart file but mpasjedi still crashed.

```
  0: ==== backtrace (tid: 799561) ====
  0:  0 0x00000000000534e9 ucs_debug_print_backtrace()  ???:0
  0:  1 0x0000000000012cf0 __funlockfile()  :0
  0:  2 0x0000000000a473b0 vert_interp_mod_mp_vert_interp_weights_()  /scratch1/BMC/wrfruc/gge/myRDASApp/bundle/ufo/src/ufo/utils/vert_interp.F90:73
  0:  3 0x0000000000a9a828 ufo_vertinterp_mod_mp_vertinterp_simobs__()  /scratch1/BMC/wrfruc/gge/myRDASApp/bundle/ufo/src/ufo/operators/vertinterp/ufo_vertinterp_mod.F90:341
  0:  4 0x0000000000a96567 ufo_vertinterp_simobs_f90()  /scratch1/BMC/wrfruc/gge/myRDASApp/bundle/ufo/src/ufo/operators/vertinterp/ObsVertInterp.interface.F90:98
  0:  5 0x0000000000a8cc3b ufo::ObsVertInterp::simulateObs()  /scratch1/BMC/wrfruc/gge/myRDASApp/bundle/ufo/src/ufo/operators/vertinterp/ObsVertInterp.cc:60
[...]

Issue #87, opened 2024-06-21
Need to update mpas and mpasjedi in RDASApp


guoqing-noaa:
We tried using [Chunhua's MPAS case](https://github.com/NOAA-EMC/RDASApp/issues/41) to create a CONUS_12km MPASJEDI case. 
The test directory can be found at:
`/scratch1/BMC/wrfruc/gge/RDASApp/expr/jedi-assim_test/rrfs-data_mpasjedi_2024052621`

We found that Chunhua used the GSL MPAS versions with GSL new physics and this cannot be handled by the current MPASJEDI and MPAS submodules in RDASApp. 

We discussed this with Tao Sun from the NCAR MPASJEDI team.  Per Tao's information, it is best to use the latest commit in the [`feature/MPAS-Model_v8.1`](https://github.com/JCSDA-internal/mpas-jedi/commits/feature/MPAS-Model_v8.1/) branch of mpas-jedi and commits after May 21, 2024 of [the develop branch of MPAS-Model](https://github.com/MPAS-Dev/MPAS-Model/commits/develop/?before=7070294994eb6e53e9133f15711cf0285fc70437+105) which contain necessary codes for using new physics.

So, if RDASAPP wants to use GSL's new physics, [GSL's MPAS-Model](https://github.com/NOAA-GSL/MPAS-Model/commits/gsl/develop/) needs to include the above-mentioned commits from the community MPAS-Model. This will happen but not in the near future

Thank @Junjun-NOAA, @spanNOAA, @xyzemc and @TingLei-NOAA for help and discussion in this process.
Tag @ShunLiu-NOAA @hu5970

guoqing-noaa:
For reference, here is Xiaoyan's MPAS-model used to generate the 2022052619 mpasjedi case in current RDASApp:

```
/scratch2/NCEPDEV/fv3-cam/Xiaoyan.Zhang/noscrub/MPAS/MPAS-Model.befRocky8

Merge: 1d6ec7c21 2bfdc6ba3
Author: Michael Duda <duda@ucar.edu>
Date:   Thu Jul 6 23:46:22 2023 +0000

    Merge branch 'hotfix-v8.0.1'
```

And Ting's script to make it a mpasjedi case:
`/scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-mpas-bundle/rdas-mpas-tests/CONUS_15km_da/prep_denvar_reg_xzhang.csh`

[...]

PR #91, opened 2024-06-26
Add the MPASJEDI 2024052700 case - CONUS12km with 30 ensembles

guoqing-noaa:
See issue #92 for more details


SamuelDegelia-NOAA:
@guoqing-noaa Thanks for putting this case data together! It will be nice to have a much more realistic case to run MPAS experiments with. I have a couple of early questions here to open some discussion: 

1) Do we plan to keep the original `2022052619` case for running a CTest? Or should we maybe just take a small subset of the data added in this PR and rebuild the CTest? That way we aren't keeping multiple sets of test data around. 
2) What about making use of the existing `rrfs-test/` directory instead of `expr/` to include some of these scripts for setting up the experiment? 
3) Do we need to include these fix files (e.g., `geovars.yaml`, `namelist.atmosphere`, and `streams.atmosphere`) directly in the repo, or should we keep them in staged data that is copied in via `ush/init.sh`? For the existing CTest data we have been keeping those in the staged data. They are very small files but I think keeping them with the rest of the staged data makes sense because a user likely will not need to edit or look at them.

ShunLiu-NOAA:
> @guoqing-noaa Thanks for putting this case data together! It will be nice to have a much more realistic case to run MPAS experiments with. I have a couple of early questions here to open some discussion:
> 
> 1. Do we plan to keep the original `2022052619` case for running a CTest? Or should we maybe just take a small subset of the data added in this PR and rebuild the CTest? That way we aren't keeping multiple sets of test data around.
Let's keep  `2022052619` case for a while until we fully practice the new test case. 
> 2. What about making use of the existing `rrfs-test/` directory instead of `expr/` to include some of these scripts for setting up the experiment?
We should keep the current infrastructure and use "rrfs-test/'.
> 3. Do we need to include these fix files (e.g., `geovars.yaml`, `namelist.atmosphere`, and `streams.atmosphere`) directly in the repo, or should we keep them in staged data that is copied in via `ush/init.sh`? For the existing CTest data we have been keeping those in the staged data. They are very small files but I think keeping them with the rest of the staged data makes sense because a user likely will not need to edit or look at them.
We should discuss this.

guoqing-noaa:
Thanks @SamuelDegelia-NOAA  and @ShunLiu-NOAA for inputs.

> 1. Do we plan to keep the original `2022052619` case for running a CTest? Or should we maybe just take a small subset of the data added in this PR and rebuild the CTest? That way we aren't keeping multiple sets of test data around.

I agree with Shun that we can keep the 2022052619 case for a while. 
It will take a while to get the "2024052700" case more complete.

> 2. What about making use of the existing `rrfs-test/` directory instead of `expr/` to include some of these scripts for setting up the experiment?

My understanding of `rrfs_test/` is that it is used mostly for RRFS CTest. I might be wrong.
I can move expr to rrfs_test if no objections. 
However, I would like to skip the `setup_experiment.sh` step for the 2024052700 case. 
I am open to changes if anyone would like to make `setup_experiment.sh` work for this case. The bottom line is we would like to link data instead of copying data.

> 3. Do we need to include these fix files (e.g., `geovars.yaml`, `namelist.atmosphere`, and `streams.atmosphere`) directly in the repo, or should we keep them in staged data that is copied in via `ush/init.sh`? For the existing CTest data we have been keeping those in the staged data. They are very small files but I think keeping them with the rest of the staged data makes sense because a user likely will not need to edit or look at them.

Users will modify those files (I did) for different needs. It is good to include them in the repo directory. They usually will be included under parm/.

expr/mpasjedi_2024052700/colormap.py
delippi, line 1: This is a duplicate of `rrfs-test/ush/colormap.py`. Maybe we could find a better place for it and then we can can just put it in PYTHONPATH.

expr/mpasjedi_2024052700/mpasjedi_increment_singleob.py
delippi, line 1: This is also a duplicate script to `rrfs-test/ush/mpasjedi_increment_singleob.py`. If there were changes it might be best to just update the one in rrfs-test?

expr/mpasjedi_2024052700/testinput/msonet_airTemperature.yaml
delippi, line 1: This is also a duplicate and probably not needed for the mpas case? This looks like an fv3jedi yaml.

expr/mpasjedi_2024052700/testinput/msonet_singleob_airTemperature_mpasjedi.yaml
delippi, line 1: This and the other yaml probably do not need the "singleob" tag in the filename. I know I did that a while ago but realized it wasn't needed because this singleob yaml should also work just fine for full ob test given your obs file has the correct name.

delippi (commented):
Hi @guoqing-noaa, I think @SamuelDegelia-NOAA's suggestion of using rrfs-test is fine. While it is used for ctests, it is also used for setting up "rrfs tests" in general... At least that is my understanding. So that is why I had the setup_experiment script located there and it pulls the staged data, plotting scripts, etc. It would be nice to have this case ready to use for setup_experiment.sh. That setup script doesn't necessarily HAVE to copy data either, although that is/was just my preference since you mostly just copy data once and I like having hard copies incase anything went wrong. But it is understandable when datasets get too large.

Overall, it would be great to incorporate your init.sh into setup_experiment.sh and put things in rrfs-test. Idk if you've used setup_experiments.sh yet, but my intention was to make this as user friendly as possible when getting started. And it would be great to have one singular place where a new user can go and configure their experiments independent on which case or dycore they are working with.

expr/mpasjedi_2024052700/colormap.py
guoqing-noaa: Yes, it is a duplicate. I can copy it from its original location.

expr/mpasjedi_2024052700/mpasjedi_increment_singleob.py
guoqing-noaa: It is changed as the new mpasjedi case has different directory/name setting, such as different bkg name etc.
I think it is fine to have a variant copy here.

expr/mpasjedi_2024052700/testinput/msonet_airTemperature.yaml
guoqing-noaa: Thanks, I can remove this one.

expr/mpasjedi_2024052700/testinput/msonet_singleob_airTemperature_mpasjedi.yaml
guoqing-noaa: I did not test this. I can change the name or remove it for now. We can add it back when we complete a test for the new case.

guoqing-noaa:
@delippi Thanks for the review! 

I had an alternate view for `setup_experiment.sh` at the current stage. This script is very helpful for set up different experiments, FV3 or MPAS, hofx validation, etc. 

However, going forward, we will focus the MPAS JEDI case. There is no direct way to do hofx validation for MPASJEDI.
My expr setting enables a quick plug-in for anyone to try the MPASJEDI case (only need to specify the slurm account/queue/partition and it is done). Most developers don't need to deal with the FV3 core and the hofx validation any more.

On the other hand, `setup_experiment.sh` or a similar concept may be helpful when we work on the whole RRFSv2 workflow.

expr/mpasjedi_2024052700/mpasjedi_increment_singleob.py
delippi: I think we should try to keep duplicate plot scripts to a minimum. We should be able to configure them to be used generically. You could also configure them to take arguments. I have examples. I'm not sure what the best solution is. At the same time, this case will replace the current MPAS case so we could just edit the script that already exists to be used for this case. It might be worth reconfiguring the MPAS ctests and everything based on this case and removing the current MPAS case completely since it is only good to test functionality.

expr/mpasjedi_2024052700/testinput/msonet_singleob_airTemperature_mpasjedi.yaml
delippi: Does this yaml use the built in single ob tool or does it just take an IODA file containing a single ob? If the latter then there is no need for the "singleob" tag in the filename. Let's remove this one for now and then consider renaming the other yaml with "singleob" in its name.

delippi:
@guoqing-noaa, will we have a corresponding FV3JEDI and/or GSI case to go along with this new MPAS case? I'm under the impression or was at least hoping that we would. MPAS still needs validated somehow. While there is no way to do a direct comparison that we can make I think it would be useful to compare GSI with MPASJEDI to show that they agree to some degree expecting some differences. It seems dangerous to not have such validation.

delippi:
Also, setup_experiment can be configured to turn off GSI and EVA. So it can just copy and setup your MPAS directory only.

expr/mpasjedi_2024052700/mpasjedi_increment_singleob.py
guoqing-noaa: The current ctest mpas case is till valuable for providing a baseline test.
I think it is better to make a big update to current ctests after the summer when we will update to MPASv8.2 and MPASJEDIv3.0 (some more information in this issue #87 )

If you agree, we can make this version of py file the default one and users will need to modify settings if using it on the ctest mpas case.

expr/mpasjedi_2024052700/testinput/msonet_singleob_airTemperature_mpasjedi.yaml
guoqing-noaa: Thanks!

expr/mpasjedi_2024052700/mpasjedi_increment_singleob.py
delippi: I agree that the current MPAS case can provide some good baseline tests, but at the same time it might be good to do some house keeping. It might take a little bit of effort to reconfigure all the ctests with the new case, but it might be better long term to only have the one case. I'm also okay with doing this after the summer with those updates though.

I'm not currently using the mpas version of the python script. I'm not sure if anyone else is either since that case really isn't too helpful other than for ctests. I'm okay with you updating the one that we currently have.

guoqing-noaa:
> @guoqing-noaa, will we have a corresponding FV3JEDI and/or GSI case to go along with this new MPAS case? I'm under the impression or was at least hoping that we would. MPAS still needs validated somehow. While there is no way to do a direct comparison that we can make I think it would be useful to compare GSI with MPASJEDI to show that they agree to some degree expecting some differences. It seems dangerous to not have such validation.

Unfortunately, we cannot do direct comparisons between MPASJEDI and GSI like we can do between FV3JEDI and GSI. We have to live with this reality.

But we still manage to have some baseline GSI tests to compare with MPASJEDI.
@Junjun-NOAA and @spanNOAA has been exploring to establish some comparison capability. There are two routes to go. We can discuss this further in our joint meeting. Or we can create an issue now to update the progress. Thanks for the suggestion!

guoqing-noaa:
@SamuelDegelia-NOAA @delippi @ShunLiu-NOAA 

Thanks all for the review!
I've made updates following suggestions. Welcome your second round of review. Thanks!

guoqing-noaa:
I should note that `rrfs-test/ush/mpasjedi_spread.py` comes from @SamuelDegelia-NOAA 
I just added it to the repo.

delippi:
Hi @guoqing-noaa, thanks for making some changes. I would still say that your link_expr script could be incorporated into the setup_experiment script. Then the run_jedi script should go rrfs-test/scripts/templates--well, just update the one that is already there. You could do the same thing with run_bump script. The namelist and streams files could either go in the staged location or not sure where. Then we wouldn't need this additional mpasjedi_2024052700 directory (that directory would be made as a work space after running the setup_experiment script). Basically I'm trying to make sure we don't have too many scripts that do similar things--if they can be combined I would like to see them combined.

The setup_experiment script doesn't necessarily only have to be for hofx validation. I have options to configure so that you don't have to copy GSI or EVA. So this script could be used to setup your MPAS case. Also, you could add some logic that if it is the MPAS case, then just link instead of copy/rsync. The old MPAS case could be commented out and replaced with this case in that script.

I'm happy to discuss further.

guoqing-noaa:
@delippi Thanks for the reivew. It will be good if you can help with adopting `setup_experiment `. I am open to any changes.

delippi:
@guoqing-noaa, sure. I will help.

guoqing-noaa:
Thanks, @delippi!

I would provide a second opinion. Most people will use this MPASJEDI case to do JEDI exploration and only a very few developers may use the FV3JEDI case or the hofx validation to examine some mysteries if we are not able to get a clear answer when working on MPASJEDI. So my current setting is much easier and simpler for them.

guoqing-noaa:
A few suggestions on the `setup_experiment ` while you are working on it:

1. A script can detect the RDASApp directory automatically, user don't need to manually set it. You may refer the run_mpasjedi.sh in this PR.

2. I suggest we use the following default setting:
```
DYCORE="MPAS" #FV3 | MPAS
GSI_TEST_DATA="NO"
YOUR_PATH_TO_GSI="/path/to/your/installation/of/GSI" # Not needed for MPASJEDI
EVA="NO"
```

3. Users will still need to modify the final job card to change partition, qos, or reservation in current `setup_experiment ` configuration

guoqing-noaa:
@delippi @SamuelDegelia-NOAA @TingLei-NOAA @ShunLiu-NOAA 
I think this PR is good to go. Thank you for the help, @delippi!

TingLei-NOAA:
@guoqing-noaa  and/or others, A general quesiton. When I check out rdasapp with your PR,  what scripts/commands shoud I use to set up and run the mpasjedi case added in this PR?

guoqing-noaa:
@delippi Are you able to help @TingLei-NOAA to run the MPAS2024052700 case? My internet connection is very slow. It will be great if you can help. Thanks a lot!

delippi:
> @guoqing-noaa and/or others, A general quesiton. When I check out rdasapp with your PR, what scripts/commands shoud I use to set up and run the mpasjedi case added in this PR?

@TingLei-NOAA, please use `rrfs-test/scripts/setup_experiment.sh`. This script is not only for the hofx validation but also for setting up experiments in general. Just make sure to have dycore="MPAS". The GSI and EVA settings can be turned off. These should all be the default settings now.

TingLei-NOAA:
@delippi  Thanks.  Using the setup_experiment.sh, it successfully created a exp dir.  The changed script is /scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-guoqing/RDASApp//rrfs-test/scripts/setup_experiment.sh. 
The expdir is dr-test . But running "sbatch run_mpasjedi.sh sonde_singeob_airTemperature_mpasjedi.yaml"  got a failed run with error messages like 
```
ERROR: In file ./streams.atmosphere, definition of stream "background" references file stream_list/stream_list.atmosphere.background that cannot be opened for reading.
```
I guess some files are not copied to this expdir. 
I will leave the fix to @guoqing-noaa  when he is available or other rdasapp developers.

delippi:
Hey @TingLei-NOAA, I think I know what the issue is. I made a patch to @guoqing-noaa's PR but was unsure about one part so this part was left out. I think I understand how that was supposed to work now. I will try to make another patch to the PR.

guoqing-noaa:
@ShunLiu-NOAA @delippi Do we have any remaining questions on this PR? It will be best to merge this PR as earlier as possible. A few GSL developers would like to use this case to test/do some developments. Thanks!

SamuelDegelia-NOAA:
@guoqing-noaa It sounds like we were having some issues running the experiments per @TingLei-NOAA's recent comment. Are you able to merge @delippi's PR in your fork so that we could test it again?

guoqing-noaa:
Thanks, @SamuelDegelia-NOAA!
Let me work on this.

guoqing-noaa:
@delippi Thanks for your new PR to my branch! We will need to have a tag up to resolve differences.
Generally, `init.sh `is not only used by the MPAS JEDI case. It has more general use and will have more functionalities. We would like to keep it. I can discuss this more with you next week.

At the same time, I will share my personal branch to those who would like to try the new MPAS case as soon as possible.
Once this PR is resolved sometime next week, we can transition to your setup_expr method. Thanks!

TingLei-NOAA:
@guoqing-noaa ,so for the current PR, do we have a way to set up you case or shall we wait for a new update?

ShunLiu-NOAA:
@guoqing-noaa Do you have a chance to look at Ting's question? Ting will test the case for reviewing this PR.

guoqing-noaa:
> @guoqing-noaa ,so for the current PR, do we have a way to set up you case or shall we wait for a new update?

Hi @TingLei-NOAA , I sent you an email last Friday on how to test this MPASJEDI case. I will reply to that email to remind you.
For this PR, I will think more about how to resolve the differences with @delippi

TingLei-NOAA:
@guoqing-noaa   Thanks for this clarification!

guoqing-noaa:
Per Donnie's test, the PR now works correctly.  
@ShunLiu-NOAA @delippi @SamuelDegelia-NOAA @TingLei-NOAA Please make a new review. Thanks!

rrfs-test/scripts/setup_experiment.sh
```diff
@@ -89,9 +133,9 @@ if [[ $GSI_TEST_DATA == "YES" && $DYCORE == "FV3" ]]; then
 fi
 
 # Copy rrts-test yamls and obs files.
-mkdir -p testinput
-cp -p $YOUR_PATH_TO_RDASAPP/rrfs-test/testinput/* testinput/.
-cp -p $YOUR_PATH_TO_RDASAPP/rrfs-test/obs/* Data/obs/.
+#mkdir -p testinput
+#cp -p $YOUR_PATH_TO_RDASAPP/rrfs-test/testinput/* testinput/.
```
SamuelDegelia-NOAA, line 137: Could we add this back in? The job script `run_mpasjedi.sh` checks for the `testinput` directory if a yaml file is not passed into the job script. Plus it helps to have all of those yamls in the run directory if we want to try different configurations.

TingLei-NOAA (dismissed):
The changes look good to me when it had been tested by @delippi as @guoqing-noaa mentioned. 
Thanks!

rrfs-test/ush/mpasjedi_increment_singleob.py
```diff
@@ -39,11 +39,12 @@
 
 # JEDI data
 datapath = "./"
-janalysis   = "./an.2022-05-26_19.00.00.nc"            # analysis file
-jbackgrnd   = "./Data/bkg/bg.2022-05-26_19.00.00.nc"            # background file
-jstatic      = "./Data_static/static.93175.nc"                         # to load the MPAS lat/lon
```
SamuelDegelia-NOAA, line 44: Could we make the same edits in `mpasjedi_increment_fulldom.py`? I am fine with renaming the output file from `an.` to `anl.` but now the `mpasjedi_increment_fulldom.py` script will not work without changing these corresponding lines in it.

SamuelDegelia-NOAA:
Overall things are working for me and look good after removing the Jet-specific job submission lines above. I ran the EnVar single observation test and the results look more realistic than our previous case. Looking forward to working on this case once the PR is merged! 

Apart from my other line-specific comments, I have a couple other suggestions. Apologies if these have already been discussed offline with @delippi. 

1. Could we add something to the wiki or build instructions about needing to run `ush/init.sh` to set up this case? Just running `setup_experiment.sh` alone will not work unless the former is run first. An alternative is to integrate `ush/init.sh` into `setup_experiment.sh`. I know you mentioned earlier about not wanting to do this, but with how fast it runs (nearly instant) I still think it could be a good idea to save a step. 

1. Do the `namelist.atmosphere` and `streams.atmosphere` files need to go in `rrfs-test/testinput`? We have mostly been housing JEDI yamls in there. Could we instead put those in the new `fix/` dir created by `ush/init.sh`?

guoqing-noaa:
> Overall things are working for me and look good after removing the Jet-specific job submission lines above. I ran the EnVar single observation test and the results look more realistic than our previous case. Looking forward to working on this case once the PR is merged!
> 
> Apart from my other line-specific comments, I have a couple other suggestions. Apologies if these have already been discussed offline with @delippi.
> 
> 1. Could we add something to the wiki or build instructions about needing to run `ush/init.sh` to set up this case? Just running `setup_experiment.sh` alone will not work unless the former is run first. An alternative is to integrate `ush/init.sh` into `setup_experiment.sh`. I know you mentioned earlier about not wanting to do this, but with how fast it runs (nearly instant) I still think it could be a good idea to save a step.
> 2. Do the `namelist.atmosphere` and `streams.atmosphere` files need to go in `rrfs-test/testinput`? We have mostly been housing JEDI yamls in there. Could we instead put those in the new `fix/` dir created by `ush/init.sh`?

Good comments! Thanks @SamuelDegelia-NOAA 

1. Sorry that I actually should commit init.sh first and not mix it with this PR. 
init.sh is to initialize fix/ directory for a centralized data location and create a pre-commit hook to prevent accidental commit of binary data into the repo. Its final form will be more similar to this:
https://github.com/rrfs2/rrfs-workflow/blob/main/ush/init.sh
init.sh usually runs at the building time (this is what rrfsv1 did). But people may skip the building process. To prevent this, I would suggest adding `init.sh` into `setup_experiment.sh` as a solution as well. @delippi are you okay with that?
I understand you mentioned corner cases to me where one may not be able to run `init.sh`. Let's ignore corner cases for now?

2. I am fine to put them under parm/ (namelists usually go into parm/).

Thanks!

delippi:
@guoqing-noaa & @SamuelDegelia-NOAA, I think `init.sh` should just be part of the build process. The path `YOUR_PATH_TO_RDASAPP` in `setup_experiment.sh` should only point to an RDASApp that has already been built. I hope that answers the question? I didn't have any such problem because I pointed to Guoqing's RDASApp build and it worked fine for me.

SamuelDegelia-NOAA:
I think integrating `init.sh` into the build process is also a good solution.

SamuelDegelia-NOAA:
Hi @guoqing-noaa, I am trying to run some LETKF tests with the new 2024052700 case here. But I am getting an error with missing pressure data in the ensemble files. It looks like the only pressure variable in the ensemble files is `surface_pressure`. The `pressure_p` and `pressure_base` variables are present the the deterministic background file for your new case, but the LETKF also needs these variables in the ensemble members. Is it possible to add this?

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for testing this.
Yes, this is an issue. We created a ticket here:
https://github.com/rrfs2/rrfs-workflow/issues/22

I am not sure who would like and will be able to work on this.

For RDASApp,  we can update the ensemble BEC by using 6h MPAS forecasts. But that will also have a problem that we don't add any perturbations to the 30 members yet. @chunhuazhou is working on applying realistic ensemble initialization. Her work may provide some realistic ensemble BEC.

SamuelDegelia-NOAA:
@guoqing-noaa I see, so the deterministic background is a 6h MPAS forecast which adds the pressure variables? But the ensemble members are cold start files that do not have it yet? If I am understanding that correctly, could we just run forecasts with the ensemble to generate the pressure fields?

guoqing-noaa:
@SamuelDegelia-NOAA 
Yes, if you run ensemble forecasts, we will get pressure fields. We have had that before. But it gives unexpected small single ob increments so we changed them.

SamuelDegelia-NOAA:
@guoqing-noaa Ahh okay, so that is the issue we found in #89. Do you happen to still have those ensemble forecasts saved? I would be happy to work on adding random perturbations to them to increase spread, just so we could get the LETKF working for this new test case.

guoqing-noaa:
@SamuelDegelia-NOAA You are right. It is the issue in #89.
If you can add random perturbation, that would be great! Thanks!

The 30 member 6h forecasts are here
`/scratch1/BMC/wrfruc/gge/RDASApp/expr/jedi-assim_test/rrfs-data_mpasjedi_2024052700/Data/ens.czhou.00z`

guoqing-noaa:
@SamuelDegelia-NOAA FYI, @chunhuazhou is working on updating the ensembles in this case to be 1h MPAS forecasts initialized from GEFS 29h fcst (Per David Dowell and Chunhua's previous examination, GEFS 30 h fcsts have better ensemble spread).

SamuelDegelia-NOAA:
@guoqing-noaa thank you for the update. I did add perturbations to the forecast ensemble to be usable with LETKF, but I've been running into this new issue with the hofx for each member being the exact same (https://github.com/NOAA-EMC/RDASApp/issues/111). I'm working on debugging that now. I'd assume the new forecasts @chunhuazhou is working on will have the same problem, but we'll see.

guoqing-noaa:
@ShunLiu-NOAA @TingLei-NOAA @delippi @SamuelDegelia-NOAA @hu5970  Please review the updates.

The ensembles have been updated to use the fhr35 GEFS initialized 1-h MPAS forecasts prepared by @chunhuazhou .  
The fix/ directory has been added (this will facilitate a centralized location for RDAS_DATA, I will create a new PR for this).   
A quick start option has been added. Donnie and I previously agreed that we keep both the powerful setup_experiment.sh and this quick start option. But I did not get time to work on this until now.

Thanks and let me know if you have any concerns.

SamuelDegelia-NOAA:
I tested LETKF using the new ensembles and things are looking good. Thanks @chunhuazhou for adding these. Increments for a single observation test are about double the magnitude compared to the old ensemble using the same obs error and innovation.

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for testing!
Yes, the ensemble spread looks better using the 35h GEFS forecasts. This is copied from issue #89 

level=23, 20240527_00z ensemble spread from 1h MPAS fcst initialized from 35h GEFS forecast

image
    
    
The associate temperature increment at level 23:
    
image

guoqing-noaa:
@ShunLiu-NOAA @TingLei-NOAA @delippi @SamuelDegelia-NOAA @hu5970
Let me know if you have any concerns.

.gitignore
```diff
@@ -148,3 +148,8 @@ install*/
 # Ignore the following bundle repositories
 !/bundle/CMakeLists.txt
 /bundle/*
+
+# 
+expr
+fix/INIT_DONE
+fix/.agent
```
TingLei-NOAA, line 155: @guoqing-noaa  Would you explain what the purposes are for use of .agent dir ?

SamuelDegelia-NOAA (dismissed):

rrfs-test/scripts/setup_experiment.sh
guoqing-noaa: @SamuelDegelia-NOAA I think I did not change this file. @delippi Could you help address Sam's question here?

rrfs-test/ush/mpasjedi_increment_singleob.py
guoqing-noaa: Thanks, @SamuelDegelia-NOAA What change we should make? Do you mean we need to make this file also work for the previous MPASJEDI case? 

I have suggested I keep a standalone copy of this file for the new MPASJEDI case. But after discussion with Donnie, we went another route.  I think it is still a good idea  that we don't change the behavior of the previous MPASJEDI case. I will make a change and keep two copies of this file. Does this work for you?

SamuelDegelia-NOAA: Sorry @guoqing-noaa, I think my review comments just showed up despite typing them a couple of weeks ago because I still had my review as "pending". I am still learning how to use Github, hah. 

But anyways I just meant to also edit `mpasjedi_increment_fulldom.py` to be consistent with these changes to  `mpasjedi_increment_singleob.py` for the new MPAS-JEDI case. This new case changes the analysis filename prefix from `an.${DATE}` to `ana.${DATE}` and changes the path for these static files. This change to the singleob plotting script highlighted here represents those new file names, but `mpasjedi_increment_fulldom.py` still uses the original names. So basically I am just asking to make these same three line changes to `mpasjedi_increment_fulldom.py`.

guoqing-noaa: Ah, I got it. Thanks for the clarification and mentioning this. I forgot that part.

.gitignore
guoqing-noaa: > @guoqing-noaa Would you explain what the purposes are for use of .agent dir ?

@TingLei-NOAA Good question!  The .agent link is a byproduct of the fix file management which was used in the current rrfsv1 workflow.  

The general idea is that binary fix files and text fix files coexist under the fix/ directory. Text files are tracked by the repo similar as other script/config/etc files while binary files are tracked differently using links which point to the real data locations which is given by the RDAS_DATA variable. In different machines, RDAS_DATA takes different values. It will be tedious if we change every links when transitioning from one machine to another machine. The .agent link is used to simplify this task and we only need to update the .agent link on different machines. Here is a diagram to show this. [This link](https://github.com/rrfs2/rrfs-workflow/wiki/fix-file-management-policy) provides more details on the fix file management. Let me know if you want to have a short tagup to have a further discussion on this.

image

guoqing-noaa:
@SamuelDegelia-NOAA I have updated the mpasjedi_increment_fulldom.py file as suggested and I can generated a figure as follows.  
Sorry that this commit automatically dismissed your approval. Could you review again? Thanks!
    
image

SamuelDegelia-NOAA (approved):
Thanks for the quick update @guoqing-noaa!

.gitignore
guoqing-noaa: @TingLei-NOAA we can talk more on this but I think it is not a roadblock to merge this PR. I will resolve this conversation to get ready for a merge. Thanks!

rrfs-test/scripts/setup_experiment.sh
guoqing-noaa: I will leave this to @delippi

.gitignore
TingLei-NOAA: @guoqing-noaa  I think this PR don't have to have my approval to get merged . But from my point of view, it is a question to be clarified before merging. So, I suggest  to leave it unresolved as it is ( and, I think it is not required that every conversion should be resolved before a PR's merging).  Thanks.

guoqing-noaa:
I plan to merge this PR early tomorrow if no objections. There may be a few minor things but we can always update it and make it better later. A few is awaiting to use changes in this PR (they have been using my forked branch) and it is better to use the official RDASApp.

.gitignore
guoqing-noaa: @TingLei-NOAA  Sorry that we have to resolve all conversations otherwise a merge is NOT allowed. 
But even the conversation is resolved at the PR level, we can continue to make comments, raise concerns and even we can create a new issue for this.

guoqing-noaa:
I had a tag-up meeting with @TingLei-NOAA this morning. We discussed and addressed the remaining concerns. I think it is good to merge this PR now.

TingLei-NOAA (approved):

ShunLiu-NOAA (approved):

ShunLiu-NOAA:
@guoqing-noaa is it possible that you stage the data and test case to role account on Hera and Jet before we merge this PR? We can also have another PR to move test cases and data to role account. Which one is better for you now?

guoqing-noaa:
@ShunLiu-NOAA all data has been staged under the role account role.rrfs-fv3-cam on Hera.

PR #95, opened 2024-06-28
Creating mega rdas.x to replace individual executables

### LINKED ISSUE #96: Create mega rdas.x executable
To be more NCO-compliant, it is desired to create a single `rdas.x` executable instead of using individual executables for each application (i.e., `fv3jedi_letkf.x`, `fv3jedi_var.x`, `fv3jedi_hofx.x`, `fv3jedi_error_covariance_toolbox.x`). I have begun work on this process in #95.

SamuelDegelia-NOAA:
This PR creates two "mega" executables called `rdas_fv3jedi.x` and `rdas_mpasjedi.x`. These larger executables combine many of the individual executables that previously only worked for a single application (i.e., variational, letkf, bump). This is needed for NCO compliance and generally follows [GDASApp PR #1075](https://github.com/NOAA-EMC/GDASApp/pull/1075). 

The choice to create separate executables for fv3jedi and mpasjedi will make it easier to simply remove one whenever we start real-time runs. 

There are some differences for `rdas_${dycore}.x` from `gdas.x` including the new mpas-jedi version, removal of soca options, addition of bump option, and a slightly different build route via `RDASApp/bundle/CMakeLists.txt`. 

The mega executables have been tested on Hera for both fv3-jedi and mpas-jedi including the newly added bump option. I also updated the ctests and the template scripts `run_fv3jedi_template.sh` and `run_mpasjedi_template.sh` for this new executable structure. One thing I noticed is that some of the mpasjedi ctests initially failed due to a memory error. I am not sure what change could have caused this, but I added a `--ntasks-per-node=18` option to the ctests to allow them to use 2 nodes instead of 1. After that, everything seems to be working as expected. 

Usage examples: 

```
./rdas_fv3jedi.x variational ${yaml}
./rdas_fv3jedi.x localensembleda ${yaml}
./rdas_fv3jedi.x bump ${yaml}
./rdas_mpasjedi.x variational ${yaml}
./rdas_mpasjedi.x localensembleda ${yaml}
./rdas_mpasjedi.x bump ${yaml}
```

CTest results: 

```
[Samuel.Degelia@hfe08 rrfs-test]$ cd RDASApp/build/rrfs-test
[Samuel.Degelia@hfe08 rrfs-test]$ ctest
Test project /scratch1/NCEPDEV/da/Samuel.Degelia/RDASApp_dev/RDASApp/build/rrfs-test
    Start 1: rrfs_fv3jedi_hyb_2022052619
1/5 Test #1: rrfs_fv3jedi_hyb_2022052619 ............   Passed  126.21 sec
    Start 2: rrfs_fv3jedi_letkf_2022052619
2/5 Test #2: rrfs_fv3jedi_letkf_2022052619 ..........   Passed   21.84 sec
    Start 3: rrfs_mpasjedi_2022052619_Ens3Dvar
3/5 Test #3: rrfs_mpasjedi_2022052619_Ens3Dvar ......   Passed   53.59 sec
    Start 4: rrfs_mpasjedi_2022052619_atms_npp_qc
4/5 Test #4: rrfs_mpasjedi_2022052619_atms_npp_qc ...   Passed   78.00 sec
    Start 5: rrfs_mpasjedi_2022052619_letkf
5/5 Test #5: rrfs_mpasjedi_2022052619_letkf .........   Passed   42.73 sec

100% tests passed, 0 tests failed out of 5

Label Time Summary:
mpi            = 322.37 sec*proc (5 tests)
rdas-bundle    = 322.37 sec*proc (5 tests)
script         = 322.37 sec*proc (5 tests)

Total Test time (real) = 323.51 sec
```


guoqing-noaa:
@SamuelDegelia-NOAA Thanks for working on this. Look forward to it.

It is a good idea for everyone to create an issue for what he is working on or planning to work on to avoid duplicate efforts and facilitate discussions.

SamuelDegelia-NOAA:
Thank you for the suggestion, @guoqing-noaa. I mentioned that I would be working on this during some internal EMC meetings but it would be better to open an issue here on RDASApp so that all collaborators are aware. I will do that now.

SamuelDegelia-NOAA:
I think this PR is now good to go for review. I updated the main text with some new information.

mains/rdas_fv3jedi.cc
```diff
@@ -0,0 +1,108 @@
...
+    fv3jedi::instantiateObsLocFactory();
+  }
+
+  // Application pointer
+  std::unique_ptr<oops::Application> app;
+
+  // Define a map from app names to lambda functions that create unique_ptr to Applications
+  std::map<std::string, std::function<std::unique_ptr<oops::Application>()>> apps;
+
+  apps["convertstate"] = []() {
```
CoryMartin-NOAA, line 51: you may want the new "converttostructuredgrid" to make lat/lon plotting easier

CoryMartin-NOAA (dismissed):
One suggestion but otherwise this looks great!

mains/rdas_fv3jedi.cc
SamuelDegelia-NOAA: Thank you Cory for the quick review! Unfortunately the oops hash linked in RDASApp is from [four months ago](https://github.com/jcsda/oops/tree/f02f283b7c4ea1edae349413964f3edf0f7fab33), before the ConvertToStructuredGrid app was added. I don't think we have a made a decisions about when or how often to update these. Is it an easy change to update?

CoryMartin-NOAA: ah good point, it may be non-trivial to update everything. I would suggest doing that in a future PR.

TingLei-NOAA:
@SamuelDegelia-NOAA  I have two general comments.  First, It is good to follow gdas to have a super executable to be ready when it is the solution to meet the nco requirement.  But I don't think it is good to use now in rrfs scripts/ctests ( because, use of such a super executable only makes the readability/organization less clear).  I would suggest this PR only include the changes for codes.     Second,  I believe, there would be similar situations in the future when rdas could borrow stuff from gdas and do rdas's own changes/modifications to them.  If possible, I would suggest to explore "git subtree" to transfer those components from gdas to rdas.   Doing so will not only keep a complete history of the codes but, also make rdas developer could use similar git "pull" command to incorporate new changes to the incorporated codes in GDAS.

SamuelDegelia-NOAA:
@TingLei-NOAA I feel like it is good to get developers used to using the `rdas.x` version so that it isn't a big adjustment down the road. That is why I updated the ctests and run scripts. Plus, I do not think using `./rdas_fv3jedi.x variational` instead of  `./fv3jedi_variational` is significantly less readable or clear. 

Regarding the `git subtree`, this sounds like a good idea going forward. I wasn't aware of this option but it would be great to track changes from the original gdas version.

TingLei-NOAA:
@SamuelDegelia-NOAA  The key is that use of gdas.x or rdas.x is for compliance with current nco requirement (and not the only work-around). It is definitely not contributing to readability/maintainability ( of course, I would not argue how "significantly" it does on the other direction). If needed, changes in the scripts from the original jedi executables to the use of the super executable with corresponding extra parameters  is only a text "finding/replacing" process and I don't think developers needs a time to be familiar with that.  But, starting from such use of that super executable for developers to dig/understand jedi is to add an extra (and unnecessary) wrapping interface.   Of course, that is my opinion.   The decision would be made by rdasapp managing team.  I just want to emphasize that the use of such a super executable is a decision/choice to be made by rdasapp community and not a logical following up after current gdas' example.

SamuelDegelia-NOAA:
@TingLei-NOAA I guess that adding this mega executable can make it harder to track down an issue with a specific application. For example, knowing that you should search for `fv3jedi_variational.x` if you have an issue with the variational component of `rdas.x`.  If others think that the ctests and template scripts should use the older exes then I am fine reverting that part of the PR.

hu5970 (dismissed):
looks good to me

SamuelDegelia-NOAA:
After some discussion today with @TingLei-NOAA and @ShunLiu-NOAA, I reverted the ctests and experiment scripts to use the original executables. Everything else with the PR is the same, but we only plan to use the mega executable once we move to the workflow and working on WCOSS.

ShunLiu-NOAA:
@SamuelDegelia-NOAA Thank you for the update.

ShunLiu-NOAA (approved):

ShunLiu-NOAA:
@hu5970 @guoqing-noaa please review Sam's modification.

rrfs-test/CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ if (CLONE_RRFSDATA)
          set(RESTORE_MPI_ARGS ${MPI_ARGS})
          set(MPI_ARGS "${MPI_ARGS} --exclusive")
       endif()   
+      set(MPI_ARGS "${MPI_ARGS} --ntasks-per-node=18")
```
CoryMartin-NOAA, line 76: Just noting we should make this machine specific "eventually"

CoryMartin-NOAA (approved):

hu5970 (approved):

Issue #104, opened 2024-07-16
Use "ignore=all" for all submodules


guoqing-noaa:
@SamuelDegelia-NOAA added "ignore=dirty" for all submodules following the discussion in issue https://github.com/NOAA-EMC/RDASApp/issues/65. This greatly speeds up normal git operations.

However, I found this is NOT enough. I still get very slow responses from `git status` or `git add` quite often. 
After re-reading [the gitmodules page](https://git-scm.com/docs/gitmodules), it looks like `ignore=dirty` still checks all submodules recursively to see if there are any changes under some situations.  I suggest we use "ignore=all" for all submodules.  

![image](https://github.com/user-attachments/assets/4c651664-e306-4fcf-a0ed-7ca183bbc1e0)
   
<br><br>
I did a few tests and can confirm that `ignore=all` will not prevent us from updating the submodule hashes which is enough for us.
This means `ignore=all` essentially acts as "checkout_externals" in rrfsv1 but with the benefit of being able to quickly navigate to submodule codes just by clicking their links in the RDASApp code page.

SamuelDegelia-NOAA:
@guoqing-noaa I am surprised that `git status` and `git add` is still slow for you. Since adding `ignore=dirty`, my git commands have all been near-instantaneous. But if changing it to `ignore=all` solves your problem then I am happy to switch it. It sounds like switching will have no impact on updating the submodules given your tests. I also just checked and we can still commit and push to the submodules from their `sorc/` directory with `ignore=all`.

guoqing-noaa:
@SamuelDegelia-NOAA Thanks!  I vote to go ahead with `ignore=all`.  
I will write a quick wiki on how to update submodules under the `ignore=all` situation.

SamuelDegelia-NOAA:
@guoqing-noaa Sounds good, I'll go ahead and do a quick PR for this.

guoqing-noaa:
For reference, a wiki page is created about how to work with RDASApp submodules:
https://github.com/NOAA-EMC/RDASApp/wiki/work-with-RDASApp-submodules

Feel tree to make edits.

PR #105, opened 2024-07-17
Further speed up git operations using ignore=all for submodules

SamuelDegelia-NOAA:
Simple fix to solve #104 and further speed up git operations like `git status` or `git add`. According to @guoqing-noaa, this will have no impact on updating the submodules. 

Also, please refer to the [wiki page here about how to update submodules](https://github.com/NOAA-EMC/RDASApp/wiki/work-with-RDASApp-submodules) following these changes. You can still update or commit to them, but will need to go into the individual `sorc/` directories to do so.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,79 +2,79 @@
 	path = sorc/crtm
 	url = https://github.com/jcsda/crtm
 	branch = release/crtm_jedi_v2.4.1
-	ignore = dirty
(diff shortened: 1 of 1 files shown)


guoqing-noaa (approved):

TingLei-NOAA (approved):

ShunLiu-NOAA (approved):

Issue #109, opened 2024-07-19
Update all submodules


hu5970:

hu5970:
Update all submodules based on:
https://github.com/JCSDA/jedi-bundle

hu5970:
Sync with JCSDA/jedi-bundle:
1) sync the submodules: fv3 fv3-jedi fv3-jedi-lm femps gsibec ioda mpas-jedi oops saber ufo vader
2) add fv3-interface.cmake
3) remove submodule fms because it needs to be load through model
4) change compile setups for new fv3 model
5) Keep submodule for crtm as is for this issue. Will update it later.


Tested on Hera, Jet and Orion.

hu5970:
Tested the RRFS cases on Hera, Jet:

The RDAS can be compiled on JET successfully. The test for RRFS cases shows:

```
The following tests passed:
	rrfs_fv3jedi_hyb_2022052619
	rrfs_fv3jedi_letkf_2022052619
	rrfs_mpasjedi_2022052619_Ens3Dvar
	rrfs_mpasjedi_2022052619_letkf

80% tests passed, 1 tests failed out of 5

Label Time Summary:
mpi            = 400.01 sec*proc (5 tests)
rdas-bundle    = 400.01 sec*proc (5 tests)
script         = 400.01 sec*proc (5 tests)

Total Test time (real) = 403.44 sec

The following tests FAILED:
	2156 - rrfs_mpasjedi_2022052619_atms_npp_qc (Failed)
Errors while running CTest
Output from these tests are in: /mnt/lfs4/BMC/rtwbl/mhu/jedi/RDASApp/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```

The reason for failure is because of the location of CRTM coefficient, The same error exists in current test cases too:
```
1996:  CRTM_SpcCoeff_Load(FAILURE) : Error reading SpcCoeff file #1, Data/crtm_2.4.0/atms_npp.SpcCoeff.bin
1996:  ufo_radiancecrtm_setup error: failed CRTM_Load_SpcCoeff
```


The RDAS can be compiled on Hera successfully. The test for rrfs cases shows::

```
The following tests passed:
	rrfs_fv3jedi_hyb_2022052619
	rrfs_fv3jedi_letkf_2022052619
	rrfs_mpasjedi_2022052619_Ens3Dvar
	rrfs_mpasjedi_2022052619_atms_npp_qc
[...]

PR #121, opened 2024-08-04
Sync submodules with the JCSDA/jedi-bundle

hu5970:
This is to address issue #109 .  Sync all submodules with JCSDA/jedi-bundle:

1) sync the submodules: fv3 fv3-jedi fv3-jedi-lm femps gsibec ioda mpas-jedi oops saber ufo vader
2) add fv3-interface.cmake
3) remove submodule fms because it needs to be load through model
4) change compile setups for new fv3 model

Tested on Hera, Jet and Orion. All compiled successfully. All rrfs test cases are success except for the "rrfs_mpasjedi_2022052619_atms_npp_qc" on JET, which had trouble to find the crtm coefficient. This is an error in the current version of RDAS.

The submodule crtm is not synced. Will do it later.


guoqing-noaa:
@hu5970 Besides the CTests, we better run the 2024052700 MPASJEDI  case to see whether it works under your new updates.
To do this, we need to merge PR #91 first.

guoqing-noaa:
It will be good to document which branch, tag, or commit is used for each major component updated in this PR. Thanks!

hu5970:
We need to find a good way to document the hash numbers of each syncing with JCSDA JEDI bundle. But the new commits from each submodule can be checked through each submodule itself.

guoqing-noaa:
I tested Ming's updated RDASApp executables at: 
/scratch1/BMC/wrfruc/mhu/jedi/sync/RDASApp

I first tested the new `mpasjedi_variational.x` file and it can generate similar results as the current RDASApp.
Then I tested the `mpasjedi_error_covariance_toolbox.x` file and found the the bumploc files generated from Ming's PR are different and smaller than the pre-generated one in current RDASApp:
```
ls -lh bumploc_sync/bumploc_401km11levels_nicas_local_000120-000118.nc 
-rw-r--r-- 1 Guoqing.Ge wrfruc 6.2M Aug  6 16:14 bumploc_sync/bumploc_401km11levels_nicas_local_000120-000118.nc
```
```
ls -lh RDASApp.conus12km-401km11levels/bumploc_401km11levels_nicas_local_000120-000118.nc    
-rw-r--r-- 1 role.rrfs-fv3-cam fv3-cam 16M Jun 24 18:58 RDASApp.conus12km-401km11levels/bumploc_401km11levels_nicas_local_000120-000118.nc
```
You can check more between two different sets of bumploc files:
`/scratch1/BMC/wrfruc/gge/RDASApp/expr/mpas_2024052700/data/bumploc/bumploc_sync/`   
vs   
`/scratch1/BMC/wrfruc/gge/RDASApp/expr/mpas_2024052700/data/bumploc/RDASApp.conus12km-401km11levels`   

When using the bumploc files from Ming's PR, it gives different increment patterns. See the figure below, the left side is from current RDASApp, the right side is from Ming's PR:
image

guoqing-noaa:
My opinion is that the left figure looks more reasonable. @TingLei-NOAA @SamuelDegelia-NOAA Do you have any ideas on this? There might be some recent changes in the bumploc generation process?

TingLei-NOAA:
@guoqing-noaa It seems in the _sync dir, the bumploc files contains fewer sampling points (like nsa , I am not sure of its' exact definition, and others).  Maybe those two sets of bumploc files were not generated using the same yaml files? 
If using more sampling points (modeling the B ), it is understandable that it gave more reasonable results (left) while I have to  admit it is difficult for me to tell :).

guoqing-noaa:
@TingLei-NOAA Thanks for checking this. I did use the same yaml file in the generation of two different sets of bumploc files.  

Here is what I did:
clone the latest RDASApp, build, and then enter `rrfs-test/scripts`, use either `link_mpasjedi_expr.sh` or `setup_experiment.sh` to create an expdir directory. Enter the expdir directory, `sbatch run_bump.sh` . Create a directory under data/bumploc, such as `bump_rdas`, move the new 120 bumploc files into this new directory,  then `ln -snf bump_rdas conus12km-401km11levels` under `data/bumploc`. Go back to expdir, `sbatch run_jedi.sh`   
  
To use Ming's executable, create a new expdir directory,  change the `RDASApp` variable in `run_bump.sh` or `run_jedi.sh` to `/scratch1/BMC/wrfruc/mhu/jedi/sync/RDASApp`, and then repeat the above process.

SamuelDegelia-NOAA:
@guoqing-noaa Would you happen to still have the run directories where the bumploc files were created?

guoqing-noaa:
@SamuelDegelia-NOAA Here is the bumploc run directory:
/scratch1/BMC/wrfruc/gge/RDASApp/expr/mpas_2024052700

Thanks!

SamuelDegelia-NOAA:
Thanks @guoqing-noaa for catching this! ~~I ran the bump component and compared only changing the executables. The bumpfiles still have different dimensions, though the differences are not as large as @guoqing-noaa's test. For mine, `nsa=8876` for the current version and `nsa=8484` for the version with updated submodules. Also, the new executable with the updated submodules took much longer to run (~10 min compared to ~3 min for the current version)~~. 

~~Looking at the ---Configuration parameters`section for BUMP in the stdout, it looks like there are some parameters that are added/removed between versions. So it might be worth looking into whether any of those are important.~~

~~But it seems like something might have changed in saber that caused these differences. I also agree that the current version looks more accurate from @guoqing-noaa's plots. So we should hold off on merging this PR until we understand what the difference is here.~~

SamuelDegelia-NOAA:
Whoops... when I cloned Ming's fork I forgot to actually switch to the sync_202408 branch. So the results in my last post did not use the updated submodules. I will rebuild and look into this issue some more.

hu5970:
@guoqing-noaa Please check the iteration numbers of your two tests for BUMP to see if there are difference. The difference could come from minimization numbers. If they are the same, we can  create an issue in:

https://github.com/JCSDA-internal/mpas-jedi/issues

Thanks,
Ming

SamuelDegelia-NOAA:
I reran bump with the correct branch and find similar results to @guoqing-noaa where there is a much lower `nsa` using the exe with updated submodules. 

One thing thing I noticed in the output using the new exe that is not output with the old exe: 

```
  0: Warning  : in geom_from_atlas: cannot find corners, assuming a non-rotated domain, not over pole
  0: Info     :        Work regional grid accuracy:  0.352E+04 km
  0: Info     :           Corner #1: -0.146E+03 (-0.169E+03)  /  0.193E+02 ( 0.443E+02)
  0: Info     :           Corner #2: -0.146E+03 (-0.131E+03)  /  0.612E+02 ( 0.312E+02)
  0: Info     :           Corner #3: -0.507E+02 (-0.308E+02)  /  0.193E+02 ( 0.433E+02)
  0: Info     :           Corner #4: -0.507E+02 ( 0.166E+02)  /  0.612E+02 ( 0.599E+02)
```

And later, the domain area between the two exes is different: 

Old exe: `Domain area (% of Earth area): 0.550E+01%`
New exe: `Domain area (% of Earth area): 0.371E+01%`

SamuelDegelia-NOAA:
I narrowed the BUMP issue down to this PR: https://github.com/JCSDA-internal/saber/pull/866. Reverting to the hash before that reproduces the bumploc files we previously created. So it seems that something changed with the geometry that is impacting our results. 

I tried running the new executable with `resolution: 16` and it produces `nsa=9790` which is close to our original bumploc files with `nsa=8876`. This also produces EnVar increments much closer to our original tests. There are some differences but it is hard to say which one is "correct". Is it worth digging into this more, or should we just increase the resolution on `bumploc.yaml`? 

Increment comparisons:
Capture

guoqing-noaa:
@SamuelDegelia-NOAA Great work on pinpointing the problem! 

I think we should report this issue to PR https://github.com/JCSDA-internal/saber/pull/866 and check whether Benjamin has any thoughts on this. I think a change of resolution from 8 to 16 is not a small change and a clarification on why we need to do this will be beneficial. Thanks!

SamuelDegelia-NOAA:
Sounds good @guoqing-noaa - I'll post a message in that PR.

TingLei-NOAA:
@guoqing-noaa @SamuelDegelia-NOAA  Seems that PR didn't  change results in jedi's own various tests (otherwise, all neccessary changes would have been made before that PR). If that PR is the reason,  I would assume adding more explicit  parameters as in the current jedi's bumploc yaml might help avoid such differences .

guoqing-noaa:
Thanks, @SamuelDegelia-NOAA!

@TingLei-NOAA Good point. But it is also possible that JEDI may not have enough tests for this specific situation.  If we have to change the resolution from 8 to 16 due to that PR, there should exist a reasonable explanation. Or as you point out, we may have to specify some parameters previously ignored.

guoqing-noaa:
@SamuelDegelia-NOAA For your resolution=16 run, what about the bumploc file sizes? Are they similar to what we have in current RDASApp?

SamuelDegelia-NOAA:
@TingLei-NOAA There are some different settings used for the bumploc yaml in mpas-jedi (https://github.com/JCSDA-internal/mpas-jedi/blob/develop/test/testinput/parameters_bumploc.yaml) listed under `sampling:`. Though, the two runs I did had the same values for these parameters (all default values). So they are not impacting this comparison, but setting some of those parameters might be better than increasing the resolution like I did.

SamuelDegelia-NOAA:
@guoqing-noaa No the files with resolution=16 are quite a lot larger - thanks for bringing that up. 

```
ls -lh bumpfiles_origexe/bumploc_401km11levels_nicas_local_000120-000118.nc
-rw-r--r-- 1 Samuel.Degelia stmp 8.3M Aug  9 13:11 bumpfiles_origexe/bumploc_401km11levels_nicas_local_000120-000118.nc
```

```
ls -lh bumpfiles_updatesubmod_res16/bumploc_401km11levels_nicas_local_000120-000118.nc
-rw-r--r-- 1 Samuel.Degelia stmp 49M Aug  9 19:16 bumpfiles_updatesubmod_res16/bumploc_401km11levels_nicas_local_000120-000118.nc
```

49 MB for the files with resolution=16 vs. 8.3 MB for the bumploc files created with the original exe. So yeah resolution=16 might not be the way to go.

hu5970:
@SamuelDegelia-NOAA Thanks for adding our problem with this PR:https://github.com/JCSDA-internal/saber/pull/866 .
Benjamin is on vacation and will be ack later August. We should get some feed back from Benjamin end of this Month.

hu5970:
Anyone has comments on other submodules. I would like to move this PR forward by using role back the saber to the hash before it cause trouble.

hu5970:
@guoqing-noaa @SamuelDegelia-NOAA @CoryMartin-NOAA Do you have any comments or concerns on other submodules other than saber?

guoqing-noaa:
Thanks, @hu5970 !  I think keeping the SABAR component before PR JCSDA-internal/saber#866 is a good idea.  Have you pushed your changes to this PR?

CoryMartin-NOAA:
as long as everything builds together with the hashes you all have selected that works for me

SamuelDegelia-NOAA:
I just tested LETKF and GETKF single obs tests and the results are equivalent after the submodule update.  `bufr2ioda.x` results for the mesonet obs were also identical. So I think the PR is good if we can revert SABER to hash [0a23f67c](https://github.com/JCSDA-internal/saber/tree/0a23f67cca26e634ff271a371c642823c56fee42). 

Reverting to that hash still allows the rest of RDASApp to build successfully.

SamuelDegelia-NOAA:
@ShunLiu-NOAA and @xyzemc Did you want to add anything about the CRTM version?

xyzemc:
> @ShunLiu-NOAA and @xyzemc Did you want to add anything about the CRTM version?

As discussed with Ming and Shun, we will keep the current crtm2.4.0 in RDASApp.

hu5970:
I have reset the saber submodule to hash "0a23f67cc" which is a version before the hash we have trouble with. 
Compilation and RRFS ctests on Here/Jet/Orion are all successful.

hu5970:
@guoqing-noaa @SamuelDegelia-NOAA @CoryMartin-NOAA Please check my latest comments. I think this PR is good to go. We need to continue our JEDI work based on the latest versions of the code.

guoqing-noaa (approved):

CoryMartin-NOAA (approved):

TingLei-NOAA (commented):
Timely update of rdasapp submodules is important!  Thanks

SamuelDegelia-NOAA:
I tested [saber PR #928](https://github.com/JCSDA-internal/saber/pull/928) with @benjaminmenetrier's bugfix and it gives results much closer to what we had before. The domain sizes output by BUMP are also now equal to what we had before and the file sizes are very similar. The `nsa` values are still slightly different, but this is to be expected per Benjamin's response [here](https://github.com/JCSDA-internal/saber/pull/928#issuecomment-2307363406). 

The analysis increments using the new PR are also close to what we had before. There are some differences due to the different `nsa` values, but there is no way to tell which one is "correct": 

Capture

Once [saber PR #928](https://github.com/JCSDA-internal/saber/pull/928) is merged, I think we will be good to update the saber hash.

guoqing-noaa:
@SamuelDegelia-NOAA Looks good! thanks for the update!

PR #141, opened 2024-08-31
streamline the module load process and the rrfs-test process

guoqing-noaa:
Simplify the module load process and the rrfs-test step so that each can be done with one simple command of the following:
```
source ush/load_rdas.sh
```
```
source ush/load_eva.sh
```
```
ush/run_rrfs_test.sh
```
I did encounter a few situations where I tried to `module load RDAS/jet_intel` on Hera or vice versa.
Using a unified simple command can reduce this human error.

Also, this PR combines a few commands needed for doing ctests into one command.
The build_and_test.md has been updated accordingly.


guoqing-noaa:
Since this PR, we can do some kind of streamlined regression tests from a fresh copy. Here is an example:

```
git clone https://github.com/rrfsx/rdas_build_test.git
cd rdas_build_test
./build_test.sh guoqing-noaa simplify rtrr
```
The usage of build_test.sh is as follows:
`./build_test.sh github_name branch_name account_name`  

So in the above example, `guoqing-noaa` is my github account name, `simplify` is my branch name and `rtrr` is my slurm account to be used on Hera.

ShunLiu-NOAA:
@guoqing-noaa This is a good idea. Have you had a chance to test it on HERA and ORION?

guoqing-noaa:
Thanks, @ShunLiu-NOAA 
I tested on Hera and Jet. It worked very well.

guoqing-noaa:
@ShunLiu-NOAA   Current RDASApp breaks on Orion/Hercules (CMake failed). 
I created issue #144 and would hope @SamuelDegelia-NOAA will be able to help with this after the long weekend.

For this PR, it is platform-independent. As long as RDASApp works on a platform, the script will run on that platform.

ush/module-setup.sh
```diff
@@ -22,6 +22,13 @@ elif [[ $MACHINE_ID = orion* ]] ; then
     fi
     module purge
 
+elif [[ $MACHINE_ID = hercules* ]] ; then
+    # We are on Orion
```
TingLei-NOAA: should this be "on Hercules"?

TingLei-NOAA:
> Simplify the module load process and the rrfs-test step so that each can be done with one simple command of the following:
> 
> ```
> source ush/load_rdas.sh
> source ush/load_eva.sh
> ush/run_rrfs_test.sh
> ```
> 
> I did encounter a few situations where I tried to `module load RDAS/jet_intel` on Hera or vice versa. Using a unified simple command can reduce this human error.
> 
> Also, this PR combines a few commands needed for doing ctests into one command. The build_and_test.md has been updated accordingly.

@guoqing-noaa  A general question.  Shall  running tests  which needs "load_rdas.sh" and "running a certain postprocessing which needs "load_eva.sh"  be put into two separate steps?    I am wondering if the python packages installed in those two modules would be on conflict with each other (even, if not now, it could when they are managed separately).  
For the current run_rrfs_test.sh, which is to run ctests,  I think  "load_eva.sh" is not needed to be sourced. .

ush/module-setup.sh
guoqing-noaa: @TingLei-NOAA Thanks for catching this! I will change it as suggested.

guoqing-noaa:
@TingLei-NOAA Good question. I should not put them into one code block to cause misleading information. I just modified them to three separate code blocks in the first comment segment of this PR.

You are right, they are separate.  

If one wants to use the RDAS build env, one can do `source ush/load_rdas.sh`  
If one wants to use the EVA python env, one can do `source ush/load_eva.sh`  
If one wants to run the rrfs-test, one only needs to do `ush/run_rrfs_test.sh`  which will handle the module load automatically.
Thanks!

TingLei-NOAA (approved):

SamuelDegelia-NOAA (approved):
Looks good to me, good idea!

ShunLiu-NOAA:
> Looks good to me, good idea!

So let's merge this PR first and make a test after we fix the issue of CMake failed on Orion.

PR #153, opened 2024-09-06
Update offline domain check script

delippi:
The offline domain check script has been updated.

This program determines if observations are in/outside of a convex hull computed via a lat/lon grid file (see note below about the grid file). A convex hull is the smallest convex shape (or polygon) that can enclose a set of points in a plane (or in higher dimensions). Imagine stretching a rubber band around the outermost points in a set; the shape that the rubber band forms is the convex hull. So, if there are any concave points between vertices, then there would be whitespace between the red and blue box. I've shrunk the convex hull such that there wouldn't be such whitespace which, of course, in tern means that it is going to be not an exact match of the domain grid (e.g., near corners). This can be tuned via the "hull_shrink_factor".

The grid file is what contains variables grid_lat/grid_lon OR latCell/lonCell for FV3 and MPAS respectively.
Examples can be found in the following rrfs-test cases:

- rrfs-data_fv3jedi_2022052619/Data/bkg/fv3_grid_spec.nc
- mpas_2024052700/data/restart.2024-05-27_00.00.00.nc

And example set of obs can be found here:
`/scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2/jedi-assim-aircft/mpas_2024052700/prepbufr.mesonet.nc`

Quick usage guide:
```
module use RDASApp/modulefiles
module load EVA/hera
python offline_domain_check.py -f -o prepbufr.mesonet.nc -g /scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2_demo/jedi-assim_demo/rrfs-data_fv3jedi_2022052619/Data/bkg/fv3_grid_spec.nc
OR
python offline_domain_check.py -f -o prepbufr.mesonet.nc -g /scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2/jedi-assim-aircft/mpas_2024052700/data/restart.2024-05-27_00.00.00.nc
```
The "-f" flag tells the script to also make the plot. Excluding that flag simply turns off plot generation.

I currently just put this in the IODA path where the original tool was. Do we want to keep it there or would it be better to have it in a util directory?

This is the result using the  MPAS grid file:
domain_check_MPAS


SamuelDegelia-NOAA (dismissed):
Looks great, thanks a ton for expanding the check. This will be very helpful!

TingLei-NOAA (dismissed):
A wonderful tool! Thanks!

guoqing-noaa (dismissed):
@delippi Thanks for expanding the offline domain check tool! LGTM.

I have two minor comments or questions:
1. Can we make this file executable and hence don't need to add "python" before running this tool?
2.  what is "-f" abbreviated from?

guoqing-noaa:
For my first question, you can add the following line at the beginning of the file:  
`#!/usr/bin/env python`   
and then `chmod +x` the file
Thanks!

delippi:
@guoqing-noaa,  

> @delippi Thanks for expanding the offline domain check tool! LGTM.
> 
> I have two minor comments or questions:
> 
> 1. Can we make this file executable and hence don't need to add "python" before running this tool?

done.
> 2. what is "-f" abbreviated from?

--fig

guoqing-noaa:
Thanks, @delippi

guoqing-noaa (approved):

Issue #160, opened 2024-09-10
update the mpas_2024052700 case to use the mpasout files


guoqing-noaa:
@chunhuazhou has generated background and 30 member mpasout files which include `u`, `uReconstructZonal`, `uReconstructMeridional` variables.

A PR will be created soon to update the mpas_2024052700 case to use these mapsout files.

Hope this helps the further investigation of issue https://github.com/NOAA-EMC/RDASApp/issues/148

guoqing-noaa:
For reference, this is the ensemble spread at level=23:
![image](https://github.com/user-attachments/assets/f006ffd5-b836-4175-bdae-c029555e9be2)

guoqing-noaa:
Here are the analysis increments from a single observation test:
![image](https://github.com/user-attachments/assets/8d23b937-dd39-4e07-a50b-0bc8474d4eb6)

guoqing-noaa:
Here are the analysis increment from the bumploc files generated from the mpasout files:
The max value is changed from 1.68 to 1.64 and there are small pattern changes as well when you zoom in.

![image](https://github.com/user-attachments/assets/e6120356-5d41-484c-a847-8b2e5fad9ce7)

guoqing-noaa:
@SamuelDegelia-NOAA @TingLei-NOAA  If you want to take a close examination of the two sets of bumploc files, they are at:
`/mnt/lfs5/BMC/wrfruc/gge/RDASApp/expr/mpas_2024052700/data/bumploc`  

Thanks!

chunhuazhou:
@guoqing-noaa The small difference might be from the model change to tag v8.2.1, which I used for the mpasout/mpasin testing previously.

guoqing-noaa:
> @guoqing-noaa The small difference might be from the model change to tag v8.2.1, which I used for the mpasout/mpasin testing previously.

The differenences between the above two analysis increment fields only come from using different bumploc files (the bkg and ensembles are the same)
@TingLei-NOAA @SamuelDegelia-NOAA Does the bump localization depend on the model meteorological fields? Correct me if I am wrong, but I had an impression that they are invariant with changed meteorological fields.

SamuelDegelia-NOAA:
> @TingLei-NOAA @SamuelDegelia-NOAA Does the bump localization depend on the model meteorological fields? Correct me if I am wrong, but I had an impression that they are invariant with changed meteorological fields.

[...]

Issue #19, opened 2024-04-02
building error on Orion


ShunLiu-NOAA:
When I build RDASApp on Orion, I got the error message below:

-- Adding bundle project ioda-data
CMake Error at /work/noaa/epic/role-epic/spack-stack/orion/spack-stack-1.6.0/envs/unified-env/install/intel/2022.0.2/ecbuild-3.7.2-3n22kc3/share/ecbuild/cmake/ecbuild_log.cmake:190 (message):
  CRITICAL - Source directory
  '/work/noaa/da/Shun.Liu/noscrub/JEDI/RDASApp/bundle/ioda-data' for
  subproject 'ioda-data' does not exist or does not contain a CMakeLists.txt
  file.
Call Stack (most recent call first):
  /work/noaa/epic/role-epic/spack-stack/orion/spack-stack-1.6.0/envs/unified-env/install/intel/2022.0.2/ecbuild-3.7.2-3n22kc3/share/ecbuild/cmake/ecbuild_bundle.cmake:166 (ecbuild_critical)
  CMakeLists.txt:115 (ecbuild_bundle)

It appears the directory below is not available in Cory's account anymore. 
lrwxrwxrwx 1 sliu da   54 Apr  2 11:18 ioda-data -> /work2/noaa/da/cmartin/CI/RDASApp/data/jcsda/ioda-data

CoryMartin-NOAA:
@ShunLiu-NOAA thanks, I'll fix this later today

CoryMartin-NOAA:
If you build without `-d` flag in the build script it should at least compile.

ShunLiu-NOAA:
Thanks. I completes "build" without "-d" on Orion.

CoryMartin-NOAA:
Please @ShunLiu-NOAA or someone else try and see if this issue is resolved now. Thanks!

ShunLiu-NOAA:
Tested on Orion. The issue is resolved. Thanks.

PR #161, opened 2024-09-10
update the mpas_2024052700 case to use mpasout files

guoqing-noaa:
Use `mem001` instead of `mem01` to be consistent with the workflow

Address issue #160


guoqing-noaa:
@Junjun-NOAA @delippi 
If you are interested, you may try rerunning your case using the mpasout files in this PR to see whether you can get wind increments from the assimilation of T obs.

delippi:
@guoqing-noaa, yes, I already started. I was able to assimilate winds and get temperature increments (slightly different from what others are doing). I will have some suggestions in a review that I will work on tomorrow.

rrfs-test/scripts/setup_experiment.sh
```diff
@@ -98,8 +98,8 @@ elif [[ $DYCORE == "MPAS" ]]; then
   mkdir -p data; cd data
   mkdir -p bumploc bkg obs ens
   ln -snf ${YOUR_PATH_TO_RDASAPP}/fix/bumploc/* bumploc/.
-  ln -snf ${YOUR_PATH_TO_RDASAPP}/fix/expr_data/${TEST_DATA}/bkg/restart.2024-05-27_00.00.00.nc .
-  ln -snf ${YOUR_PATH_TO_RDASAPP}/fix/expr_data/${TEST_DATA}/bkg/restart.2024-05-27_00.00.00.nc static.nc
+  ln -snf ${YOUR_PATH_TO_RDASAPP}/fix/expr_data/${TEST_DATA}/bkg/mpasout.2024-05-27_00.00.00.nc .
+  ln -snf ${YOUR_PATH_TO_RDASAPP}/fix/expr_data/${TEST_DATA}/bkg/static.nc static.nc
```
delippi, line 102: `ln -snf ${YOUR_PATH_TO_RDASAPP}/fix/expr_data/${TEST_DATA}/bkg/static.nc static.nc` should be `ln -snf ${YOUR_PATH_TO_RDASAPP}/fix/expr_data/${TEST_DATA}/static.nc static.nc` (the `bkg` removed)

delippi (changes requested):
I found one small change that is needed. In `setup_experiment.sh` the path to `static.nc` is slightly incorrect. I noted what it should be in that file. Other than that this PR looks good to me. I did run everything using the new backgrounds and I can confirm that I was able to assimilate wind observations and was able to have increments of temperature. I was going to share my results, because I thought they were strange, but I just realized there might be issues with `data/obs`.

Edit: Actually, I think there are just issues with `data/obs/ioda_mesonet.nc` which I didn't test (due to incorrect data types and some other things--perhaps an old yaml was used to create that?). The aircraft data is correct.

rrfs-test/scripts/link_mpasjedi_expr.sh
```diff
@@ -44,10 +44,10 @@ cp ${RDASApp}/rrfs-test/ush/mpasjedi_spread.py .
 
 mkdir -p data
 cd data
-mkdir -p bumploc bkg obs ens ref
+mkdir -p bumploc obs ens ref
 ln -snf ${RDASApp}/fix/bumploc/${BUMPLOC} bumploc/${BUMPLOC}
-ln -snf ${RDASApp}/fix/expr_data/${exprname}/bkg/restart.2024-05-27_00.00.00.nc .
-ln -snf ${RDASApp}/fix/expr_data/${exprname}/bkg/restart.2024-05-27_00.00.00.nc static.nc
+ln -snf ${RDASApp}/fix/expr_data/${exprname}/bkg/mpasout.2024-05-27_00.00.00.nc .
+ln -snf ${RDASApp}/fix/expr_data/${exprname}/static.nc static.nc
```
TingLei-NOAA, line 50: could those static files in the source location be added  qualifiers of  grid-related labels?

HuiLiu-NOAA:
@guoqing-noaa : just tested the new case and got an error with the files:

17: !!! Hint: group interp_c1b_to_c0a_002
 17:        |-> cmp_1
 17:           |-> common
 17:              |-> bumploc_401km11levels_nicas
 17: Abort(1) on node 17 (rank 17 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 1) - process 17
 18: !!! NetCDF error in netcdf_inquire_grp on task #000019: NetCDF: No group found.

ShunLiu-NOAA:
@guoqing-noaa Do you have a chance to address @HuiLiu-NOAA and @TingLei-NOAA's comments?

guoqing-noaa:
@ShunLiu-NOAA Thanks for the reminder. I will address Hui and Ting's comments.

But the big concern is, as discussed with @SamuelDegelia-NOAA in this thread, the transition to mpasout files changed the bumploc files. This is unexpected and we will need more time to dig into this issue.

TingLei-NOAA:
@guoqing-noaa  Could you point to your run dirs to create bump files using restart files (old) and mpasout files (current) respectively?  This will allow us to reproduce your results and join the investigation into this issue.

guoqing-noaa:
> @guoqing-noaa Could you point to your run dirs to create bump files using restart files (old) and mpasout files (current) respectively? This will allow us to reproduce your results and join the investigation into this issue.

@TingLei-NOAA Thanks for helping with this issue.

I ran the mpasout version on Jet and it is at:
`/lfs5/BMC/wrfruc/gge/RDASApp/expr/mpas_2024052700`   

I ran the restart version on Hera and it is at:
`/scratch1/BMC/wrfruc/gge/RDASApp/expr/mpas_2024052700`  

If you want to reproduce the case comparison,  you can check two versions of RDASApp: NOAA-EMC:develop and guoqing-noaa:mpasout, generate bumploc files respectively, run the single obs jedi case, plot the increments.

I can also regenerate them at a later time (maybe weekend).

TingLei-NOAA:
@guoqing-noaa   Thanks. I had copied them to my account and will see what the differences are. 
An update:  Guqing, would you point me to the bumploc.yaml used in your hera run? seems it is missing in the dir you pointed.

TingLei-NOAA:
@guoqing-noaa  My runs showed identical results using mpasout or restart files. 
What I did is as follows:
```
I changed the bumploc.yaml for jet case to be used for hera case with another variables changes which I believe shouldn't affect results (but we could check on it later if needed) . 
The two files are in https://github.com/TingLei-NOAA/RDASApp/tree/feature/for-test/rrfs-test/example-files. 
I ran both cases on hera using the executable: mpasjedi_error_covariance_toolbox.x from https://github.com/NOAA-EMC/RDASApp/pull/175.  I used it because its executables are already ready on hera.   
Using "diff" to compare the generated "bumploc_401km11levels_nicas_local*.nc" from the two runs and no differences were found. 
```

This confirms our pre-assumption that using mpasout or restart files should not change the output in the nicas files (while, there could be still other possibility like the differences in the vertical coor. to cause differences), which is , at least, understandable. 

So, the differences you found might come from other differences.  For further investigation, you could run those two cases also on the same machine , like hera. 
(my copy of your jet case on hera is /scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-guoqing/dr-jet/mpas_2024052700).
 
 We could first see you could get the same comparison results when running them on the same machine.

ShunLiu-NOAA:
> @guoqing-noaa My runs showed identical results using mpasout or restart files. What I did is as follows:
> 
> ```
> I changed the bumploc.yaml for jet case to be used for hera case with another variables changes which I believe shouldn't affect results (but we could check on it later if needed) . 
> The two files are in https://github.com/TingLei-NOAA/RDASApp/tree/feature/for-test/rrfs-test/example-files. 
> I ran both cases on hera using the executable: mpasjedi_error_covariance_toolbox.x from https://github.com/NOAA-EMC/RDASApp/pull/175.  I used it because its executables are already ready on hera.   
> Using "diff" to compare the generated "bumploc_401km11levels_nicas_local*.nc" from the two runs and no differences were found. 
> ```
> 
> This confirms our pre-assumption that using mpasout or restart files should not change the output in the nicas files (while, there could be still other possibility like the differences in the vertical coor. to cause differences), which is , at least, understandable.
> 
> So, the differences you found might come from other differences. For further investigation, you could run those two cases also on the same machine , like hera. (my copy of your jet case on hera is /scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-guoqing/dr-jet/mpas_2024052700).
> 
> We could first see you could get the same comparison results when running them on the same machine.

@guoqing-noaa do you get a chance to take a look at this?

guoqing-noaa:
Convert this to draft PR as there are lingering issues and we may update to the 66 levels directly and skip this PR.

SamuelDegelia-NOAA:
Does #194 supersede this PR? Should we close this?

guoqing-noaa:
> Does #194 supersede this PR? Should we close this?

Yes and good suggestion, @SamuelDegelia-NOAA 

@ShunLiu-NOAA, I quoted @TingLei-NOAA's replies below and thanks @TingLei-NOAA for the testing!

Yes, as Ting suggested, the differences came from the RDASApp hash I used is different from that was used to generate the staged bumploc files.  The `mpasout` and `restart` files themselves will not cause any differences in the bumploc files.

> @guoqing-noaa My runs showed identical results using mpasout or restart files. What I did is as follows:
> 
> ```
> I changed the bumploc.yaml for jet case to be used for hera case with another variables changes which I believe shouldn't affect results (but we could check on it later if needed) . 
> The two files are in https://github.com/TingLei-NOAA/RDASApp/tree/feature/for-test/rrfs-test/example-files. 
> I ran both cases on hera using the executable: mpasjedi_error_covariance_toolbox.x from https://github.com/NOAA-EMC/RDASApp/pull/175.  I used it because its executables are already ready on hera.   
> Using "diff" to compare the generated "bumploc_401km11levels_nicas_local*.nc" from the two runs and no differences were found. 
> ```
> 
> This confirms our pre-assumption that using mpasout or restart files should not change the output in the nicas files (while, there could be still other possibility like the differences in the vertical coor. to cause differences), which is , at least, understandable.
> 
> So, the differences you found might come from other differences. For further investigation, you could run those two cases also on the same machine , like hera. (my copy of your jet case on hera is /scratch2/NCEPDEV/fv3-cam/Ting.Lei/dr-guoqing/dr-jet/mpas_2024052700).
> 
> We could first see you could get the same comparison results when running them on the same machine.

ShunLiu-NOAA:
Yes. We should close this one. @guoqing-noaa I did mention in this PR that we may consider keeping the restart file name.

guoqing-noaa:
Close this PR as discussed. Thanks!

PR #147, opened 2024-09-05
Centralize test data and reduce unnecessary copying of test data 

guoqing-noaa:
This PR is the first step to centralize test data and reduce unnecessary copying.
It addresses issue #93 

As a byproduct, it also addresses issue #127 

All these will reduce the unnecessary turnaround time to speed up the build process and save disk space.

PR #162, opened 2024-09-10
Fix build error on Orion by reducing the number of build tasks

### LINKED ISSUE #146: Building fails for fv3jedi or mpasjedi executables on Orion
When first running `build.sh` on Orion, the build fails when trying to build `rdas_fv3jedi.x` and `rdas_mpasjedi.x`. Restarting `build.sh` a second time allows these executables to be built successfully. I am looking into the problem.

SamuelDegelia-NOAA:
This PR resolves #146 by updating `build.sh` to use a lower number of parallel build tasks on Orion. Previously, the build would fail towards the end due to going over the memory limit for the login nodes. This update does increase the total build time (~70 min previously to ~100 min now), but at least it works now without having to restart the build multiple times. Note that this memory limit is not a problem on any other machine, including Hercules. 

Thanks @CoryMartin-NOAA for the tip here.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -137,6 +137,13 @@ if [[ $DYCORE == 'MPAS' || $DYCORE == 'FV3andMPAS' ]]; then
   $dir_root/rrfs-test/scripts/link_mpasjedi_expr.sh
 fi
 
+# Set lower number of build jobs on Orion due to memory limit on login nodes
+if [[ $BUILD_TARGET == 'orion' ]]; then
+  BUILD_JOBS=${BUILD_JOBS:-4}
+else # hera, hercules, jet
+  BUILD_JOBS=${BUILD_JOBS:-6}
+fi
+
 CMAKE_OPTS+=" -DMPIEXEC_MAX_NUMPROCS:STRING=120"
 # Configure
 echo "Configuring ..."
@@ -150,11 +157,11 @@ set +x
 echo "Building ..."
 set -x
 if [[ $BUILD_JCSDA == 'YES' ]]; then
-  make -j ${BUILD_JOBS:-6} VERBOSE=$BUILD_VERBOSE
+  make -j $BUILD_JOBS VERBOSE=$BUILD_VERBOSE
 else
   for b in $builddirs; do
     cd $b
-    make -j ${BUILD_JOBS:-6} VERBOSE=$BUILD_VERBOSE
+    make -j $BUILD_JOBS VERBOSE=$BUILD_VERBOSE
     cd ../
   done
 fi


ShunLiu-NOAA (approved):
Looks good to me.

TingLei-NOAA (dismissed):

guoqing-noaa (changes requested):
@SamuelDegelia-NOAA , thanks for finding a solution for the build failure on Orion.
Generally we would avoid hardwiring the number of processes to be used by build.sh.
It would be expected to be passed from the command line, such as `build.sh -j8`.

Also, are you okay that I merge PR #147 first as there are lots of changes in build.sh there?  Thanks!

build.sh
```diff
@@ -155,6 +155,13 @@ if [[ $BUILD_TARGET == 'hera' ]]; then
   ln -sf $RDASAPP_TESTDATA/crtm $dir_root/bundle/test-data-release/crtm
 fi
 
+# Set lower number of build jobs on Orion due to memory limit on login nodes
+if [[ $BUILD_TARGET == 'orion' ]]; then
+  BUILD_JOBS=4
```
CoryMartin-NOAA: How about making these 
`BUILD_JOBS=${BUILD_JOBS:-4}`?

SamuelDegelia-NOAA:
> @SamuelDegelia-NOAA , thanks for finding a solution for the build failure on Orion. Generally we would avoid hardwiring the number of processes to be used by build.sh. It would be expected to be passed from the command line, such as `build.sh -j8`.

I am fine with that but I know we have made some edits over the last month to simplify the amount of user input needed for `build.sh`. The previous version of the script used `make -j ${BUILD_JOBS:-6}` which, if I am understanding correctly, would use 6 jobs unless the `${BUILD_JOBS}` env variable was set. So following Cory's recommendation above, we can set `BUILD_JOBS=${BUILD_JOBS:-4}` for Orion and `BUILD_JOBS=${BUILD_JOBS:-6}` for other machines. This would set a default value but does not technically hard code it so that it could still be changed by setting the env variable. Does that work for you? 

> Also, are you okay that I merge PR #147 first as there are lots of changes in build.sh there? Thanks!

Yes that is fine with me.

build.sh
SamuelDegelia-NOAA: Yes I think that would be a good idea, thank you.

guoqing-noaa:
@SamuelDegelia-NOAA Maybe we can take this opportunity to update build.sh for the changed sitution:
1.  remove `-t  target to build for <target>"`  as we detect machine_id automatically and don't need this parameter any more.
    
2. Do we have situations where we only want to build partially for only the following directories?
`fv3-jedi mpas-jedi iodaconv`
If we don't need this, we may safely remove the `-s  only build a subset of the bundle` option
    
3. add an option `-j number_of_processes for building` so that users can specify how many processes easily at the command line, such as `build.sh -j8` on Hera/Jet/hercules and `build.sh -j4` on Orion.
    
4. we may want to make the rrfs(mpas) test as a mandate since this is a rrfs(mpas) system and every new PRs should pass all rrfs(mpas) tests. Also, we now don't copy binary data but link them instead, so the extra cost should be negligible. If we agree with this, we can remove the `ADD_RRFS_TESTS` variable and always add rrfs tests.

SamuelDegelia-NOAA:
> @SamuelDegelia-NOAA Maybe we can take this opportunity to update build.sh for the changed sitution:
> 
> 1. remove `-t  target to build for <target>"`  as we detect machine_id automatically and don't need this parameter any more.
> 2. Do we have situations where we only want to build partially for only the following directories?
>    `fv3-jedi mpas-jedi iodaconv`
>    If we don't need this, we may safely remove the `-s  only build a subset of the bundle` option
> 3. add an option `-j number_of_processes for building` so that users can specify how many processes easily at the command line, such as `build.sh -j8` on Hera/Jet/hercules and `build.sh -j4` on Orion.
> 4. we may want to make the rrfs(mpas) test as a mandate since this is a rrfs(mpas) system and every new PRs should pass all rrfs(mpas) tests. Also, we now don't copy binary data but link them instead, so the extra cost should be negligible. If we agree with this, we can remove the `ADD_RRFS_TESTS` variable and always add rrfs tests.

@guoqing-noaa I think these are all good suggestions. I am happy to further modify this PR to incorporate those changes. However, I would like to just double check with @CoryMartin-NOAA on the intended functionality of the `-s` option and associated `BUILD_JCSDA` and `builddirs` variables. When would we not need to build the JSCDA libraries? And is anyone using that functionality?

CoryMartin-NOAA:
The main functionality of the `-s` argument is that it does save some compile time by not compiling all executables that may not be used. If we have the single `rdas.x` executable, and compile that, it will definitely be faster (albeit not by much) than compiling fv3jedi_var.x, fv3jedi_letkf.x, etc.

This is basically to have the capability to build two ways, one for active development, and one for the workflow for cycling

SamuelDegelia-NOAA:
Thank you @CoryMartin-NOAA, I think it is worth leaving this functionality then. I will modify this PR to incorporate points (1,3,4) from @guoqing-noaa.

ShunLiu-NOAA:
> @SamuelDegelia-NOAA Maybe we can take this opportunity to update build.sh for the changed sitution:
> 
> 1. remove `-t  target to build for <target>"`  as we detect machine_id automatically and don't need this parameter any more.
_I think we can remove -t option._
> 2. Do we have situations where we only want to build partially for only the following directories?
>    `fv3-jedi mpas-jedi iodaconv`
>    If we don't need this, we may safely remove the `-s  only build a subset of the bundle` option
It will be a good function if we can build the package partially to save building/compiling time.

> 3. add an option `-j number_of_processes for building` so that users can specify how many processes easily at the command line, such as `build.sh -j8` on Hera/Jet/hercules and `build.sh -j4` on Orion.
_Let's stay on the current settings with the default task number for the different HPCs. It may be easy for debugging work if one encounters any compiling issues._ 

> 4. we may want to make the rrfs(mpas) test as a mandate since this is a rrfs(mpas) system and every new PRs should pass all rrfs(mpas) tests. Also, we now don't copy binary data but link them instead, so the extra cost should be negligible. If we agree with this, we can remove the `ADD_RRFS_TESTS` variable and always add rrfs tests.
_It is too early to use mpas as the only rrfs test. Let's discuss this later._ 

For this PR, if there is no other concerns, I will merge it with the existing form. We can have another PR to address these suggestions.

SamuelDegelia-NOAA:
@ShunLiu-NOAA I think this PR should be merged ASAP just because without it the building will almost always fail on Orion. But I can prepare a PR for those other suggestions and get it submitted relatively soon.

ShunLiu-NOAA (dismissed):

build.sh
```diff
@@ -155,6 +155,13 @@ if [[ $BUILD_TARGET == 'hera' ]]; then
...
+else # hera, hercules, jet
+  BUILD_JOBS=6
+fi
+
```
guoqing-noaa, line 146: @SamuelDegelia-NOAA I think we will not need line 158-163 if we add an option "-j" and then we can keep line 171 and line 175.
If this is urgent, we may add "-j" first and update build_and_test.md

guoqing-noaa:
@SamuelDegelia-NOAA We need to resolve the following conflict first.

image

SamuelDegelia-NOAA:
I updated the PR to use @CoryMartin-NOAA's suggestion so that the build tasks can still be set by an environment variable `${BUILD_JOBS}`. So @TingLei-NOAA and @ShunLiu-NOAA can re-review. 

This PR is urgent in my opinion to resolve building errors on one of our supported machines. Since the changes here are now consistent with our original build script in RDASApp, I think it should be merged soon. Then, we can submit a separate PR to clean up the various options as suggested by @guoqing-noaa.

SamuelDegelia-NOAA:
Will resolve the conflicts now.

build.sh
CoryMartin-NOAA: you can also do, without even a PR:
`BUILD_JOBS=4 ./build.sh`

SamuelDegelia-NOAA:
Sorry, I made a mistake in resolving the commits that closed this PR. Trying to reopen it now.

build.sh
guoqing-noaa: @CoryMartin-NOAA  Yes, that works. 
It is rare, but there are still some guys using `csh`, :)
So `build.sh -j4` may be simpler. Just my 2 cents.

SamuelDegelia-NOAA:
Okay merge conflicts have been resolved. Is anyone opposed to just merging this as-is and then we can have a separate PR to implement the various other changes suggested by @guoqing-noaa? That way we can get things working ASAP on Orion.

ShunLiu-NOAA (approved):

guoqing-noaa (approved):
Not preferred, but this does no harm either.

build.sh
SamuelDegelia-NOAA: I can add the `-jN` option in a future PR. Also, in my opinion I like keeping the default option in the script for Orion since that machine will (likely) always need the lower value. We could add info to the readme to do something like `BUILD_JOBS=4 ./build.sh` but I think someone could miss that, and keeping it as simple as just running `build.sh` would be preferable.

PR #172, opened 2024-09-16
change the HERA's RDAS_DATA location to be under the role.rrfsfix account

guoqing-noaa:
This is to make things consistent with Jet, Gaea and rrfs-workflow.

The role.rrfsfix account has special environments to facilitate the use of the '[git-mega](https://github.com/git-mega/git-mega)' tool for the level 2 management of fix files (_note: only the fix file management team needs to deal with git-mega_)

The role.rrfsfix account is used only for the fix file management and will NOT run any jobs so as to ensure full data integrity.

### DIFF
--- a/ush/init.sh
+++ b/ush/init.sh
@@ -6,7 +6,7 @@ basedir="$(dirname "$ushdir")"
 
 case ${MACHINE_ID} in
   hera)
-    RDAS_DATA=/scratch1/NCEPDEV/fv3-cam/RDAS_DATA
+    RDAS_DATA=/scratch2/BMC/rtrr/RDAS_DATA
     ;;
   jet)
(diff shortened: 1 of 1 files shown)


ShunLiu-NOAA:
@guoqing-noaa As we discussed in the email, it is not necessary to combine rrfs fix files and rdas test case.

guoqing-noaa:
@ShunLiu-NOAA Yes, we don't have to make this change.
But as we use "role.rrfsfix" for Jet, Gaea C5 and C6 for RDAS_DATA and for the HPSS archive of RDAS_DATA, this change will make things consistent.

guoqing-noaa:
close this PR per conversation

Issue #173, opened 2024-09-17
MPAS plot scripts plot incorrect domain shapes


delippi:
The plotting scripts used for visualizing data on an MPAS grid is producing incorrect plots due to issues with handling concave boundaries. The scripts are intended to create contour plots for temperature increments, spread, etc., but the visual output is distorted around the concave areas of the domain.

For example, here is the ensemble spread plot with lat/lon points overlaid. The scatter points (blue) show what the domain should look like while the contour fill is trying to fill in areas that are not actually part of the domain. This can lead to confusion when interpreting plots.
![spread_airTemperature](https://github.com/user-attachments/assets/c59d95a4-69f9-4f51-8841-6eaa8c2a86ff)

This issue will track progress with updating the plotting scripts.

delippi:
I'm able to get the plot to show the correct domain shape by using:
```
# Plot using Delaunay triangulation
triang = Triangulation(lons, lats, tri.simplices)
mask = TriAnalyzer(triang).get_flat_tri_mask(min_circle_ratio=0.1)
triang.set_mask(mask)

# Plot the variance data on the triangulation
c1 = m1.tricontourf(triang, var, cmap=cmap, levels=100, transform=ccrs.PlateCarree())
```

Left is before. Right is after.
![merged_image](https://github.com/user-attachments/assets/65e7884b-003b-4be4-8382-2825c0812045)

guoqing-noaa:
Great work. Thanks, @delippi!

delippi:
And working for the increment plot as well.

Edit: Image updated

![merged_increment_airTemperature](https://github.com/user-attachments/assets/ac3d05b5-dee4-4d7b-8f92-08ad107a6895)

delippi:
I just used scatter(lon,lat) for plotting domain area of offline domain check. We also see why 10% hull shrink is required to remove all obs that could be outside the domain (especially in the south part of the domain).
![merged_domain_check_MPAS_aircraft](https://github.com/user-attachments/assets/bb75a4cb-e406-425b-932e-7a1ffaccf28b)

SamuelDegelia-NOAA:
Thanks @delippi this is a good find. BTW, one of my versions of the offline domain check uses a concave hull approach. That could help remove any obs inside the concave section without needing to shrink it further and remove valid obs. But that code has a problem of needing python modules that are not available with any of the RDASApp modulefiles

delippi:
[...]

PR #174, opened 2024-09-17
updated the mpas plotting scripts

delippi:
Updated the mpas plotting scripts (the domain check also has plotting script built in) to correctly plot mpas domain edges. Prior to this update the concave boundaries would be filled by the plotting routines incorrectly leading to potential misinterpretation of data. The scripts were also cleaned up a little.

This fixes the following issue: https://github.com/NOAA-EMC/RDASApp/issues/173

### DIFF
--- a/rrfs-test/IODA/offline_domain_check.py
+++ b/rrfs-test/IODA/offline_domain_check.py
@@ -59,23 +59,26 @@ parser = argparse.ArgumentParser()
 parser.add_argument('-g', '--grid', type=str, help='grid file', required=True)
 parser.add_argument('-o', '--obs', type=str, help='ioda observation file', required=True)
 parser.add_argument('-f', '--fig', action='store_true', help='disable figure (default is False)', required=False)
+parser.add_argument('-s', '--shrink', type=float, help='hull shrink factor', required=True)
 args = parser.parse_args()
 
 # Assign filenames
 obs_filename = args.obs
 grid_filename = args.grid  # see note above.
 make_fig = args.fig
+hull_shrink_factor = args.shrink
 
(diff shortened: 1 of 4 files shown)


ShunLiu-NOAA (dismissed):
Looks good to me. Thank you for the update.

SamuelDegelia-NOAA:
Nice solution @delippi. It looks like these python scripts assume "mpasout" files for the backgrounds. Do you want to hold off on merging this until #161 is merged? Also out of curiousity, what is reason for the difference in the increment vs. spread plots where the latter uses the extra `tri.simplices` argument in the `Triangulation` function?

delippi:
> Nice solution @delippi. It looks like these python scripts assume "mpasout" files for the backgrounds. Do you want to hold off on merging this until #161 is merged? Also out of curiousity, what is reason for the difference in the increment vs. spread plots where the latter uses the extra `tri.simplices` argument in the `Triangulation` function?

Good catch. That was not needed. I will update accordingly. That just allowed you to specify the triangulation. It works fine without that option.

guoqing-noaa:
@delippi Maybe you can change "mpasout" back to "restart" and test it with the latest RDASApp?
PR #161 will need more time to investigate. We would like to get your PR merged first.

delippi:
@SamuelDegelia-NOAA & @guoqing-noaa, I've reverted to the use of restart.nc instead of mpasout.nc and checked that it still works.

SamuelDegelia-NOAA (approved):

ShunLiu-NOAA (approved):

guoqing-noaa (approved):

PR #182, opened 2024-09-26
Prepbufr yamls from rrfs-workflow/rrfs-mpas-jedi

delippi:
These are the prepbufr*yaml from rrfs-workflow's rrfs-mpas-jedi branch. These are just starting points so that updates to them can be tracked.


guoqing-noaa (approved):
Thanks, @delippi

ShunLiu-NOAA (approved):

Issue #122, opened 2024-08-05
Efficiency issues in EnKF solvers for MPAS-JEDI due to vertical localization


SamuelDegelia-NOAA:
When assimilating the full set of RAP mesonet obs, I found significant slowdown in LETKF for MPAS-JEDI. Using 120 MPI tasks, the LETKF took ~70 min to run. 

I found that this is primarily due to vertical localization. After disabling vertical localization, the LETKF runs in ~24 min. The bottleneck now is hofx calculation, which could be sped up by running LETKF in split observer/solver mode for each member separately, and then catting the hofx file together. I do not think this will have any affect on the vertical localization issue, though. 

It seems that this is a known issue due to the 3d iterator. Some discussion of the problem is [here](https://github.com/JCSDA-internal/mpas-jedi/issues/840), but no solution is available yet other than using GETKF for model-space localization (which has other known efficiency issues).

guoqing-noaa:
Just curious,  is the total time of observer+solver much smaller than running LETKF in one step directly?

SamuelDegelia-NOAA:
@guoqing-noaa, I will work on using this split method for MPAS-JEDI and get back to you. I only ever tried it for FV3-JEDI and did not notice any significant difference in the timing. But I was mainly working on it there to see if the split method could solve the halo observation distribution issue in #51 and did not pay very close attention to the cost.

guoqing-noaa:
Great! Thanks, @SamuelDegelia-NOAA !

Do you have an instruction on how to run LETKF? A few GSL folks (@chunhuazhou, @hongli-wang ,  @HaidaoLin-NOAA, etc) may want to learn and get some hands on experiences on that.

SamuelDegelia-NOAA:
I do not really have any direct instructions, but the yaml file included in RDASApp (https://github.com/NOAA-EMC/RDASApp/blob/develop/rrfs-test/testinput/rrfs_mpasjedi_2022052619_letkf.yaml) works to run LETKF for our older test case. No additional fix files or changes are needed in the run directory compared to running EnVar with `setup_experiment.sh`. Other than changing the yaml file, the only difference compared to EnVar is that you change the executable from `mpasjedi_variational.x` to `mpasjedi_enkf.x`. 

[...]

PR #66, opened 2024-06-10
Added yaml file that include bias correction for atms_npp.

xyzemc:
Addressing [issue#52](https://github.com/NOAA-EMC/RDASApp/issues/52 ), this pull request include a new yaml file includes the bias correction part on top of rrfs_mpasjedi_2022052619_atms_npp_qc.yaml. This new yaml file named as 'rrfs_mpasjedi_2022052619_atms_npp_qc_**bc**.yaml'
Also, three fixed files have been added into the test data Data/obs directory: 
- atms_npp.satbias.nc4
- atms_npp.satbias_cov.nc4
- atms_npp.tlapse.txt

There will be a new output directory Data/bc, which contains two output bias files 
- out_atms_npp.satbias.nc  
- out_atms_npp.satbias_cov.nc

The ctest has been successfully passed, the test result as bellow:

> [Xiaoyan.Zhang@hfe03 rrfs-test]$ ctest
Test project /scratch2/NCEPDEV/fv3-cam/Xiaoyan.Zhang/noscrub/JEDI/RDASApp_atms/build/rrfs-test
    Start 1: rrfs_mpasjedi_2022052619_Ens3Dvar
1/4 Test #1: rrfs_mpasjedi_2022052619_Ens3Dvar .........   Passed   90.76 sec
    Start 2: rrfs_mpasjedi_2022052619_letkf
2/4 Test #2: rrfs_mpasjedi_2022052619_letkf ............   Passed   51.20 sec
    Start 3: rrfs_mpasjedi_2022052619_atms_npp_qc
3/4 Test #3: rrfs_mpasjedi_2022052619_atms_npp_qc ......   Passed   90.54 sec
    Start 4: rrfs_mpasjedi_2022052619_atms_npp_qc_bc
4/4 Test #4: rrfs_mpasjedi_2022052619_atms_npp_qc_bc ...   Passed  155.86 sec
100% tests passed, 0 tests failed out of 4
Label Time Summary:
mpi            = 388.37 sec*proc (4 tests)
rdas-bundle    = 388.37 sec*proc (4 tests)
script         = 388.37 sec*proc (4 tests)
Total Test time (real) = 388.72 sec

PR #184, opened 2024-10-02
Update ctests to include more obs types for more realistic testing 

### LINKED ISSUE #178: Expand ctests to include additional obs for more realistic testing
As mentioned in our last meeting, we do not plan to add additional ctests for each new observation type. Instead, we will expand current ctests for each DA method (3DEnVar, GETKF, etc.) to use additional observations as the yamls get added to RDASApp. This will allow us to build up ctests for more realistic scenarios and observation counts. 

My initial thoughts about how to implement this are: 

- Generate new versions of the `rrfs-test/testinput/*yaml` files, combining the DA yamls in `rrfs-test/validated_yamls/templates/basic_config` with observation yamls in `rrfs-test/validated_yamls/templates/obtype_config`. This will make use of the `gen_yaml.sh` tool that @delippi has developed. 
- The older yamls in `rrfs-test/testinput` are still useful for running smaller experiments, so I plan to move those into something like `rrfs-test/expr_input`.

SamuelDegelia-NOAA:
This PR addresses #178 by updating the mpas-jedi ctests in RDASApp to now use a larger set of observations for more realistic testing. The yaml files for these new ctests are generated from the individual yamls added by other developers into `rrfs-test/validated_yamls/templates/obtype_config`. A script is run during build called `gen_yaml_ctest.sh` that concatenates these yamls, merges them with one of the ctest yamls in `rrfs-test/validated_yamls/templates/basic_config`, and then moves those new files into `rrfs-test/testinput` for use in the ctests.

Currently included in the ctests are mesonet, aircraft, aircar, and ATMS obs.

Other notes
- New observations are staged for these tests in `RDAS_DATA/fix/expr_data/mpas_2024052700/obs_ctest`. Thanks to @delippi for providing these new obs. 
- The standalone LETKF and GETKF ctests are removed given that these take a very long time to run. LETKF is also very slow even in split mode (>1 hour) so the only ctest for EnKF that remains is GETKF in split-mode. This test is still slow compared to EnVar (~20 min vs 5 min), but is still much faster than LETKF. **The GETKF efficiency should be further investigated as we continue development.** 
- Old yamls used for experiments are moved into `rrfs-test/testinput_expr` and linked into the `expr` directory so as to not interupt any ongoing testing by other developers

Ctest output for a fresh clone and build on Hera: 

```
(eva) [Samuel.Degelia@hfe05 ush]$ ./run_rrfs_tests.sh wrfruc
+ cd /scratch1/BMC/zrtrr/Samuel.Degelia/RDASApp_dev/RDASApp/ush/../build/rrfs-test
+ pwd
/scratch1/BMC/zrtrr/Samuel.Degelia/RDASApp_dev/RDASApp/build/rrfs-test
+ ctest -j6
Test project /scratch1/BMC/zrtrr/Samuel.Degelia/RDASApp_dev/RDASApp/build/rrfs-test
    Start 4: rrfs_mpasjedi_2024052700_getkf_observer
    Start 1: rrfs_fv3jedi_hyb_2022052619
    Start 2: rrfs_fv3jedi_letkf_2022052619
    Start 3: rrfs_mpasjedi_2024052700_Ens3Dvar
    Start 6: rrfs_mpasjedi_2024052700_bumploc
1/6 Test #2: rrfs_fv3jedi_letkf_2022052619 .............   Passed  548.44 sec
2/6 Test #1: rrfs_fv3jedi_hyb_2022052619 ...............   Passed  662.11 sec
3/6 Test #6: rrfs_mpasjedi_2024052700_bumploc ..........   Passed  954.16 sec
4/6 Test #3: rrfs_mpasjedi_2024052700_Ens3Dvar .........   Passed  1126.73 sec
5/6 Test #4: rrfs_mpasjedi_2024052700_getkf_observer ...   Passed  1322.96 sec
    Start 5: rrfs_mpasjedi_2024052700_getkf_solver
6/6 Test #5: rrfs_mpasjedi_2024052700_getkf_solver .....   Passed  1331.22 sec

100% tests passed, 0 tests failed out of 6

Label Time Summary:
mpi            = 5945.62 sec*proc (6 tests)
rdas-bundle    = 5945.62 sec*proc (6 tests)
script         = 5945.62 sec*proc (6 tests)

Total Test time (real) = 2655.35 sec
+ exit 0

```

Once we have some initial reviews, one of the code managers can add the tags for CI tests on each machine. Also, if this PR is approved, I will update the Wiki to show how to further extend these ctests by adding additional observation types in `rrfs-test/validated_yamls/templates/obtype_config`. It should be pretty easy now - all that needs to be done is to add the yaml into `obtype_config`, update the `gen_yaml_ctest.sh` script, and then update the test reference data.


SamuelDegelia-NOAA:
Also, here some analysis increments for the new EnVar and GETKF ctests: 

temp
qv

SamuelDegelia-NOAA:
As mentioned [here](https://github.com/NOAA-EMC/RDASApp/issues/122#issuecomment-2392563792), GETKF can be significantly sped up using the `reduce obs space` filter action. But we will need to update the OOPS submodule to make use of it.

guoqing-noaa:
The bottom boundaries of the shaded domain in the increment plots are still straight lines. I think @delippi made an update to be able to plot concave lines there?

SamuelDegelia-NOAA:
Here are the same plots but using the updated plotting scripts: 

temp
qv

SamuelDegelia-NOAA:
Also, @guoqing-noaa I should mention that the Hercules test will fail because I have not yet staged the data there due to permissions. Would you mind copying `RDAS_DATA/fix/expr_data/mpas_2024052700/obs_ctest` from either Hera/Jet onto the corresponding location on Orion/Hercules?

guoqing-noaa:
@SamuelDegelia-NOAA  I just sync'ed new files to Orion/Hercules. It's fine if we see a failure message. We can remove the fail label and redo the CI test.

The `role-zrtrr` account is now ready to use on Orion/Hercules. I will send you an email on this soon.

SamuelDegelia-NOAA:
Thanks for syncing the data @guoqing-noaa. BTW I just tested it and role-zrtrr is working for me on Orion now, thanks!

ShunLiu-NOAA (dismissed):

SamuelDegelia-NOAA:
I just pushed a new commit that adds the ATMS yaml from #66 into the ctests. You can click on that latest commit to see an example of how I added the new observation type. The impact to the runtime and analysis increments is very small given how few ATMS obs are available for this test case.

SamuelDegelia-NOAA:
Hmm, I am not able to recreate the Hercules error from @rrfsbot. I just did a fresh clone+build+test of my branch on Orion and everything passed. I did not think that there was any difference in the file system between Orion and Hercules but maybe there was some issue when this test was run.  

Also, the errors for each ctest from @rrfsbot are the same:

`Reason: Input file is not readable, will stop execution. File: data/obs_ctest/atms_npp_obs_2024052700.nc` 

But that file exists (e.g., `/work/noaa/wrfruc/rrfsbot/PRs_RDASApp/184/build/rrfs-test/rundir-rrfs_mpasjedi_2024052700_Ens3Dvar/data/obs_ctest/atms_npp_obs_2024052700.nc`) and I can access (ncdump) it just fine. Could this have just been a file system error?

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for checking the failed CI tests on Hercules. Let's start the tests again.

SamuelDegelia-NOAA:
Looks like it was the same error. I'll try to build on Hercules instead of Orion and see if I can recreate.

guoqing-noaa:
@SamuelDegelia-NOAA You don't need to do a test. I found the reason and will get back to you soon.

guoqing-noaa:
It is because files under `fix/expr_data/mpas_2024052700/obs_ctest` are not globally readable.
`role-wrfruc` does not belong to the 'zrtrr' group and hence cannot access the files created by `role-zrtrr`.

Anyway, this is a good test and demonstrated that it is a good practice to separate the fix file account and the CI test account.

SamuelDegelia-NOAA:
Ah good catch, that explains why I did not have the issue when running the ctests on my account. Thanks @guoqing-noaa!

ShunLiu-NOAA (dismissed):

SamuelDegelia-NOAA:
The PR has been updated to allow yaml files in `validated_yamls` and `testinput` to be parsed by yaml tools as in #186 and #187. Note that the files in `rrfs-test/validated_yamls/templates/obtype_config` still throw an error when using `yamlparser.py` because of the line `seed_time: *analysisDate`. The use of an alias that is undefined is the cause of the error. However, the alias becomes defined when the obspace yamls are merged with the basic config yamls, and the resulting yamls can be parsed.

guoqing-noaa:
Thanks, @SamuelDegelia-NOAA!  
I had expected we may have another PR to do this. But it is great that you took quick actions on this and made changes in this PR! Thanks!

guoqing-noaa:
@SamuelDegelia-NOAA  Sorry that my following comments are NOT on this PR itself. But since we are trying to make YAML files better, I will put them here:

In the combined `rrfs_mpasjedi_2024052700_Ens3Dvar.yaml`, the following anchors are not referenced anywhere:
```
&heightAndHorizObsLoc
&multiIterationFilter
&horizObsLoc
&ObsSpace
&ObsErrorDiagonal
&GetValues
&multiIterationFilter
```
If not used, we would like to remove them.

The `&iterationConfig` anchor is only referenced twice and it contains only a few lines. We may consider putting them back without using anchors/aliases (especially overwrite aliases) to reduce mental workload, :)

guoqing-noaa:
If only generating a yaml file for 131_T (`aircft_airTemperature_131.yaml`), I got this wield thing:
```
name: ""RoundRobin""
```
and
```
name: ""Halo""
```
I checked the `gen_yaml_ctest.sh` file:
```
    # Replace the @OBSFILE@ placeholder with the appropriate observation file (if it hasn't been done already)
    sed -i "s#@OBSFILE@#\"${obs_filename}\"#" ./$temp_yaml
    # Replace the @DISTRIBUTION@ placeholder with the appropriate observation distribution
    sed -i "s#@DISTRIBUTION@#\"${distribution}\"#" ./$temp_yaml
```
I think we don't need to add escaped quotes here as `distribution` and `obs-filename` are simple strings without any special characters, so it is fine to keep as is.

SamuelDegelia-NOAA:
@guoqing-noaa I just removed the stale anchors from the `basic_config` yaml files. I left `&iterationConfig` since it could be nice to have if adding more outer loops. 

For your comment about the double quotes, are you using the latest version of my branch? My [commit from yesterday](https://github.com/NOAA-EMC/RDASApp/pull/184/commits/edde1fc07086a0747e2f2741ec00c6bd00f79a7c) removed those escaped quotes since I noticed the same thing.

guoqing-noaa:
> @guoqing-noaa I just removed the stale anchors from the `basic_config` yaml files. I left `&iterationConfig` since it could be nice to have if adding more outer loops.


Thanks, @SamuelDegelia-NOAA!

For the outer loops, in the past more than 20 years, we have been using two outer loops. I don't expect we will use more than 3 outer loops and the `iterationConfig` only contains a couple of lines. I would suggest we avoid using "overwrite" anchors/aliases which is a mental workload challenge to lots of users.
 
> For your comment about the double quotes, are you using the latest version of my branch? My [commit from yesterday](https://github.com/NOAA-EMC/RDASApp/pull/184/commits/edde1fc07086a0747e2f2741ec00c6bd00f79a7c) removed those escaped quotes since I noticed the same thing.

Ah, thanks! I did not check out your latest changes.

rrfs-test/testinput/rrfs_fv3jedi_letkf_2022052619.yaml
```diff
@@ -32,7 +32,7 @@ background:
         filename_sfcd: 20220526.190000.sfc_data.nc
         filename_sfcw: 20220526.190000.fv_srf_wnd.res.nc
         filename_cplr: 20220526.190000.coupler.res
-      pattern: %mem%
+      pattern: "%mem%"
```
TingLei-NOAA, line 35: Curious, what is the differences using \" or not?

rrfs-test/testinput_expr/namelist.atmosphere
```diff
@@ -0,0 +1,56 @@
+&nhyd_model
```
TingLei-NOAA, line 1: @SamuelDegelia-NOAA  That will be very helpful if a "doc " block could be given on those set ups.  where have they come from ?  If they should be used as a starting point for our rrfs tests/experiments even outside of rdasapp.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
```diff
@@ -62,27 +63,27 @@ cost function:
...
-          nmembers: 20
+          pattern: %mem%
+          nmembers: 20 
           zero padding: 3
         localization:
           localization method: SABER
           saber central block:
             saber block name: BUMP_NICAS
-            active variables: *3dvars
+            active variables: *3dvars
```
TingLei-NOAA: @SamuelDegelia-NOAA  Would you please use this PR to add the below  (shown alongside with the existing "drivers.." block. 
```
         drivers:
         multivariate strategy: duplicated
         compute nicas: true

         model:
                level for 2d variables: last
``` to the fv3jedi var yaml?    As I mentioned in our meetings, this tell the bump to treat the 2d variables as the last level in a 3d structure data, which is needed since the fset is stored from top to bottom when passed to saber.  This option is not needed for this ctest since no 2d analysis variable is used like surface pressure. But adding it here would avoid possible future overlooing on this issue.

File: rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml [hunk 4]
```diff
@@ -99,134 +100,86 @@ cost function:
...
-         - filter: GOMsaver
-           filename: ./msonet_geovals_rrfs_2022052619.nc4
 
 variational:
   minimizer:
-    algorithm: DRPCG    # Derber-Rosati Conjugate Gradients
+    algorithm: DRPCG
   iterations:
-  - ninner: 50
+  - ninner: 5
```
TingLei-NOAA: for real data ,  ninner=5 is too small?

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
```diff
@@ -239,12 +192,12 @@ variational:
...
       layout: [8,10]
       io_layout: [1,1]
-      ntiles: 1
+      ntiles: 1 
       fieldsets:
       - fieldset: DataFix/fix/dynamics_lam_cmaq.yaml
     diagnostics:
       departures: ombg
-  - ninner: 50
+  - ninner: 5
```
TingLei-NOAA: the same question,  5 is a too few for optimization using real data sets.

TingLei-NOAA: In fact "gradient norm reduction: 1e-60" is too strict /ambitious. 
I would recommend to use values like 1e-20 even 1e-10.

rrfs-test/testinput/rrfs_fv3jedi_letkf_2022052619.yaml
SamuelDegelia-NOAA: The quotes are needed because a value cannot start with a `%` if it is to be read by the yaml tools. See #186. Adding the quotes have no impact on the JEDI runs.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
```diff
@@ -257,16 +210,19 @@ variational:
...
+      departures: ombg
 final:
   diagnostics:
     departures: oman
 
 output:
   filetype: fms restart
   datapath: ./
-  prefix: hybens3dvar-fv3_lam-C775
+  prefix: hybens3dvar-fv3_lam-C775 #Ens3dvar-fv3_lam-C775
```
TingLei-NOAA: Though the yaml is with "hyb" in its name, it is actually a pure ens3dvar run,  not sure why the name is "hybens...".  But that is ok for being now. In the future, we could make it a real hyb run if we don't keep a pure ens3dvar ctest.

rrfs-test/testinput_expr/namelist.atmosphere
SamuelDegelia-NOAA: The `namelist.atmosphere` and `streams.atmosphere` here are just copies of what was previously in `rrfs-test/testinput`. I moved that directory into `rrfs-test/testinput_expr` since I believe some people were using those files for other experiments. 

As for where they came from, these files were added in #91 when we added the new MPAS test case. Would @guoqing-noaa, @Junjun-NOAA, or @chunhuazhou be able to respond to this? It might be better to do this in another PR, too.

rrfs-test/testinput_expr/rrfs_mpasjedi_2024052700_Ens3Dvar.yaml
```diff
@@ -123,9 +123,3 @@ cost function:
              name: assign error
              error parameter: 1.2  # K
            defer to post: true
```
TingLei-NOAA, line 125: why are the online verification  removed?

rrfs-test/testinput_expr/rrfs_mpasjedi_2024052700_letkf.yaml
```diff
@@ -85,9 +85,3 @@ output mean prior:
 output: # for outputting mean posterior
   filename: ./ana.$Y-$M-$D_$h.$m.$s.nc
   stream name: analysis 
-
```
TingLei-NOAA, line 88: the same question to those changes,, why online verifications were removed?

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
SamuelDegelia-NOAA: This PR is only updating the yamls to use more data for the MPAS-JEDI tests. These fv3jedi yamls use smaller sets of obs (~800 total obs). I don't mind updating `ninner` to 50 for the fv3jedi case but those files were not created in this PR (only copied between `testinput` and `testinput_expr`).

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
SamuelDegelia-NOAA: See my response from above. This PR is only updating the yamls to use more data for the MPAS-JEDI tests. If you want to change the fv3-jedi yamls then I can do that but those files were not created in this PR (only copied).

rrfs-test/testinput/rrfs_fv3jedi_letkf_2022052619.yaml
guoqing-noaa: `%` is a special character in YAML. Here the `pattern` key expects a `string` type of value. Usually, YAML will not require a string value to be quoted (This is actually a huge drawback of YAML causing many problems and is criticized by many experts).  If a value starts with special characters or has special meaning needing to be escaped, we need to put on the quotes explicitly. Another example, if we want to take a string value of "True" instead of a boolean value, we need to add quotes as well.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
SamuelDegelia-NOAA: Is this fv3jedi yaml not running the hybrid? There is a setting for `covariance model: hybrid` and then there are two covariance sections (`ensemble` and `SABER`).

rrfs-test/testinput_expr/rrfs_mpasjedi_2024052700_Ens3Dvar.yaml
SamuelDegelia-NOAA: The new ctests are run from files in `testinput`. I copied the previous `testinput` directory into `testinput_expr` directory because, I believe, other developers are using those files for experiments. Since the yamls in `testinput_expr` are no longer being used for ctests, the reference verification sections is removed.

rrfs-test/testinput_expr/rrfs_mpasjedi_2024052700_letkf.yaml
SamuelDegelia-NOAA: See my response above.

rrfs-test/testinput_expr/namelist.atmosphere
TingLei-NOAA: @SamuelDegelia-NOAA  So , you meant you were using the set up from GSL 's case,  that is informative..  That will be good if you add those information into this namelist file.   In the future, we  need further discussion on those setup , we know where to start from.

rrfs-test/testinput_expr/rrfs_mpasjedi_2024052700_Ens3Dvar.yaml
SamuelDegelia-NOAA: @TingLei-NOAA I recommend building my branch so that you can see the new yaml files created for the ctests. They are no longer in the repo but are created when RDASApp is built.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
TingLei-NOAA: Did you mean, there are multiple copies of the same files?

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
SamuelDegelia-NOAA: Yes I can add that. Does this need to be added for just fv3-jedi or also mpas-jedi yamls?

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
SamuelDegelia-NOAA: Yes, there are multiple copies of the fv3jedi yamls. They are located in both `rrfs-test/testinput` and `rrfs-test/testinput_expr`. I should probably remove them from `testinput_expr` since that directory was only meant to save the older yamls that other developers might still be using (that were previously used for ctests). The new ctest yamls will be located in `rrfs-test/testinput`.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
TingLei-NOAA: @SamuelDegelia-NOAA   As I asked  previously, did you mean  those files will always be the same ( so, just multiple copied of the same file, or you started from copying the original file, but they are independent to each other?    if they are not supposed to be exactly multiple copies of the same file  , only this file need to be taken care of.

rrfs-test/testinput_expr/rrfs_mpasjedi_2024052700_Ens3Dvar.yaml
TingLei-NOAA: @SamuelDegelia-NOAA  so, those yamls are not for ctests which are generated when rdasapp is built?

SamuelDegelia-NOAA: Correct, the files in `rrfs-test/testinput_expr` are not used for ctests. Those are the yamls that were used for the ctests in the previous version of RDASApp. I did not want to get rid of them since others might still be using them for experiments.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
SamuelDegelia-NOAA: The files in `testinput_expr` started as copies of the older yamls in previous version of `testinput`. But then I removed the test reference section from the copies I made in `testinput_expr` since they are no longer being used for ctests. 

But as I mentioned in another response, I should probably just remove the fv3jedi files from `testinput_expr` since they are the same as in `testinput` (minus the test reference section).

SamuelDegelia-NOAA:
> For the outer loops, in the past more than 20 years, we have been using two outer loops. I don't expect we will use more than 3 outer loops and the `iterationConfig` only contains a couple of lines. I would suggest we avoid using "overwrite" anchors/aliases which is a mental workload challenge to lots of users.

@guoqing-noaa I've seen three outer loops used sometimes but yes it is unlikely to need any more than two. And the `iterationConfig` section is six lines since it contains all of the geometry info needed for each outer loop. But this is pretty minor so I'll just go ahead and remove the anchor.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
TingLei-NOAA: Thanks. I think mpas-jedi var yaml already have  it.

rrfs-test/testinput_expr/rrfs_mpasjedi_2024052700_Ens3Dvar.yaml
TingLei-NOAA: ok. got it. That make sense now.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
TingLei-NOAA: @SamuelDegelia-NOAA  Sorry, I missed that "hybrid " session with the nicas part.  Ok then, that is a hybrid run, while the static B is of unit variances.

SamuelDegelia-NOAA:
Just pushed a new commit that removed the `iterationConfig` anchor/alias. I also removed the copies of the fv3jedi yamls in `rrfs-test/testinput_expr` since those yamls are currently only used for ctests and thus they can stay in `rrfs-test/testinput`. Plus this should make the "Files Changed" section on Github a little less confusing for this PR.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
SamuelDegelia-NOAA: The existing fv3jedi hybrid yaml has `read local nicas: true` instead of `compute nicas: true`. Did you also want to change that? Or just add the `level for 2d variables` under `model:`?

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
TingLei-NOAA: @SamuelDegelia-NOAA  OK.  I will see if there would be another PR to change those setup mentioned above. 
If yamls in testinput_expr is mainly for an copy for possible use/comparison, they could be there as they are .

TingLei-NOAA (dismissed):

SamuelDegelia-NOAA:
Added a [new line](https://github.com/NOAA-EMC/RDASApp/pull/184#discussion_r1793809375) to the fv3jedi yaml per @TingLei-NOAA. It has no impact on the ctest results.

rrfs-test/testinput/rrfs_fv3jedi_hyb_2022052619.yaml
```diff
@@ -90,10 +90,11 @@ cost function:
                 alias:
                 - in code: common
                   in file: fixed_250km_0.3
-
               drivers:
                  multivariate strategy: duplicated
                  read local nicas: true
+              model:
+                 level for 2d variables: last
```
guoqing-noaa, line 97: Just curious, what does this mean?  @TingLei-NOAA @SamuelDegelia-NOAA

SamuelDegelia-NOAA: Per @TingLei-NOAA - This tells bump to treat the 2d variables as the last level in a 3d structure data, which is needed since the fset is stored from top to bottom when passed to saber.  This option is not needed for this ctest since no 2d analysis variable is used like surface pressure. But adding it here would avoid possible future overlooking on this issue.

guoqing-noaa: Thanks, @SamuelDegelia-NOAA 
`fset` is a variable in saber, right?

rrfs-test/testinput_expr/namelist.atmosphere
guoqing-noaa: This file is copied from rrfs-workflow. The MPAS model user guide should have detailed description about each namelist parameters and most of those setting are decided by MODEL or Physics guys.

guoqing-noaa: Here is the v8.0.1 user guide:
https://www2.mmm.ucar.edu/projects/mpas/mpas_atmosphere_users_guide_8.0.1.pdf

They may have a more recent user guide for v8.2.2

rrfs-test/testinput/rrfs_fv3jedi_hyb_2022052619.yaml
SamuelDegelia-NOAA: Yes, to my understanding `fset` defines the fields used by saber.

rrfs-test/testinput_expr/rrfs_fv3jedi_hyb_2022052619.yaml
TingLei-NOAA: Yes as we talked off-line. Just add the "level.." parameter. Thanks

rrfs-test/testinput/rrfs_fv3jedi_hyb_2022052619.yaml
guoqing-noaa: Thanks!

guoqing-noaa:
@SamuelDegelia-NOAA I would recommend we commit those mpasjedi yaml files generated by `gen_yaml_ctest.sh` as discussed in #187 
```
# Ignore the yamls created by gen_yaml_ctest.sh
rrfs-test/testinput/rrfs_mpasjedi_2024052700_Ens3Dvar.yaml
rrfs-test/testinput/rrfs_mpasjedi_2024052700_getkf_observer.yaml
rrfs-test/testinput/rrfs_mpasjedi_2024052700_getkf_solver.yaml
```

SamuelDegelia-NOAA:
@guoqing-noaa I added the new super yamls for the ctests to the repo. Since we are including them directly, there is no longer a need to run `gen_yaml_ctest.sh` during build. But the script will stay for when others want to update the ctest yamls.

rrfs-test/validated_yamls/templates/obtype_config/aircar_specificHumidity_133.yaml
```diff
@@ -1,9 +1,12 @@
      - obs space:
```
guoqing-noaa, line 1: Line 108:
`             is_in: 130,131,135`

This is the `133` file. Does this mean we don't need the filter at that location?

rrfs-test/validated_yamls/templates/obtype_config/aircft_specificHumidity_134.yaml
```diff
@@ -1,9 +1,12 @@
      - obs space:
```
guoqing-noaa, line 1: Check line 108 as well

rrfs-test/validated_yamls/templates/obtype_config/aircar_specificHumidity_133.yaml
SamuelDegelia-NOAA: @delippi Could you comment on this? Is the rejection filter for other obs types needed in the 133 yaml?

rrfs-test/validated_yamls/templates/obtype_config/aircar_airTemperature_133.yaml
```diff
@@ -1,9 +1,12 @@
      - obs space:
          name: aircar
+         distribution:
```
guoqing-noaa, line 3: line 1-2: 
```
- obs space:
   name: aircar
```

Is `aircar` a keyword in JEDI? Does it affect JEDI functionalities if I change it to another name, such as `aircar_T_133`? 

@SamuelDegelia-NOAA @delippi @HuiLiu-NOAA @spanNOAA

spanNOAA: No, you can choose any name you like. From what I've learned, the name simply helps you quickly locate the related data process in the log file.

guoqing-noaa: > No, you can choose any name you like. From what I've learned, the name simply helps you quickly locate the related data process in the log file.

In this situation, let's change the obs space name to be unique. For this file, it should be `aircar_T_133`. Thanks!
@SamuelDegelia-NOAA

guoqing-noaa (changes requested):
Let's make the `obs_space.name` to be unique and more meaningful.

Such as `aircar_T_133`

rrfs-test/validated_yamls/templates/obtype_config/aircar_specificHumidity_133.yaml
delippi: This is one of a few things I will update when I go back through the aircft/aircar yamls for my "Phase 2" validation. No need to worry about this right now. This isn't needed but also doesn't change anything. I will fix this in the next few weeks. We don't assimilate specifcHumidity from 130, 131, 135 which are aircft not aircar.

guoqing-noaa: @delippi This file only contains 133 obs, why we want to check whether it is 130, 131, 135?  I think we should remove this filter as it does not apply to 133.

rrfs-test/validated_yamls/templates/obtype_config/aircar_airTemperature_133.yaml
delippi: @guoqing-noaa, this is another thing that I will be updating too. In my most recent [PR](https://github.com/NOAA-EMC/RDASApp/pull/188/files) I made this change for the msonet yamls . If is okay, I would like to make this change when I go through my "Phase 2" validation and update these yamls which will likely be completed in the next couple of weeks since I've already completed it for msonet. I don't think @SamuelDegelia-NOAA needs to update any of the yamls--I will handle that. Whatever the resultant "super.yaml" from the gen_yaml.sh is what we ought to use. These yamls are not the final product anyway plus the "super.yaml" will need to be updated many times as we continue to add more observations. I think it will work as is for now.

rrfs-test/validated_yamls/templates/obtype_config/aircar_specificHumidity_133.yaml
guoqing-noaa: I understand it does nothing. But it will add one do-nothing filter in the JEDI runtime and it will confuse users.

rrfs-test/validated_yamls/templates/obtype_config/aircar_airTemperature_133.yaml
guoqing-noaa: @delippi unique `obs_space.name` is important as I will use it to filter different `observers`. This  change will not involve any tech/science issues and I think we better to get this updated as earlier as possible.

delippi: @guoqing-noaa, I understand it is helpful for use via your yaml parsing tools, but I don't think this is the PR for that. Sam is just working on getting the ctests working. Please make these comments on my PRs for the yamls. Why do we need to use the yaml-tools to parse this anyway? We can just look at the modular yamls that this is built from in `rrfs-test/validated_yamls/templates/obtype_config/`.

rrfs-test/validated_yamls/templates/obtype_config/aircar_specificHumidity_133.yaml
delippi: @guoqing-noaa, I don't think @SamuelDegelia-NOAA's PR is the place to make this change--he is just trying to update the ctest case which should just use the yamls "as is". The change should be made when I update the yamls which won't be that long. It isn't confusing, it is just unnecessary. I should've removed it before committing this file to start. These are comments that should be made when I submit a PR for the yamls.

guoqing-noaa: > @guoqing-noaa, I don't think @SamuelDegelia-NOAA's PR is the place to make this change--he is just trying to update the ctest case which should just use the yamls "as is". The change should be made when I update the yamls which won't be that long. It isn't confusing, it is just unnecessary. I should've removed it before committing this file to start. These are comments that should be made when I submit a PR for the yamls.

Do you mean only you can make changes to the yaml files under `validated_yaml`?

delippi: I'm not saying I'm the ONLY one who can make changes to the yamls in `validated_yamls`. I just mean this seems out of scope of this PR and Sam shouldn't have to update those yamls. I was planning to make this and other updates when I do my Phase 2 testing on the aircft/aircar yamls which I will start tomorrow and probably finish by end of next week. I think it can wait. Does it cause any problems with the actual ctests?

guoqing-noaa:
@SamuelDegelia-NOAA I created a PR to your branch to update `obs_space.name` to make it more descriptive and distinct among different observers. This is a small change and does not make any technical or scientific differences but it help users to locate different observers even in the "vi" mode. 

For other comments, we can wait for Donnie's new PR #188.  Thanks!

hu5970 (approved):

ShunLiu-NOAA:
Hi, All
Let me know if you still have any concern about this PR. Otherwise, let's merge this PR and open another issue/PR to discuss the obs_space.name.

guoqing-noaa:
@ShunLiu-NOAA we better do CI tests after the new changes. I don't think there will be any issues but it is still good to retest after new changes. I will add "test_*" tags now and it should be completed in an hour.

guoqing-noaa (approved):
Thanks @SamuelDegelia-NOAA for addressing some comments.

ShunLiu-NOAA (approved):

rrfs-test/CMakeLists.txt
guoqing-noaa, line 1: @SamuelDegelia-NOAA line 54:
`file(COPY ${src_yaml}/${case}.yaml DESTINATION ${casedir} )`
and other copying lines.

These copying/linking only runs at the build time.

This is sometime inconvenient in the develop/debug process as we have to track down to find what files to copy/link and then manually do it for new changes (to avoid unnecessary long re-building process).  It will be good if we have a common script which can be used by the build process and at the same time we can also run it outside of the build process to facilitate debugging/developing. 

I have figured this out for my rrfs-test eror and manually copied/linked new files. So it is not a high priority at the moment but it is good to have this capability. Thanks!

SamuelDegelia-NOAA: For my ctest developments, I normally just run `build.sh` but add an exit command after the `cmake` command. But yeah adding a short shell script similar to `rrfs-test/CMakeLists.txt` could be a little easier. I'll do that at some point.

guoqing-noaa: That will be great! Thanks, @SamuelDegelia-NOAA!

PR #183, opened 2024-09-26
Update the bufr2ioda yamls

delippi:
This PR updates the bufr2ioda yaml configurations to address issues identified in the original files. Some of those issues include:

~~change height data type from int to float~~
- change unit and/or longName to more accurately reflect the data
- separate multiple subset types (e.g., aircft, aircar, adpsfc, sfcshp, etc.) and create single ioda file for each subset type.
- separate ObsError/wind to ObsError/windEastward and ObsError/windNorthward
- using stationPressure instead of pressure for when referring to surface pressure
- clean up extra white space
- re-organize to make comparisons easier

PR #188, opened 2024-10-10
Phase 2 validation: updates for msonet yamls

delippi:
## Description
Finished phase 2 testing with great results. The main additions were updates gross error check and better use of QualityMarker. For winds, I added all the providers and specific stations from each provider similarly to the GSI mesonetuselist and mesonet_stnuselist.

## Issue(s) addressed
Results are documented in Issue https://github.com/NOAA-EMC/RDASApp/issues/80

## Dependencies
List the other PRs that this PR is dependent on:
- https://github.com/JCSDA-internal/ufo/pull/3512
- https://github.com/JCSDA-internal/ufo/pull/3526
- https://github.com/NOAA-EMC/RDASApp/pull/227

## Sample Result
Here is the combined test for all msonet data.
fv3jedi_vs_gsi_increment_airTemperature_msonet_all


rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
```diff
@@ -1,5 +1,5 @@
      - obs space:
-         name: msonet
+         name: msonet_airTemperature_188
```
guoqing-noaa, line 2: I think this is too verbose. `msonet_t_188` will be enough to be descriptive and consistent with `msonet_uv_188`

rrfs-test/validated_yamls/templates/obtype_config/msonet_specificHumidity_188.yaml
```diff
@@ -1,5 +1,5 @@
      - obs space:
-         name: msonet
+         name: msonet_specificHumidity_188
```
guoqing-noaa, line 2: similarly, suggest using `msonet_q_188`

rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
```diff
@@ -1,5 +1,5 @@
      - obs space:
-         name: msonet
+         name: msonet_stationPressure_188
```
guoqing-noaa, line 2: similarly, suggest using `msonet_ps_188`

rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
delippi: @guoqing-noaa Thanks for the comment. I was actually thinking about this a little bit when you suggested it in Sam's ctest PR. My thinking was to keep it consistent with the naming convention of the yaml file itself. I don't see why not? IMO it seems a little cleaner. And for the yaml files, I was trying to use the naming convention of the variable names used in JEDI. Where I was having more trouble was actually with my naming convention of `msonet_uv_288` which I kind of do not like. I would personally like to change this to be more consistent with the others, but this contains both windEastward and windNorthward obtypes. I don't think they can be split into two obspaces, nor would that be a better idea... although maybe... Anyway, I didn't want it to be `msonet_windEastward_windNorthward_288.yaml` because that seems too long and it seems you would agree. What do you think about these options?

1. `msonet_winds_288.yaml`
2. `msonet_windComponents_288.yaml`
3. `msonet_windEastNorth_288.yaml`
4. `msonet_windEastNorthward_288.yaml`

How firm are you on using simplified variable names like T, q, uv, ps? I just like keeping it consistent with the JEDI variable naming conventions.

guoqing-noaa: I would think `t, q, uv, ps` is descriptive enough. GSI used it for more than 20 years and I did not hear any problems or complains with that. I don't like JEDI's style to make many things cumbersome and more difficult. With that said, any place which is mandated by JEDI (such as keywords), we have to follow that. But it is lucky that `obs_space.name` is not that case and we may use traditional good names. BTW, `uv` is not bad at all.

delippi: Thanks for the feedback. I see your point about the simplicity of using shorter names like t, q, and uv, especially since they've worked well in GSI for a long time. However, as we're moving away from GSI and fully transitioning to JEDI, I think it makes sense to stick with JEDI's more descriptive naming conventions for consistency across the project.

I agree that people will understand abbreviations like t for airTemperature, but using the full names across the board keeps everything aligned with the JEDI framework and can make things like searching through files easier. For example, searching for airTemperature is more straightforward than searching for t. While this specific instance might not be an issue since the whole string would be msonet_t_188, adopting msonet_airTemperature_188 makes everything more consistent.

Additionally, the yaml file itself is named msonet_airTemperature_188, so I'd really like to keep the naming inside the file consistent with the file name for clarity. What do you think?

ShunLiu-NOAA:
@guoqing-noaa and @delippi Thank you for the discussion. I thought we could use a simpler naming convention. However, as Donnie mentioned we may have to follow JEDI's style at moment. I was told that JEDI is using the naming convention that can be expanded to include model variables beyond atmospheric model.

rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
guoqing-noaa: If succinct and simple is far enough, please don't make it lengthy and cumbersome. 
For example, we use `RDASApp`, not `RapidRefreshForecastSystemDataAssimilationSystemApplication`; 
we use `ObsSfcPCorrected.h` instead of `ObservationSurfacePressureCorrected.h`
we use `atms_npp` instead of `AdvancedTechnologyMicrowaveSounder_NationalPolarorbitingPartnership`

guoqing-noaa:
> @guoqing-noaa and @delippi Thank you for the discussion. I thought we could use a simpler naming convention. However, as Donnie mentioned we may have to follow JEDI's style at moment. I was told that JEDI is using the naming convention that can be expanded to include model variables beyond atmospheric model.

@ShunLiu-NOAA Thanks for input. That's a different story where JEDI tries to follow the CCPP naming convention for model variables (i.e. keywords). Here `obs_space.name` is just a mark for us to distinguish different obs_space in the merged yaml file and actually the most important part is the prepbufr type (e.g. 181, 120, 133, etc). If we can tell what the obs space is at a glance, that's far enough.

delippi:
Aside from the discussion above, I found some other things that I missed before. Changed PR to draft mode.

hu5970:
H All, I think we better to follow general JEDI naming convention. It is long but it is clear for all of the users.

hu5970 (dismissed):

delippi:
> 1. `msonet_wind_288.yaml`
> 2. `msonet_windComponents_288.yaml`
> 3. `msonet_windEastNorth_288.yaml`
> 4. `msonet_windEastNorthward_288.yaml`

@hu5970 @ShunLiu-NOAA @guoqing-noaa Thank you all for the input. If we do try to follow the JEDI naming convention, what is the preferred name to replace `uv` in `msonet_uv_288.yaml`? The actual names are `windEastward` and `windNorthward`, but I thought `msonet_windEastward_windNorthward.yaml` was too long. Maybe that is the preferred?

HuiLiu-NOAA:
I guess msonet_winds.yaml would be good/clear enough as "winds" here in
default in JEDI means windEastward and windNorthward.

HuiLiu-NOAA:
Is there a reason to assimilate only "windEastNorth_288"?

HuiLiu-NOAA:
What I meant is that "msonet_windEastNorth_288" is NOT clear.

delippi:
Thanks @HuiLiu-NOAA. `msonet_winds_288.yaml` sounds good to me if that is what is typically used in JEDI to refer to both. I wasn't sure if that was already established.

> Is there a reason to assimilate only "windEastNorth_288"?
> What I meant is that "msonet_windEastNorth_288" is NOT clear.

I'm not sure that I follow what you mean. I'm not assimilating only wind from 288. I have each obtype from each observing network in its own yaml. So for mesonet, I will have:

```
msonet_airTemperature_188.yaml
msonet_stationPressure_188.yaml
msonet_specificHumidity_188.yaml
msonet_winds_288.yaml
```

Is that is what you mean?

I think this is the easiest approach because it keeps each yaml as simple as possible and facilitates easier testing, debugging, and reusing validated yamls for new obtypes (since a lot of each content can be carried over to other types). I also know that when I'm editing `msonet_uv_288.yaml` I'm only making edits to that obtype. There are no unintended updates to other types. Plus I think it really simplifies the yamls. For validation/testing purposes, it is probably easiest to debug a single obtype from a single platform at a time rather than trying to do it all at once.

guoqing-noaa:
If you really want to follow the JEDI naming convention, please check the latest JEDI naming effort and make sure everything is consistent with them.

For example, check this PR:
https://github.com/JCSDA-internal/oops/issues/2737

HuiLiu-NOAA:
@delippi : the new names look fine to me. 

PS: I guess the yamls for each of the obs types are mainly for current initial testing only. For future operational, it may be better to have a consolidated and complete single yaml file with all of the obs types under the same "obs space"as @guoqing mentioned.

guoqing-noaa:
I think they will make the following change:
```
specific_humidity ->
water_vapor_mixing_ratio_wrt_moist_air OR
water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water```

**Comment by guoqing-noaa** [comment]
For surface temperature observations, please change as follows:
```
surface_temperature -> air_temperature_at_2m
```

**Comment by guoqing-noaa** [comment]
for surface humidity observations, please change as follows:
```
water_vapor_mixing_ratio_wrt_moist_air_at_2m OR
water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water_at_2m
```

**Comment by delippi** [comment]
> @delippi : the new names look fine to me.
> 
> PS: I guess the yamls for each of the obs types are mainly for current initial testing only. For future operational, it may be better to have a consolidated and complete single yaml file with all of the obs types under the same "obs space"as @guoqing mentioned.

@HuiLiu-NOAA, maybe. I'm pretty sure the long term solution is JCB. But even without JCB, I would still argue it is easiest without consolidating them how I currently have them. I still don't understand the reason why we would want to consolidate all these types into a single yaml. It makes them way too complicated and very prone to error.

**Comment by delippi** [comment]
> I think they will make the following change:
> 
> ```
> specific_humidity ->
> water_vapor_mixing_ratio_wrt_moist_air OR
> water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water```
> ```

These aren't exactly what we are talking about. In the JEDI yamls we use the following `specificHumidity`, `airTemperature`, `windEastward`, `windNorthward`, `stationPressure`. The only question is what is used to refer to BOTH `windEastward` and `windNorthward` which @HuiLiu-NOAA mentioned it should just be `winds`.

**Comment by guoqing-noaa** [comment]
I mean if you want to go verbose, we better to use `water_vapor_mixing_ratio_wrt_moist_air_at_2m`, `air_temperature_at_2m`, etc which are more accurate and descriptive, if you want to simplify it, why not just use `q`, `t`? In the places not mandated by JEDI, we really should not bring troubles to ourselves.

File: rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml [hunk 4]
```diff
@@ -0,0 +1,476 @@
...
+           action:
+             name: inflate error
+             inflation variable:
+               name: ObsFunction/ObsErrorFactorPressureCheck
+               options:
+                 variable: windEastward
+                 inflation factor: 4.0
+
+         # error inflation based on pressure check (setupw.f90)
+         - filter: Perform Action
```
spanNOAA: Should we merge the u and v error inflation into one component? Typically, we have u and v measured at the same observation location. If u fails the ObsErrorFactorPressureCheck, then v will also fail, and vice versa.

delippi: @spanNOAA, that is a good question. I will check to see if it can be done. I was thinking those might need to be separated for example only one obtype might be able to be specified at a time. I will need to double check to see if I can combine them.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -0,0 +1,476 @@
...
+           action:
+             name: reject
+
+         # Time window filter
+         - filter: Domain Check
+           apply at iterations: 0,1
+           where:
+             - variable:
+                 name: MetaData/timeOffset
+               minvalue: -0.10
```
spanNOAA: Would it be better to use the exact time here instead of a time offset?

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -0,0 +1,476 @@
...
+         #  - variable: MetaData/latitude
+         #  - variable: MetaData/longitude
+         #  - variable: MetaData/pressure
+         #  - variable: ObsType/windEastward
+         #  - variable: ObsValue/windEastward
+         #  - variable: ObsType/windNorthward
+         #  - variable: ObsValue/windNorthward
+         #  - variable: QCflagsData/windEastward
+         #  - variable: QCflagsData/windNorthward
+
```
spanNOAA, line 545: Why don’t we have a quality marker check specifically for wind observations? Is it because we already have an extensive acceptance list for specific providers and stations, making the quality marker check unnecessary? I apologize for asking so many questions; I'm working on similar YAML files for 187/287 based on yours, and I need to understand the reasoning behind your settings.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
delippi: IMO, no. But we can hear from anyone else or if you have a specific reason for doing it that way. My thinking is this is more generic and wouldn't require changing those exact dates every time you create a new yaml like when you are cycling DA. Keeping it like this would eliminate that extra step.

spanNOAA: Makes sense to me.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
spanNOAA: And I have a similar question regarding the gross error check.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
delippi: Not too many questions. And definitely don't worry about it. I'm thinking, like you said, the mesonet use list is quite extensive. Perhaps non-mesonet winds will need to be handled differently. Perhaps, I should investigate further for the mesonet winds as well.

rrfs-test/validated_yamls/templates/obtype_config/msonet_uv_288.yaml
```diff
@@ -1,210 +0,0 @@
-     - obs space:
```
guoqing-noaa, line 1: It looks like you committed the removal first and then add the renamed file? We expect to see the differences between the old and new file here but it is not.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -0,0 +1,476 @@
...
..."UNVPL","MARNO","CLMSN","RALGH","MNMNV","CHRDN","FRDBG","HMDEN","MRMAR","LAWNC","FRMNJ","DDLEY","WESSN","TACPE","WHNT2","DOYLT","MCHVL","WTRTN","ESTRR","MDLJR","NWTN1","PKTNN","JKKSN","NWIVN","PRRFR","MNNDR","BFMMA","OCCTY","OLYWB","ANDRW","CMBDC","KETUM","BNNYL","WLLWP","MSHIM","DIMCK","HFFMN","ALFRD","FLMDT","A0082","CRNLS","SEEKK","TKKWL","HGHSV","JSPRH","RSFHM","YANKE","MDGRD","WSTGR","PUALP","KNTWA","TACMA","HIGDN","OLDWS","DTRTV","RTTMN","WALLR","HESPE","WTNHT","HNVRT","ASTRA","FRKNN","MDBJH","BSO07","SARVR","ESTHS","CNTPL","PACSC","OKHRB","CAMAS","BRWLL","CBANY","MNTPL","BTHHA","DACSC","OFLCS","ROCHE","OZRDC","THMLM","ATOWN","GGAPL","EVADL","KUNTZ","MINRL","WNSKT","SLTHL","DNDNN","LGNTN","HLLRT","RLGHD","LXNGT","BRESC","RYREV","RBSNC","FWOTH","TCMAW","ALLNN","ATTLL","BOYER","UNVDE","JIMLW","CHNTL","WLCTT","LKWDA","CRMEL","SNYVL","CUMBR","OLEAN","SPGVL","PIRSN","SPNDR","HUTSV","NEWNJ","CLLMN","SHPBG","ETNVL","OTHLL","MOSCW","LXNGS","NTCHZ","GBFHP","NBUOH","ALTMN","MANSF","MDDON","STHLD","CENSQ","CHTTN","SBURY","MCHNC","VNCMB","NRTBY","DFBS1","PHLFF","DLND4","SMSTX","BBRNC","ALLCS","CHATT","HAMBR","FLKVL","GULF2","ELBAC","GBORO","WSTPA","MRRRJ","SLDAD","LYNJS","WRWNY","NORWH"]
```
guoqing-noaa, line 97: My gosh on this

guoqing-noaa: tag @hu5970 to see this.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -0,0 +1,476 @@
...
+         # Gross Error Check
+         - filter: Background Check
+           apply at iterations: 0,1
+           filter variables:
+           - name: windEastward
+           function absolute threshold:
+           - name: ObsFunction/WindsSPDBCheck
+             options:
+               wndtype:    [ 288 ]
+               cgross:     [ 5.0 ]
```
spanNOAA, line 463: I think we also have a gross error check with a threshold * 0.7 when QualityMarker = 3 for wind observations?

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
hu5970: If there is not function or filter to read the mesonet uselist file for now, we can add this as a item for the future development.

hu5970:
Hi All, thanks for great discussions, I can see we have the following results from review:
1) change msonet_uv_288.yaml to msonet_winds_288.yaml to follow the JEDI name conventions as much as possible.
2) Sijie have some comments to make MESONET yaml consistent with METAR yaml that he is working on
3) The list of MESONET uselist inside the Yaml need to be investigated.  We should find or create a function to read in the uselist instead of list it.

Let's me know if I missed any comments. Otherwise, I will approve this PR.

Thanks,
Ming

guoqing-noaa:
Names in [`fv3-jedi/src/fv3jedi/FieldMetadata
/FieldsMetadataDefault.h`](https://github.com/JCSDA/fv3-jedi/blob/d3c800b82eef8d1eaadcbc07c3cf7a96d8425233/src/fv3jedi/FieldMetadata/FieldsMetadataDefault.h):
```
    md.longName = "eastward_wind";
    md.shortName = "ua";
    md.units = "ms-1";
    md.kind = "double";
    md.tracer = "false";
    md.horizontalStaggerLocation = "center";
    md.levels = "full";
    md.space = "magnitude";
    addFieldMetadata(fieldsmetadata, nlev, md);

    md.longName = "northward_wind";
    md.shortName = "va";
    md.units = "ms-1";
    md.kind = "double";
    md.tracer = "false";
    md.horizontalStaggerLocation = "center";
    md.levels = "full";
    md.space = "magnitude";
    addFieldMetadata(fieldsmetadata, nlev, md);

    md.longName = "air_temperature";
    md.shortName = "t";
    md.units = "K";
    md.kind = "double";
    md.tracer = "false";
    md.horizontalStaggerLocation = "center";
    md.levels = "full";
    md.space = "magnitude";
    addFieldMetadata(fieldsmetadata, nlev, md);

    md.longName = "specific_humidity";
    md.shortName = "sphum";
    md.units = "kgkg-1";
    md.kind = "double";
    md.tracer = "true";
    md.horizontalStaggerLocation = "center";
    md.levels = "full";
    md.space = "magnitude";
    addFieldMetadata(fieldsmetadata, nlev, md);

    md.longName = "air_pressure";
    md.shortName = "p";
    md.units = "Pa";
    md.kind = "double";
    md.tracer = "false";
    md.horizontalStaggerLocation = "center";
    md.levels = "full";
    md.space = "magnitude";
    addFieldMetadata(fieldsmetadata, nlev, md);

    md.longName = "surface_pressure";
    md.shortName = "ps";
    md.units = "Pa";
    md.kind = "double";
    md.tracer = "false";
    md.horizontalStaggerLocation = "center";
    md.levels = "1";
    md.space = "magnitude";
    addFieldMetadata(fieldsmetadata, nlev, md);

```

guoqing-noaa:
Names in [`vader/src/vader
/DefaultCookbook.h`](https://github.com/JCSDA/vader/blob/e3457cba4541bc30b6e0fde0b632bbe81a67861b/src/vader/DefaultCookbook.h):

```
        {oops::Variable{"air_temperature"},
                        {AirTemperature_A::Name, AirTemperature_B::Name}},
        {oops::Variable{"specific_humidity"},
                                     {WaterVaporMixingRatioWrtMoistAirAndCondensedWater_A::Name}},
        {oops::Variable{"specific_humidity_at_two_meters_above_surface"},
                                     {WaterVaporMixingRatioWrtMoistAirAndCondensedWater2m_A::Name}},
        {oops::Variable{"uwind_at_10m"},
                                     {uwind_at_10m_A::Name}},
...
        {oops::Variable{"vwind_at_10m"},
                                     {vwind_at_10m_A::Name}}
```

HuiLiu-NOAA:
It looks like "uwind_at_10m" and "vwind_at_10m" are more adequate for
msonet winds.

guoqing-noaa:
I can only find `t`, `air_temperature`, `AirTemperature` in JEDI codes and cannot find `airTemperature`.
Also I can only find `sphum`, `specific_humidity`, `WaterVaporMixingRatioWrtMoistAirAndCondensedWater_A` in JEDI codes and cannot find `specificHumidity`.

@HuiLiu-NOAA Could you help me to find `airTemperature` and `specificHumidity` in JEDI codes? Thanks!

HuiLiu-NOAA:
The complete UFO variable names are defined in:

RDASApp/sorc/ufo/src/ufo/ufo_variables_mod.F90

guoqing-noaa:
@HuiLiu-NOAA  Thanks for pointing out the [`ufo_variables_mod.F90`](https://github.com/JCSDA/ufo/blob/master/src/ufo/ufo_variables_mod.F90) file. 

Here are some relevant lines:
```
character(len=MAXVARLEN), public, parameter :: var_tv   = "virtual_temperature"
character(len=MAXVARLEN), public, parameter :: var_ts   = "air_temperature"
character(len=MAXVARLEN), public, parameter :: var_t    = "temperature"
character(len=MAXVARLEN), public, parameter :: var_q    = "specific_humidity"     ! kg/kg
character(len=MAXVARLEN), public, parameter :: var_u    = "eastward_wind"
character(len=MAXVARLEN), public, parameter :: var_v    = "northward_wind"
character(len=MAXVARLEN), public, parameter :: var_ps     = "surface_pressure"
```

So you can see even JEDI itself use short names in their actual coding work. 

We should not make things complicated than that.     
I would strongly advocate using succinct unique strings for the `obsSpace.name` (or even for the file names if appropriate).   

These obsSpace.name will only be used by ourselves to quick identify different obs spaces. Let's make it simple and enough. (in my opinion, even `133_t`, `135_uv`, `120_q` will be enough).

rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
```diff
@@ -76,35 +76,109 @@
              is_in: 188
            action:
              name: inflate error
-             #inflation factor: 1.09757
              inflation variable:
                name: ObsFunction/ObsErrorFactorPressureCheck
```
spanNOAA, line 99: I couldn't find any description of the ObsErrorFactorPressureCheck function. Could you explain what it does and how to use it? When I tried using it in my experiment, I encountered an assertion failure because the pressures were infinitely large, resulting in NaN values for the log pressures.

delippi: @spanNOAA, yes, I've encountered this several times. What is most likely happening is you have observations outside the domain. You can't use an online domain check here since you're not able to drop those obs from memory and the ObsErrorFactorPressureCheck doesn't know about the QC flags set by the online check. You must use the offline domain check for now. If that doesn't fix your problem let me know. 

As for documentation:
https://github.com/JCSDA-internal/ufo/blob/develop/src/ufo/filters/obsfunctions/ObsErrorFactorPressureCheck.h
https://github.com/JCSDA-internal/ufo/blob/develop/src/ufo/filters/obsfunctions/ObsErrorFactorPressureCheck.cc

It basically mimics some ob error inflation done in GSI if the ob height is within the model levels or below it.

There isn't any such documentation on the JEDI doc webpage, but one of the tricks I use to find these things is to head to the top of my RDASApp build and then
```
cd RDASApp/sorc
grep -ir --include="*.yaml" <search string> .
```
and then you can browse yamls and if you find a function like this one, you can easily find it in the code by
```
find . -name "ObsErrorFactorPressureCheck*"
```

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
delippi: @spanNOAA, I don't think it is possible to combine these, unfortunately. I put both of the filters you asked about below for easy reference. Both filters have `ObsFunction --> options --> variable` so you can only specify one name under `filter variables` and we have to list them separately. The only way I would see to merge them is to eliminate having a separate ObsError for both wind components and just have one. Maybe that is something to think about for the future.
```
         # error inflation based on pressure check (setupw.f90)
         - filter: Perform Action
           filter variables:
           - name: windEastward
           where:
           - variable: ObsType/windEastward
             is_in: 288
           action:
             name: inflate error
             inflation variable:
               name: ObsFunction/ObsErrorFactorPressureCheck
               options:
                 variable: windEastward
                 inflation factor: 4.0

         # Gross Error Check
         - filter: Background Check
           apply at iterations: 0,1
           filter variables:
           - name: windEastward
           function absolute threshold:
           - name: ObsFunction/WindsSPDBCheck
             options:
               wndtype:    [ 288 ]
               cgross:     [ 5.0 ]
               error_min:  [ 1.0 ]
               error_max:  [ 5.0 ]
               variable: windEastward
           action:
             name: reject
```

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
delippi: @spanNOAA, I added some quality marker checks for when it is 3 and it didn't change my result. Specifically, I rejected anything with a QM (quality mark) > 3 and applied the cgross*0.7 when QM==3. It seems like the mesonet use list is enough, but I'm thinking it would be a good idea to those qc checks in the mesonet yaml (even though they don't seem to do anything... at least for this particular instance). What do you think?

rrfs-test/validated_yamls/templates/obtype_config/msonet_uv_288.yaml
delippi: @guoqing-noaa, sorry about that if I did something out of order. I thought I committed my changes to the file with the "uv" name and then committed the name change. I was thinking when you change a file name it doesn't like to show the differences--seems to be the case when I do it. Maybe there is a better move order that I should learn.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
spanNOAA: I didn't notice any reject list applied based on the quality marker, except for surface pressure. Does this mean that for wind observations, the station/provider list includes only those with a quality marker between 0 and 2, while for air temperature and humidity, we also accept observations with a quality marker greater than 3?

HuiLiu-NOAA: @spanNOAA : you can confirm these by looking at the geovals outputs.

guoqing-noaa:
@delippi I think the short name and the long name can coexist for `obsSpace.name` just as the JEDI or NetCDF format practices, for example:
```
- obs space:
  name: t188=msonet_airTemperature_188
```
What do you think?

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -0,0 +1,483 @@
...
+         observed variables: [windEastward, windNorthward]
+         simulated variables: [windEastward, windNorthward]
+
+       obs operator:
+         name: Composite
+         components:
+         - name: VertInterp
+           hofx_scaling: true
+           hofx scaling field: wind_reduction_factor_at_10m
+           hofx scaling field group: GeoVaLs
```
spanNOAA, line 26: Do these hofx_scaling options require an external GSI geovals file, similar to the configurations you commented out below?

delippi: No, JEDI will create the `wind_reduction_factor_at_10m` geovals itself. The comments below I will eventually delete. Those require special code in ufo to use which would override the JEDI created geovals including for f10m. I just wasn't ready to delete the commented out parts yet in case I needed to go back to my Phase 1 validation for something.

delippi:
> @delippi I think the short name and the long name can coexist for `obsSpace.name` just as the JEDI or NetCDF format practices, for example:
> 
> ```
> - obs space:
>   name: t188=msonet_airTemperature_188
> ```
> 
> What do you think?

Hi @guoqing-noaa, are you saying that will allow you to set a short AND long name? I will definitely look into it, but I don't have an answer for my preference yet until I know how it works. Thanks for the suggestion.

delippi:
@spanNOAA, I wanted to let you know I opened an issue in (https://github.com/JCSDA-internal/ufo/issues/3496). I'm not sure if you've had similar issues with ObsErrorFactorPressureCheck with your wind obs (287).

spanNOAA:
> @spanNOAA, I wanted to let you know I opened an issue in (https://github.com/JCSDA-internal/ufo/issues/3496). I'm not sure if you've had similar issues with ObsErrorFactorPressureCheck with your wind obs (287).

@delippi, the offline domain check resolved my issue. Thanks.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
spanNOAA: That's weird. When I try to use wind_reduction_factor_at_10m, I get the error "field wind_reduction_factor_at_10m cannot be obtained from input fields." It seems like this variable is sourced from the background file?

delippi: @spanNOAA, send me the path to your yaml and I will check it out.

spanNOAA: Here's the yaml file: /work/noaa/wrfruc/span/jedi_summer_retro/jedi_prepbufr_analysis/2023061012/3denvar_adpsfc_winds_287.yaml.
And here's the log file: /work/noaa/wrfruc/span/jedi_summer_retro/jedi_prepbufr_analysis/2023061012/fv3jedi_envar_conv.o2848743.

spanNOAA: I'm assuming that "cannot be obtained from input fields" means "cannot be derived from input fields"?

delippi: Hmm that is strange. Yeah, I think it just comes from the input field `f10m`. I see you have that in your sfcdata so I wonder why it isn't working. We'll have to dig into that some more. 

p.s. don't take my current yamls to be a sane starting place, for now. I'm testing some changes related to the order of the obfilters and trying to remove most/all instances of `defer to post`. I'm also struggling some with the ObsErrorFactorPressureCheck (for winds).

spanNOAA: It seems that the issue was resolved by adding f10m to the state variable.

delippi: Okay great! When I was looking at your yaml, `f10m` was in your state variables.

spanNOAA: I do notice that "defer to post" affects the results differently. I'm testing it out as well.

spanNOAA:
@delippi, could you share the horizontal and vertical radius you're using for this JEDI analysis result?

delippi:
> @delippi, could you share the horizontal and vertical radius you're using for this JEDI analysis result?

@spanNOAA, my yaml is using `fixed_250km_0.3`. It looks like I have my GSI localizations set different. Thanks.

guoqing-noaa:
> > @delippi I think the short name and the long name can coexist for `obsSpace.name` just as the JEDI or NetCDF format practices, for example:
> > ```
> > - obs space:
> >   name: t188=msonet_airTemperature_188
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you think?
> 
> Hi @guoqing-noaa, are you saying that will allow you to set a short AND long name? I will definitely look into it, but I don't have an answer for my preference yet until I know how it works. Thanks for the suggestion.

@delippi Thanks for looking further into this.
This will facilitate some quick examination/manipulation of JEDI DA xml files. 
    
For example, If I try to do two DA retro experiment based on Haidao's existing work from a previous RDASApp version. I can take his JEDI DA xml file, check what observations he has assimilated, and then make some adjustment based on my retro needs. 
    
I may use  `my_command haidao.yaml keep=t187,q187,t120,q120,uv120,t133 > my.yaml` to meet my needs.
     
I can also use the following command to achieve the same need:
`
my_command haidao.yaml keep=metar_airTemperature_187,metar_specifiyHumidity_187,sounding_airTemperature_120,sounding_specificHumidity_120, sounding_winds_120, aircar_airTemperature_133
`
    
I understand Different strokes for different folks. In my situation, I would prefer the first way.    

And since `obsSpace.name` only works like "comment", it does not harm any RDASApp functionalities to include both the short names and long names and different folks can pick up what they want. 

If only the long names are used, I may need to add a mapping in the script to achieve the need. But a combination of both short and long formats for obsSpace.name will provide lots of conveniences for needed people. Thanks!

delippi:
> > > @delippi I think the short name and the long name can coexist for `obsSpace.name` just as the JEDI or NetCDF format practices, for example:
> > > ```
> > > - obs space:
> > >   name: t188=msonet_airTemperature_188
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > What do you think?
> > 
> > 
> > Hi @guoqing-noaa, are you saying that will allow you to set a short AND long name? I will definitely look into it, but I don't have an answer for my preference yet until I know how it works. Thanks for the suggestion.
> 
> @delippi Thanks for looking further into this. This will facilitate some quick examination/manipulation of JEDI DA xml files.
> 
> For example, If I try to do two DA retro experiment based on Haidao's existing work from a previous RDASApp version. I can take his JEDI DA xml file, check what observations he has assimilated, and then make some adjustment based on my retro needs.
> 
> I may use `my_command haidao.yaml keep=t187,q187,t120,q120,uv120,t133 > my.yaml` to meet my needs.
> 
> I can also use the following command to achieve the same need: `my_command haidao.yaml keep=metar_airTemperature_187,metar_specifiyHumidity_187,sounding_airTemperature_120,sounding_specificHumidity_120, sounding_winds_120, aircar_airTemperature_133`
> 
> I understand Different strokes for different folks. In my situation, I would prefer the first way.
> 
> And since `obsSpace.name` only works like "comment", it does not harm any RDASApp functionalities to include both the short names and long names and different folks can pick up what they want.
> 
> If only the long names are used, I may need to add a mapping in the script to achieve the need. But a combination of both short and long formats for obsSpace.name will provide lots of conveniences for needed people. Thanks!

@guoqing-noaa, what is "`my_command`"? Is that a placeholder for one of your yaml-tools? What is the actual command?

delippi:
@spanNOAA I still have some things to look into for the bufr2ioda PR #183 before I'm ready to have this merged. There are some things that I will likely need to change such as `mesonetProvider` strings to whatever is the IODA convention. These yamls should at least work as better references for now.

guoqing-noaa:
> > > > @delippi I think the short name and the long name can coexist for `obsSpace.name` just as the JEDI or NetCDF format practices, for example:
> > > > ```
> > > > - obs space:
> > > >   name: t188=msonet_airTemperature_188
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > What do you think?
> > > 
> > > 
> > > Hi @guoqing-noaa, are you saying that will allow you to set a short AND long name? I will definitely look into it, but I don't have an answer for my preference yet until I know how it works. Thanks for the suggestion.
> > 
> > 
> > @delippi Thanks for looking further into this. This will facilitate some quick examination/manipulation of JEDI DA xml files.
> > For example, If I try to do two DA retro experiment based on Haidao's existing work from a previous RDASApp version. I can take his JEDI DA xml file, check what observations he has assimilated, and then make some adjustment based on my retro needs.
> > I may use `my_command haidao.yaml keep=t187,q187,t120,q120,uv120,t133 > my.yaml` to meet my needs.
> > I can also use the following command to achieve the same need: `my_command haidao.yaml keep=metar_airTemperature_187,metar_specifiyHumidity_187,sounding_airTemperature_120,sounding_specificHumidity_120, sounding_winds_120, aircar_airTemperature_133`
> > I understand Different strokes for different folks. In my situation, I would prefer the first way.
> > And since `obsSpace.name` only works like "comment", it does not harm any RDASApp functionalities to include both the short names and long names and different folks can pick up what they want.
> > If only the long names are used, I may need to add a mapping in the script to achieve the need. But a combination of both short and long formats for obsSpace.name will provide lots of conveniences for needed people. Thanks!
> 
> @guoqing-noaa, what is "`my_command`"? Is that a placeholder for one of your yaml-tools? What is the actual command?

It can be any commands. The yaml-tools can be one of the options.

delippi:
> > > > > @delippi I think the short name and the long name can coexist for `obsSpace.name` just as the JEDI or NetCDF format practices, for example:
> > > > > ```
> > > > > - obs space:
> > > > >   name: t188=msonet_airTemperature_188
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > What do you think?
> > > > 
> > > > 
> > > > Hi @guoqing-noaa, are you saying that will allow you to set a short AND long name? I will definitely look into it, but I don't have an answer for my preference yet until I know how it works. Thanks for the suggestion.
> > > 
> > > 
> > > @delippi Thanks for looking further into this. This will facilitate some quick examination/manipulation of JEDI DA xml files.
> > > For example, If I try to do two DA retro experiment based on Haidao's existing work from a previous RDASApp version. I can take his JEDI DA xml file, check what observations he has assimilated, and then make some adjustment based on my retro needs.
> > > I may use `my_command haidao.yaml keep=t187,q187,t120,q120,uv120,t133 > my.yaml` to meet my needs.
> > > I can also use the following command to achieve the same need: `my_command haidao.yaml keep=metar_airTemperature_187,metar_specifiyHumidity_187,sounding_airTemperature_120,sounding_specificHumidity_120, sounding_winds_120, aircar_airTemperature_133`
> > > I understand Different strokes for different folks. In my situation, I would prefer the first way.
> > > And since `obsSpace.name` only works like "comment", it does not harm any RDASApp functionalities to include both the short names and long names and different folks can pick up what they want.
> > > If only the long names are used, I may need to add a mapping in the script to achieve the need. But a combination of both short and long formats for obsSpace.name will provide lots of conveniences for needed people. Thanks!
> > 
> > 
> > @guoqing-noaa, what is "`my_command`"? Is that a placeholder for one of your yaml-tools? What is the actual command?
> 
> It can be any commands. The yaml-tools can be one of the options.

@guoqing-noaa, in your particular example `my_command haidao.yaml keep=t187,q187,t120,q120,uv120,t133 > my.yaml`, what would be the command you are using? I wanted to try it myself to better understand what you are doing and how it works.

guoqing-noaa:
> @guoqing-noaa, in your particular example `my_command haidao.yaml keep=t187,q187,t120,q120,uv120,t133 > my.yaml`, what would be the command you are using? I wanted to try it myself to better understand what you are doing and how it works.

In my situation, that command is still in the development and not completed yet. 
I am trying to give an example how the short names and/or long names will be used.

Right now, we haven't have unique obsSpace.name for individual RDASApp yaml files yet. When we have the unique short/long names ready, I can give it a try. Thanks!

_Clarify: it will NOT be named as `my_command` and will have a meaningful name when it is ready_

delippi:
> > @guoqing-noaa, in your particular example `my_command haidao.yaml keep=t187,q187,t120,q120,uv120,t133 > my.yaml`, what would be the command you are using? I wanted to try it myself to better understand what you are doing and how it works.
> 
> In my situation, that command is still in the development and not completed yet. I am trying to give an example how the short names and/or long names will be used.
> 
> Right now, we haven't have unique obsSpace.name for individual RDASApp yaml files yet. When we have the unique short/long names ready, I can give it a try. Thanks!
> 
> _Clarify: it will NOT be named as `my_command` and will have a meaningful name when it is ready_

Can you give it a try with this yaml:
`/scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2/jedi-assim-aircft/rrfs-data_fv3jedi_2022052619/msonet.yaml`? There are 4 uniquely named obs spaces in that file (sorry it is still the long name at the moment), but you could modify it for testing purposes. Let me know if/when you have such command ready. I would like to see what you are envisioning here.

delippi:
I think this PR is ready to go. I didn't add any of the short/long name stuff just because I wanted to see how that was exactly going to be used for first. It can always be something added later. The results from with these yamls are in the following https://github.com/NOAA-EMC/RDASApp/issues/80#issuecomment-2430051112

delippi:
> I think this PR is ready to go. I didn't add any of the short/long name stuff just because I wanted to see how that was exactly going to be used for first. It can always be something added later. The results from with these yamls are in the following [#80 (comment)](https://github.com/NOAA-EMC/RDASApp/issues/80#issuecomment-2430051112)

Actually, I need to run all these tests again for MPAS and make sure the yamls work for that as well.

delippi:
> > I think this PR is ready to go. I didn't add any of the short/long name stuff just because I wanted to see how that was exactly going to be used for first. It can always be something added later. The results from with these yamls are in the following [#80 (comment)](https://github.com/NOAA-EMC/RDASApp/issues/80#issuecomment-2430051112)
> 
> Actually, I need to run all these tests again for MPAS and make sure the yamls work for that as well.

Okay, I've rerun these tests for MPAS-JEDI using these yamls. I fixed a couple of things... one of the things was stationPressure was using `surface_geometric_height` where that isn't available for MPAS-JEDI. The only yaml that fails is still specificHumidity which was expected because for some reason MPAS-JEDI can't compute `saturation_specific_humidity` in the ObsErrorFactorPressureCheck which is required for scaling the RH oberrors. I need to investigate that more. So all the yamls work except the specificHumidity (the ObsErrorFactoPressureCheck could just be turned off in MPAS-JEDI tests until I figure that one out). The only result I wasn't happy with was stationPressure increments in MPAS-JEDI. I will dig more into that with "Phase 3" testing which will compare the FV3-JEDI result against MPAS-JEDI. We don't yet have that matching case. It functionally works though.

rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
```diff
@@ -52,29 +71,15 @@
...
-         #        pressure: MetaData/pressure
-         #        distance threshold: -1.
-         #  defer to post: true
+             error function:
+               name: ObsFunction/ObsErrorModelStepwiseLinear
+               options:
+                 xvar:
+                   name: MetaData/pressure
+                 xvals: [110000, 105000, 100000, 95000, 90000, 85000, 80000, 75000, 70000, 65000, 60000, 55000, 50000, 45000, 40000, 35000, 30000, 25000, 20000, 15000, 10000, 7500, 5000, 4000, 3000, 2000, 1000, 500, 400, 300, 200, 100, 0]
+                 errors: [4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0]
```
SamuelDegelia-NOAA, line 87: I might have missed this in previous discussion, but how is this different from the previous version that sets a constant error of 4 K to all obs?

delippi: It isn't. I was just doing that for consistency between all yamls. So side-by-side comparisons or when I create a new yaml based on the msonet obs there will be less to change. It's effectively the same, but just a matter of my own preference for managing the yamls.

SamuelDegelia-NOAA: Okay, makes sense. I guess this is similar to how things are defined in the GSI `errtable` too with a profile set for each obtype including mesonet

rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
```diff
@@ -83,61 +88,98 @@
...
-         - filter: Domain Check
-           apply at iterations: 0,1
            where:
-             - variable:
-                 name: MetaData/timeOffset
-               minvalue: -0.10
-               maxvalue: 0.10
+           - variable: ObsType/airTemperature
+           - variable: QualityMarker/airTemperature
+             is_not_in: 3
```
SamuelDegelia-NOAA, line 170: Also curious why do we not use the gross error checks for QualityMarker=3?

rrfs-test/validated_yamls/templates/obtype_config/msonet_specificHumidity_188.yaml
```diff
@@ -40,10 +40,29 @@
...
+         # Time window filter
+         - filter: Domain Check
+           apply at iterations: 0,1
+           where:
+             - variable:
+                 name: MetaData/timeOffset # units: s
+               minvalue: -360
+               maxvalue:  360
+           action:
+             name: reduce obs space
```
SamuelDegelia-NOAA: As discussed today, could we remove this until #209 is merged? Using `reduce obs space` breaks GETKF until OOPS is updated.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -0,0 +1,513 @@
...
+               error_max:  *error_max
+               variable: windNorthward
+           where:
+           - variable: QualityMarker/windNorthward
+             is_in: 3
+           action:
+             name: reject
+
+         - filter: GOMsaver
+           filename: ./data/geovals/msonet_geovals_rrfs.nc4
```
SamuelDegelia-NOAA: Could we comment this out for now like with the other ob types? We do not have write access to `data/` in RDASApp currently.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -0,0 +1,513 @@
...
+               wndtype:    [ 288 ]
+               cgross:     &cgross    [ 5.0 ]
+               error_min:  &error_min [ 1.0 ]
+               error_max:  &error_max [ 5.0 ]
+               variable: windEastward
+           where:
+           - variable: QualityMarker/windEastward
+             is_not_in: 3
+           action:
+             name: reject
```
SamuelDegelia-NOAA: It looks like this function is setting minimum and maximum errors within the gross error check - am I interpreting that right? Could we use something similar for t/q/p embedded within the gross error checks? Instead of using the individual `Perform Action` filters for min/max error

rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
delippi: You might have overlooked it but there is a separate gross error check for QualityMarker=3. That one uses cgross*0.7. So for this obtype cgross=5 and cgross*0.7=3.5

rrfs-test/validated_yamls/templates/obtype_config/msonet_specificHumidity_188.yaml
delippi: Yes, I can remove it in this PR. I think you'll still probably have conflicts when you add it back in because I reordered a bunch of the filters.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
delippi: Yep! I meant to comment that back out and just forgot.

rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
SamuelDegelia-NOAA: Ahh yes, I did miss that. I was just looking at the track changes view on GitHub so it didn't show the other filter. Thanks!

rrfs-test/validated_yamls/templates/obtype_config/msonet_specificHumidity_188.yaml
SamuelDegelia-NOAA: Yeah - I'll probably have a bit to work through with my PR after this one is merged hah.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
delippi: Yes, you are interpreting that correct. In the gross error check there is supposed to be a min/max error used in that check. I would prefer it already built into that filter, but it is not. I don't know why.

SamuelDegelia-NOAA: Is it possible to do the same for t/q/p? So that we wouldn't need those extra filters (e.g., L116-144 in `msonet_airTemperature_188.yaml`).

delippi: Oh sorry, for some reason I was thinking something else. Please disregard my previous statement... Let me clarify.

All of the gross error checks need to have an error_min error_max in the gross error check. You can do this in the t/q/ps gross checks which I do in the following steps (which I've noted in the comments).

1. Create temporary ObsErrorData
2. Set ObsError set "error parameter" if < "max value"
3. Set ObsError set "error parameter" if > "min value"
4. Gross Error Check
5. Re-assign err ObsErrorData <--- TempObsErrorData after gross error check

delippi: What I was referring to in my initial statement is that I would like the gross error checks to already have the error min/max built in, but it doesn't so I have to do it in this round about way. The error check for wind does have it built in. I hope that was clear.

SamuelDegelia-NOAA: I understand now, those filter actions are needed for t/q/p to define the min and max errors used during the gross error check. Your explanation was helpful. Hopefully they can be set in the non-wind gross error checks sometime in the future.

SamuelDegelia-NOAA (dismissed):
LGTM, thanks for answering all of my questions @delippi!

delippi:
@SamuelDegelia-NOAA thanks for the quick review and great questions!

HuiLiu-NOAA:
@delippi : better to add a "maxvalue: 0.5" to the online regional domain check since the MPAS domain mask index can be up to 1-7 in the boundary zones.

delippi:
> @delippi : better to add a "maxvalue: 0.5" to the online regional domain check since the MPAS domain mask index can be up to 1-7 in the boundary zones.

@HuiLiu-NOAA, I think I do have that. Here's an example of how I have it used:
```
         # Online domain check
         - filter: Bounds Check
           filter variables:
           - name: windEastward
           - name: windNorthward
           test variables:
           - name:  GeoVaLs/observable_domain_mask
           minvalue: 0.0
           maxvalue: 0.5
```

rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
```diff
@@ -86,56 +102,99 @@
...
          - filter: Background Check
            apply at iterations: 0,1
            filter variables:
            - name: stationPressure
-           absolute threshold: 5.0
-           threshold: 5.0
-           where:
-           - variable: ObsType/stationPressure
-             is_in: 188
+           threshold: 50.0
```
spanNOAA: I think the threshold should be 5 here, and 3.5 below?

delippi: @spanNOAA, good question. I actually really wasn't too sure about this. Let me just show some results to start some discussion. I think we need to look into it some more. Neither result looks particularly great, but the magnitudes of the increments looked a little closer at x10.

**Result with threshold at x1 (5 & 3.5):** 
fv3jedi_vs_gsi_increment_airTemperature_msonet_stationPressure_188

**Result with threshold at x10 (50 & 35):** 
fv3jedi_vs_gsi_increment_airTemperature_msonet_stationPressure_188

SamuelDegelia-NOAA (dismissed):

rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
spanNOAA: Hmm, this is interesting. What’s behind this difference? Are there more observations passing the gross check and entering the minimization process? Have you looked into the observations assimilated by both systems in this case?

delippi: The only difference is the cgross thresholds for the gross error checks. Increasing the cgross threshold values would be less restrictive and allow more obs through the gross check. I haven't looked at the observations themselves, but I have ob counts (nobs) in each figure.

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -0,0 +1,521 @@
...
+         - filter: Background Check
+           filter variables:
+           - name: windNorthward
+           function absolute threshold:
+           - name: ObsFunction/WindsSPDBCheck
+             options:
+               wndtype:    [ 288 ]
+               cgross:     *cgross
+               error_min:  *error_min
+               error_max:  *error_max
```
spanNOAA: I'm not pretty sure if we should use a reference here, as well as for temperature, humidity, and surface pressure. It might be fine when using just one type of observations, but does the reference still hold when we combine multiple observations? Even if it does work, I think these references could lead to confusion for users since each may represent different values for different observations.

spanNOAA: If we want to use a reference here, it might be better to choose a name like &error_max_wind_288?

delippi: Yeah, I think you might be right. I hadn't tried it with another wind obs space. I wasn't defining these in the ps/t/q msonet obs space, so I didn't catch the issue. I was only defining them in the uv obs space because it has both eastward and westward winds so 4 total gross checks I think I'm now leaning toward just removing the references. What is your preference @spanNOAA?

spanNOAA: I was using the exact numbers. I'm preparing to align with your standards and submit a draft PR, and then this question came to mind.

delippi: @spanNOAA, I just removed the anchors and aliases.

rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml
```diff
@@ -83,61 +95,98 @@
...
+                 variables:
+                 - name: ObsErrorData/airTemperature
+
+         # Set ObsError set "error parameter" if < "max value"
+         - filter: Perform Action
+           filter variables:
+           - name: airTemperature
+           action:
+             name: assign error
+             error parameter: &error_min 1.0
```
spanNOAA: The same concern applies to the error_min and error_max for T, Q, and PSFC observations.

delippi: @spanNOAA, thanks for reminding me that I had that in the other files as well. I just removed and explicitly set max min error.

rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
delippi: @spanNOAA, I'm adding print statements in GSI:setupps.f90 and UFO:BackgroundCheck.cc to look into this more with a single ob test. 

There are these lines in setupps.f90 which draw my attention
```
obserror = min(r10/max(ratio_errors*error,tiny_r_kind),huge_r_kind)
residual = abs(r10*ddiff)
```

My print in GSI:
```
281:    obserror = 0.541315926311159     
281:    obserrlm = 1.00000000000000     
281:    residual = 3.47478724557078     
281:    ratio = 3.47478724557078     
281:    ddiff = -0.347478724557078     
281:    error = 18.5563178156275     
```

My prints in UFO:
```
 0:   errorMultiplier[jobs] = 54.1881
 0:   hofx[jobs] - obs[jv][jobs] = 334.414
```

In GSI ratio > cgross is compared. In JEDI hofx[jobs] - obs[jv][jobs] > thr*errorMultiplier[jobs] is compared. where I have thr (threshold) set to x10 GSI's value of 5. Here is how I see the calculation going in each system:

In GSI:
```f90
obserror=0.5413
obserrlm=1
residual=10*0.3474=3.474
ratio=3.474/1

#Then check if ratio > cgross
3.474 > 5
```

In UFO:
```f90
thr=50 #I got 50 by taking cgross in convinfo and x10
errorMultiplier=54.18
hofx - obs = 334.414

#Then check if hofx - obs - bias) > thr * errorMultiplier
334.414/54.18 > 50
```

I'm thinking a couple of things. The thing I am leaning closer toward is that the errorMultiplier in UFO just needs to use the 100 Pa min error that I was trying to set in the yaml. That would then make the calculation such that I can compare against 5 instead of 50.

spanNOAA: It seems that errorMultiplier depends on thresholdWrtBGerror to determine whether to use hofxerr or obserr_[iv]. What I’m wondering is: what exactly is thresholdWrtBGerror, and how is it configured? In GSI, the ratio is calculated as abs(pob - pges) in hPa divided by obserrlm, which is set to errmin since obserr is smaller than errmin (which is 1). It appears that errorMultiplier is intended to function similarly to obserrlm in GSI, but its value seems to be equal to obserr in JEDI.

delippi: hofxerr is only used if you specify it in the yaml. We are not. You have to specify it as:
```
threshold wrt background error: true
```
So you can use the background error instead of the obserror.

I'm thinking that the min ob error of 100 Pa isn't getting into the Background Check. When I remove the "Create temporary ObsErrorData" part, it does use the 100 Pa. The only thing I can think of is the Background Check tries using that TempObsErrorData/stationPressure values instead of ObsErrorData/stationPressure. My understanding is that it should use ObsErrorData unless otherwise specified.

delippi: It seems like adding `defer to post: true` after the Re-assign err ObsErrorData might work. I need to test more.

delippi: @spanNOAA, after trying a few things this was the best I could get it to be using the same thresholds as GSI. There could very well be something I'm missing too. The ob counts actually look a lot closer, but the spatial structure and magnitude of the increments still looks a little off. Perhaps it is slightly closer to GSI now than in the other two options I have shared in this thread? Other than now having the thresholds the same as GSI, I just added `defer to post: true` after each step starting with creating the TempObsErrorData, setting max/min oberrors for gross checks, the gross checks themselves, and re-assigning the TempObsErrorData back into ObsErrorData. I also put some debug print statements in the code to make sure that the min/max error values to be used in the gross check is being used. I was finding that without the `defer to post: true` it was just using the TempObsErrorData values in the gross error check--hence the reason I needed to have a larger cgross value.

 
fv3jedi_vs_gsi_increment_airTemperature_msonet_stationPressure_188

ShunLiu-NOAA:
@delippi Is there anything else that needs to be changed with the PR? Or are you waiting for the reviewers' comments?

delippi:
@ShunLiu-NOAA, I think we should wait until I finalize some other PRs I have in ufo. There will be some updates needed to these yamls to invoke some of the options that I am implementing. Here are those ufo PRs I'm currently working on.

https://github.com/JCSDA-internal/ufo/pull/3512
https://github.com/JCSDA-internal/ufo/pull/3526

delippi:
@ShunLiu-NOAA, I think https://github.com/JCSDA-internal/ufo/pull/3512 is the only one we needed to wait for. I had approval for it yesterday but had to change some things with the coding norms. The naming of the new options are finalized now, so I think I'm done with everything here. The only hold up now will be if we want to wait for both of those UFO PRs to be merged and then sync the submodules in RDASApp first? The yamls will "probably" work functionally until then. I don't think it will be too much longer on those PRs.

ShunLiu-NOAA:
@delippi Thanks for the update. Let's wait for UFO PRs to be merged and then sync submodules in RDASApp.

delippi:
@ShunLiu-NOAA, looks like JCSDA fv3-jedi and ufo develop branches have all of my updates. I'll work on updating the submodules with these changes.

delippi:
This comment https://github.com/NOAA-EMC/RDASApp/issues/80#issuecomment-2486650715 shows the most up-to-date results. Overall, I think everything looks good to go. The ob counts are a bit lower for stationPressure. I can change it from `PT06M` to `PT00M` to get more obs, but that seems counter productive to removing duplicate obs. Furthermore, it doesn't change results much just overall ob counts.

spanNOAA (approved):

ShunLiu-NOAA (approved):

SamuelDegelia-NOAA (approved):
LGTM, thanks for all of the effort @delippi. We'll probably continue updating the obs-space sections of the yamls a ton during development. So I think its best to merge this now and do separate PRs for any additional changes

Issue #185, opened 2024-10-03
Python converter for RRFS/ [3D]-RTMA MSONET BUFR


PraveenKumar-NOAA:
Description

Write a Python converter for the MSONET BUFR DUMP that converts a BUFR file into an IODA file for the following subsets:

| NC255001 | A55001 | MTYP 255-001 Mesonet from MADIS: Denver Urban Drainage |
| NC255002 | A55002 | MTYP 255-002 Mesonet from MADIS: RAWS (NIFC) |
| NC255003 | A55003 | MTYP 255-003 Mesonet from MADIS: MesoWest (many subpr) |
| NC255004 | A55004 | MTYP 255-004 Mesonet from MADIS: APRS WeatherNet |
| NC255005 | A55005 | MTYP 255-005 Mesonet from MADIS: Kansas DOT |
| NC255006 | A55006 | MTYP 255-006 Mesonet from MADIS: Florida |
| NC255007 | A55007 | MTYP 255-007 Mesonet from MADIS: Iowa DOT |
| NC255008 | A55008 | MTYP 255-008 Mesonet from MADIS: Minnesota DOT |
| NC255009 | A55009 | MTYP 255-009 Mesonet from MADIS: Anything Weather |
| NC255010 | A55010 | MTYP 255-010 Mesonet from MADIS: NOS PORTS |
| NC255011 | A55011 | MTYP 255-011 Mesonet from MADIS: Aberdeen Proving Grds |
| NC255012 | A55012 | MTYP 255-012 Mesonet from MADIS: Weather for You |
| NC255013 | A55013 | MTYP 255-013 Mesonet from MADIS: NWS Cooperative Obs |
| NC255014 | A55014 | MTYP 255-014 Mesonet from MADIS: NWS HADS |
| NC255015 | A55015 | MTYP 255-015 Mesonet from MADIS: AWS |
| NC255016 | A55016 | MTYP 255-016 Mesonet from MADIS: Iowa Environmental |
| NC255017 | A55017 | MTYP 255-017 Mesonet from MADIS: Oklahoma |
| NC255018 | A55018 | MTYP 255-018 Mesonet from MADIS: Colorado DOT |
| NC255019 | A55019 | MTYP 255-019 Mesonet from MADIS: West Texas |
| NC255020 | A55020 | MTYP 255-020 Mesonet from MADIS: Wisconsin DOT |
[...]

PR #181, opened 2024-09-26
Python converter for Msonet prepBUFR

PraveenKumar-NOAA:
- This PR adds a Python converter for the Msonet prepbufr data. 

- This converter will be used to convert Msonet prepbufr data to IODA (netCDF) format. Following steps similar to GDASApp, a Python file and a Json, description file were created. In its current form, this converter can be run using the ~ush/ioda/bufr2ioda/run_bufr2ioda.py script of the GDASApp. 

- The following two files are new: 
1. ~/IODA/python/bufr2ioda_msonet_prepbufr.py
2.  ~/IODA/python/bufr2ioda_msonet_prepbufr.json

- Related issue #123

PR #192, opened 2024-10-11
Python converter for MSONET BUFR DUMP

PraveenKumar-NOAA:
- This PR adds a Python converter for MSONET BUFR DUMP data.

- The converter will be used to transform MSONET BUFR data into IODA (netCDF) format. Following steps similar to GDASApp, a Python file and a JSON description file were created.

- The following two files are new:

1. ~/IODA/python/bufr2ioda_msonet.py
2. ~/IODA/python/bufr2ioda_msonet.json

- Related issue: https://github.com/NOAA-EMC/RDASApp/issues/185

- Validation for the output observation and IODA variable naming has been performed.


hu5970:
@delippi @SamuelDegelia-NOAA Could you review this PR. This PR should not impact other part of the App. Thanks, Ming

SamuelDegelia-NOAA:
I think a lot of the comments from #181 will probably be the same here. One thing is that we need the `run_bufr2ioda.sh` utility to run the converters added here, but that tool is only included in #181 which still hasn't been merged. I will copy it over for testing but just an FYI if anyone else wants to test this PR too. 

@PraveenKumar-NOAA do you happen to have any mesonet bufr files to test this with?  We do not have a `rap.t00z.msonet.tm00.bufr_d` file staged for any of the cases in RDASApp, only the larger prepbufr files.

PraveenKumar-NOAA:
> I think a lot of the comments from #181 will probably be the same here. One thing is that we need the `run_bufr2ioda.sh` utility to run the converters added here, but that tool is only included in #181 which still hasn't been merged. I will copy it over for testing but just an FYI if anyone else wants to test this PR too.
> 
> @PraveenKumar-NOAA do you happen to have any mesonet bufr files to test this with? We do not have a `rap.t00z.msonet.tm00.bufr_d` file staged for any of the cases in RDASApp, only the larger prepbufr files.

@SamuelDegelia-NOAA most of the comments on #181 are about naming and unit changes. I followed the JEDI standards here and do not agree with making those changes.

I will send you path for a msonet bufr file through chat, please check this PR. Thanks!

SamuelDegelia-NOAA:
> > I think a lot of the comments from #181 will probably be the same here. One thing is that we need the `run_bufr2ioda.sh` utility to run the converters added here, but that tool is only included in #181 which still hasn't been merged. I will copy it over for testing but just an FYI if anyone else wants to test this PR too.
> > @PraveenKumar-NOAA do you happen to have any mesonet bufr files to test this with? We do not have a `rap.t00z.msonet.tm00.bufr_d` file staged for any of the cases in RDASApp, only the larger prepbufr files.
> 
> @SamuelDegelia-NOAA most of the comments on #181 are about naming and unit changes. I followed the JEDI standards here and do not agree with making those changes.
> 
> I will send you path for a msonet bufr file through chat, please check this PR. Thanks!

@PraveenKumar-NOAA No problem! For any of the changes that you agree with in #181, would you mind making the corresponding changes here (if needed)? For the others, we can keep the discussion inside #181.

SamuelDegelia-NOAA:
Thanks @PraveenKumar-NOAA - I was able to run these tools successfully in RDASApp as well after adding copying over the files in #181! 

A couple of quick comments from my first test: 
- I noticed that the IODA files created here are missing the `ObsError` and `ObsType` groups.  We are using `ObsType` for some filtering in JEDI. Do these data exist in the bufr files and would it be possible to include them in the IODA data?  
- The `MetaData/timeOffset` variable is missing - we definitely need this one 
- In the prepbufr python converter, `metaData/pressure` is a float but here it is an int. I am not sure if this is important (@delippi might know)

SamuelDegelia-NOAA:
The `ObsType` and `ObsError` data do not exist in bufr files, so we cannot expect these IODA files to be the same as the prepbufr converters. 

As with #181, I am okay with merging this now to get the framework for the python-based converters into RDASApp. Then we can modify more as we go forward.

SamuelDegelia-NOAA (approved):

delippi:
I agree as well.

One thing I want to mention that doesn't hold anything up is since ObsError doesn't exist in bufr then why do we use it in prepbufr? I've been an advocate for not having ObsError in the IODA and just require users to specify the errors to use.

delippi (approved):

SamuelDegelia-NOAA:
@PraveenKumar-NOAA @delippi Do we happen to know if anyone (e.g., GDAS or another JEDI application) is using the obs error data from prepbufr? Maybe this inclusion of `ObsError` in the bufr2ioda converters is a carry over from another application? I know in GSI that regional configurations were forced to read obs errors from `errtable` and not the prepbufr files. So it makes sense not to include the prepbufr obs errors in RDASApp. 

However, I experienced an issue with running GETKF in split mode if `ObsError` does not exist in the input IODA files. An empty `DerivedObsError` then got created which led to the solver showing zero increments. So there at least seems to be some expectation in JEDI for `ObsError` to exist even if it isn't used.

delippi:
@SamuelDegelia-NOAA, my guess is the logic is that since it is prepbufr it ought to be in IODA. I don't know how anyone else is using it or if they just specify directly. I assume when you run in split mode, you are unable to specify the ObsError in some part of that application? If that is the case, then I guess that is the only argument needed to keep it in. But then what would we do in the scenario where we're using bufr that doesn't have that info?

SamuelDegelia-NOAA:
@delippi I don't have the `ObsError`-less files anymore to confirm, but the problem was related to the `DerivedObsError` group. This group was created by default in the hofx files if `ObsError` did not exist. It was getting created even if we assigned the errors through the yaml. But `DerivedObsError` was always empty for me for some reason and that led to the zero increments. 

There might have been an easy fix for this, I just never really got around to trying much. Thinking about it now, maybe this could have been solved by removing `defer to post` from the assign error filter.

But yeah, if we don't have that info in the bufr files then this would be an issue to solve (way down the road).

hu5970 (approved):

Issue #195, opened 2024-10-13
What's this? `[NOAA-EMC/RDASApp] PR run failed: Unit Tests`


guoqing-noaa:
Recently, each time we made a commit to our own fork, we would get an email notifying that `[NOAA-EMC/RDASApp] PR run failed: Unit Tests` and the Github webpage showed:
![image](https://github.com/user-attachments/assets/7ef65088-bb43-4ad2-b9df-bdf4b5d33823)

Did we recently turn on the `Unit Tests`? 

If yes, could we make it work correctly? 
If not, could we turn it off until it can work correctly?
Thanks!

SamuelDegelia-NOAA:
This is probably related to https://github.com/NOAA-EMC/GDASApp/pull/1332

We have a couple of github actions in place for automated testing. It looks like they are checking for pycodestyle and then running a 3dvar test. But I am not sure why they always complete so fast. It seems that RDASApp isn't actually getting built? 

See the unit test workflow here: https://github.com/NOAA-EMC/RDASApp/blob/develop/.github/workflows/unittests.yaml and an example run here:  https://github.com/NOAA-EMC/RDASApp/actions/runs/10966022592/job/30453068901 

Anyways, to fix the error here, we can follow @CoryMartin-NOAA's PR by changing `ubuntu-latest` to `ubuntu-22.04`.

guoqing-noaa:
@SamuelDegelia-NOAA Could you go ahead to fix the error?  Thanks!

SamuelDegelia-NOAA:
Sure, will do!

guoqing-noaa:
This was solved. Thanks, @SamuelDegelia-NOAA

PR #198, opened 2024-10-15
Hotfix to get github actions working (no more error for failed unit test)

SamuelDegelia-NOAA:
Quick fix for #195 via @CoryMartin-NOAA. This should stop the error for a failed unit test when submitting a PR or adding a new commit.

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 jobs:
   ctests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     name: Run Unit Tests with ctest
 
     steps:
--- a/.github/workflows/norms.yaml
+++ b/.github/workflows/norms.yaml
@@ -3,7 +3,7 @@ on: [push]
 
 jobs:
   check_pynorms:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     name: Check coding norms with pycodestyle
 
     steps:


guoqing-noaa (approved):

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for the quick fix.

This is a hotfix, so I will merge it right now.

.github/workflows/unittests.yaml
```diff
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 jobs:
   ctests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
TingLei-NOAA, line 6: @SamuelDegelia-NOAA  Could you explain why the changes of the virtual machine type would resolve this issue?

SamuelDegelia-NOAA: @TingLei-NOAA I am not sure why it fixes it. The error was related to pip, so I am guessing that there was an update to the latest default ubuntu version that broke pip. @CoryMartin-NOAA might have more information.

CoryMartin-NOAA: @SamuelDegelia-NOAA 's hunch is the same as mine. I'd rather just deal with this later when that version of the virtual machine is no longer available.

TingLei-NOAA: @SamuelDegelia-NOAA @CoryMartin-NOAA  Thanks for this explanation.

PR #202, opened 2024-10-16
Add utility for updating ctest input files during development

SamuelDegelia-NOAA:
This PR adds a simple shell script based on `rrfs-test/CMakeLists.txt` that allows developers to update the RRFS ctests in RDASApp. The goal here is to be able to run this quick script if making any changes to the yaml files, test reference files, or input data for the ctests. Previously, the only way to do this was to rebuild all of RDASApp or make temporary changes to `build.sh`

Please see https://github.com/NOAA-EMC/RDASApp/pull/184#discussion_r1797962906 for additional motivation. 

Note: this script cannot update the ctest configuration (number of tests, test names, or MPI configuration) since those options are handled by `ecbuild` during the cmake step.


guoqing-noaa:
@SamuelDegelia-NOAA Thanks!

This is helpful. Just curious whether it is possible that the CMAKE process also uses the same script to stage data (similar as runnning `link_mpasjedi_expr.sh` in `buid.sh`) so that we don't need to maintain two parts?

SamuelDegelia-NOAA:
@guoqing-noaa We could theoretically do that, but we still need the cmake script to create the individual ctests (using `ecbuild_add_test`). So we would still need to lists the tests inside `rrfs-test/CMakeLists.txt` even if we went the route you describe. At that point I do not think we would save much by using this new script inside the cmake script.

guoqing-noaa:
> we still need the cmake script to create the individual ctests (using ecbuild_add_test)
Right. I totally agree with that. 

I mean using the same script will avoid making changes in two different locations. 
However, I am fine with as is if you wan to maintain two different sets of file copying/linking.

SamuelDegelia-NOAA:
Yeah, I think I would prefer to just keep the two sets. I don't see us needing to change the file copying/linking much. Thanks!

guoqing-noaa (approved):

ShunLiu-NOAA (approved):

hu5970 (approved):
looks good, merge this one

Issue #205, opened 2024-10-17
Python converter for RRFS/ [3D]-RTMA EFCLAM BUFR DATA


PraveenKumar-NOAA:
Description

Write a Python converter for the EFCLAM 3[D]-RTMA/RRFS data that converts a BUFR file to an IODA file using bufr2ioda.x for the following subsets:

| NC012160 | A61109 | M TYPE 012-160 U.WI GOES Imager Effective Cld Amt data |

Use Python API to convert a BUFR file into an IODA-formatted netCDF file
Ioda obs validation using ioda-validate.x
Perform validation for the observations using the input BUFR and output IODA file

Requirements
All important mnemonics required by the RRFS/ [3D]-RTMA users need to be known.

Acceptance Criteria
It will be finished when all files have been checked and tests have passed.

Related issue https://github.com/NOAA-EMC/RDASApp/issues/76

PraveenKumar-NOAA:
Validation of the output IODA results, obtained using the efclam iPython API, is performed for the following two observations:

**cloudAmount (ECAS)**
**BUFR plot:**

![bufr_efclam_cloudAmount (1)](https://github.com/user-attachments/assets/08e15045-7537-4ef2-9553-4e207f17de9f)

**IODA plot:**

![efclam-ioda_cloudAmount (1)](https://github.com/user-attachments/assets/4d5fedac-7981-4e27-8a19-2193b5ac412c)

**cloudCoverAmount (ECAM)**
**BUFR plot:**

![bufr_efclam_cloudCoverTotal (1)](https://github.com/user-attachments/assets/533f4eb1-b26e-4105-b83d-833b8f3d69de)

**IODA plot:**

![efclam-ioda_cloudCoverTotal (1)](https://github.com/user-attachments/assets/fcb404e6-191c-4c62-b90d-fa79bb0b2db7)

PR #212, opened 2024-10-29
Python converter for LGHTNG BUFR DUMP Data

PraveenKumar-NOAA:
This PR adds a Python converter for LGHTNG BUFR DUMP data.

The converter will be used to transform LGHTNG BUFR data into IODA (netCDF) format. Following steps similar to GDASApp, a Python file and a JSON description file were created.

The following two files are new:

~/IODA/python/bufr2ioda_lghtng.py
~/IODA/python/bufr2ioda_lghtng.json

Related issue: https://github.com/NOAA-EMC/RDASApp/issues/204

Validation for the output observation and IODA variable naming has been performed.

PR #213, opened 2024-10-29
Python converter for EFCLAM BUFR DUMP Data

PraveenKumar-NOAA:
This PR adds a Python converter for EFCLAM BUFR DUMP data.

The converter will be used to transform EFCLAM BUFR data into IODA (netCDF) format. Following steps similar to GDASApp, a Python file and a JSON description file were created.

The following two files are new:

~/IODA/python/bufr2ioda_efclam.py
~/IODA/python/bufr2ioda_efclam.json

Related issue: https://github.com/NOAA-EMC/RDASApp/issues/205

Validation for the output observation and IODA variable naming has been performed.

### DIFF
--- a/rrfs-test/IODA/python/bufr2ioda_efclam.json
+++ b/rrfs-test/IODA/python/bufr2ioda_efclam.json
@@ -1,0 +1,12 @@
+{
+    "data_format": "bufr_d",
+    "data_type": "efclam",
+    "cycle_type": "rap",
+    "cycle_datetime": "{{ current_cycle | to_YMDH }}",
+    "dump_directory": "{{ DMPDIR }}",
+    "ioda_directory": "{{ COM_OBS }}",
+    "subsets": [ "NC012160" ],
+    "source": "NCEP data tank",
+    "data_provider": "U.S. NOAA",
+    "data_description": "GOES imager effective cloud amount data (U.Wisc.)",
+}
--- a/rrfs-test/IODA/python/bufr2ioda_efclam.py
+++ b/rrfs-test/IODA/python/bufr2ioda_efclam.py
@@ -1,0 +1,252 @@
+#!/usr/bin/env python3
+# (C) Copyright 2024 NOAA/NWS/NCEP/EMC
+#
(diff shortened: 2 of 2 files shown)


SamuelDegelia-NOAA:
Hi @PraveenKumar-NOAA, do you happen to have a sample `efclam` bufr file that we can use with this PR? I believe I have the `lghtng` data to test #212.

PraveenKumar-NOAA:
> Hi @PraveenKumar-NOAA, do you happen to have a sample `efclam` bufr file that we can use with this PR? I believe I have the `lghtng` data to test #212.

@SamuelDegelia-NOAA there is no efclam data file in the RRFS test data, hence I used an RTMA file to validate the converter. I can send you this file or you are welcome to pick any efclam bufr file to test the converter.

SamuelDegelia-NOAA:
Could you send me the RTMA file? I do not have any other bufr files with efclam data at the moment. Thanks!

PraveenKumar-NOAA:
> Could you send me the RTMA file? I do not have any other bufr files with efclam data at the moment. Thanks!

@SamuelDegelia-NOAA please use this file from Hercules - /work2/noaa/da/pkumar/dump/rtma.t00z.efclam.tm00.bufr_d

SamuelDegelia-NOAA (approved):
Thanks @PraveenKumar-NOAA, this converter works well for me too! 

As an aside for any other reviewers, we need to make 1-2 small changes to the code here to test the converters inside RDASApp. This is because of how we name the staged bufr/prepbufr files for our MPAS test case. But these converters are meant to be general and file names might change in the workflow. So this behavior is expected for the current status of RDASApp.

delippi (approved):
It took a little bit to figure out what needed changed for the generic setup to work with our staged data, but wasn't too difficult. One thing I noticed was that the DUMP variable doesn't seem to be used in the `run_bufr2ioda.sh` that is in the wiki instructions, but instead uses cycle_type in the .json file. I think this is a good base to easily make updates for specific use in the workflow.

PR #215, opened 2024-11-01
Update the offline domain checks to use a concave hull

SamuelDegelia-NOAA:
This PR updates the offline domain checks to use a concave hull approach instead of a convex hull. This hull better matches the shape of the MPAS domain and thus can allow more observations through that are inside the domain. I still recommend using a small value of the hull_shrink_factor (0.5% worked for me) to avoid issues with observations located right on the domain boundary. 

This updated version of the domain check takes about 15 seconds longer than the original version. 

As mentioned by @delippi, the offline domain checks are meant as temporary solutions until the online domain check is fully functional. We are very close on that, but I want to go ahead and commit this update since it will be useful for validation of that online check. 

Usage: 

`./offline_domain_check_satrad.py -g invariant.nc -o gdas.t00z.atms_all_npp.tm00.nc -s 0.005 -f`

### DIFF
--- a/rrfs-test/IODA/offline_domain_check.py
+++ b/rrfs-test/IODA/offline_domain_check.py
@@ -2,7 +2,7 @@
 import netCDF4 as nc
 import numpy as np
 from matplotlib.path import Path
-from scipy.spatial import ConvexHull
+from scipy.spatial import ConvexHull, Delaunay
 from timeit import default_timer as timer
 import argparse
 import warnings
@@ -13,6 +13,10 @@ import cartopy.crs as ccrs
 import cartopy.feature as cfeature
 import matplotlib.ticker as mticker
 from cartopy.mpl.gridliner import LONGITUDE_FORMATTER, LATITUDE_FORMATTER
+from operator import itemgetter
+import shapely.speedups
+
+shapely.speedups.enable()
 
 """
 This program determines if observations are in/outside of a convex hull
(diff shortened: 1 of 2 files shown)


SamuelDegelia-NOAA:
Figure of the updated offline domain check for AIRCAR observations: 

Capture

ATMS radiances: 

Capture

ShunLiu-NOAA:
The update looks good to me.

guoqing-noaa (commented):
LGTM.

Just one minor comment, we can make all .py scripts executable by adding the follwoing shebang:

`#!/usr/bin/env python`

SamuelDegelia-NOAA:
Thanks @guoqing-noaa! I think that shebang is already added to the top of these python scripts. But I'll update the usage example in the PR text.

guoqing-noaa (approved):

xyzemc (approved):

HuiLiu-NOAA:
Sorry for the delay. I just did a test, and got 2 problems with my ZTD bufr2ioda file:

1. It seems need to add the next line after line 225 ( as Donnie commented last week) to use the ZTD input data with longitude range (-180, 180):
    obs_lon = np.where(obs_lon < 0, obs_lon + 360, obs_lon)

2. The variable "Obs_Error" got missing values in the output file for some reason.

The complete test results are in : /scratch1/NCEPDEV/fv3-cam/Hui.Liu/domain_check. Thanks for help checking these in advance.

SamuelDegelia-NOAA:
@HuiLiu-NOAA thanks for reviewing the PR and catching the longitude issue. I have that line in `offline_domain_check_satrad.py` but it is missing from `offline_domain_check.py`. I'll make a hotfix for that once we also have #&#x2060;2 solved. 

Speaking of - do you happen to have the ZTD data saved on any other machine? Hera is down today so I cannot investigate #&#x2060;2 right now.

HuiLiu-NOAA:
Thanks @SamuelDegelia-NOAA. We can wait for Hera back.

HuiLiu-NOAA:
@SamuelDegelia-NOAA : 
  I figured out the reason for the problem #2. It seems due to the "zenithTotalDelay:valid_range = 0.f, 0.f" in the original file. 
I reset the zenithTotalDelay:valid_range = 0.f, 0.01f, and the missing values disappeared.  Sorry for the confusion and thanks for your help.

PR #218, opened 2024-11-05
Add 187/287 METAR config templates

spanNOAA:
Finalize yamls for 187/287 metar observations with reasonable results.

See issue #46 for details.

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
@@ -39,7 +39,7 @@
          # ------------------
          # airTemperature (187)
          # ------------------
-         # Reject all obs with ObsType /= 187, QualityMarker = 4 -15
+         # Reject all observations with QualityMarker > 3
          - filter: RejectList
            apply at iterations: 0,1
            where:
@@ -71,8 +71,7 @@
            filter variables:
            - name: airTemperature
            where:
-           - variable:
-               name: ObsType/airTemperature
+           - variable: ObsType/airTemperature
              is_in: 187
            action:
              name: assign error
@@ -84,7 +83,7 @@
                  xvals: [110000, 105000, 100000, 95000, 90000, 85000, 80000, 75000, 70000, 65000, 60000, 55000, 50000, 45000, 40000, 35000, 30000, 25000, 20000, 15000, 10000, 7500, 5000, 4000, 3000, 2000, 1000, 500, 400, 300, 200, 100, 0]
                  errors: [2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585]
 
-         # error inflation based on pressure check (setupt.f90)
+         # Error inflation based on pressure check (setupt.f90)
          - filter: Perform Action
            filter variables:
            - name: airTemperature
@@ -102,6 +101,14 @@
                  # needs to be added to UFO to align with the ramp
                  # options for temperature and humidity in GSI.
                  #surface observation error ramp: true
+
+         # Bounds Check
+         - filter: Bounds Check
+           apply at iterations: 0,1
+           filter variables:
+           - name: airTemperature
+           minvalue: 100
+           maxvalue: 400
 
          # Create temporary ObsErrorData
          - filter: Variable Assignment
@@ -165,7 +172,7 @@
              name: reject
            defer to post: true
 
-         # Gross Error Check
+         # Gross Error Check: cgross*0.7 if QualityMarker=3
          - filter: Background Check
            apply at iterations: 0,1
            filter variables:
--- a/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
@@ -57,14 +57,6 @@
              minvalue: -900
              maxvalue:  900
 
-         # Temporal thinning filter
-         - filter: Temporal Thinning
-           apply at iterations: 0,1
-           min_spacing: PT30M
-           seed_time: *analysisDate
-           category_variable:
-             name: MetaData/stationIdentification
-
          # Initial error assignment
          - filter: Perform Action
            filter variables:
@@ -192,6 +184,15 @@
            - variable:
                name: TempObsErrorData/airTemperature
              value: is_valid
+           defer to post: true
+
+         # Temporal thinning filter
+         - filter: Temporal Thinning
+           apply at iterations: 0,1
+           min_spacing: PT30M
+           seed_time: *analysisDate
+           category_variable:
+             name: MetaData/stationIdentification
            defer to post: true
 
          #- filter: Print Filter Data
--- a/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
@@ -21,12 +21,12 @@
          name: Composite
          components:
          - name: VertInterp
-           variables:
-           - name: airTemperature
            vertical coordinate: air_pressure
            observation vertical coordinate: pressure
            observation vertical coordinate group: MetaData
            interpolation method: log-linear
+           variables:
+           - name: airTemperature
 
        obs error:
          covariance model: diagonal
@@ -41,6 +41,7 @@
          # ------------------
          # Reject all obs with ObsType /= 187, QualityMarker = 4 -15
          - filter: RejectList
+           apply at iterations: 0,1
            where:
            - variable: ObsType/airTemperature
              is_in: 187
@@ -49,6 +50,7 @@
 
          # Time window filter
          - filter: Domain Check
+           apply at iterations: 0,1
            where:
            - variable:
                name: MetaData/timeOffset
@@ -57,6 +59,7 @@
 
          # Temporal thinning filter
          - filter: Temporal Thinning
+           apply at iterations: 0,1
            min_spacing: PT30M
            seed_time: *analysisDate
            category_variable:
@@ -101,6 +104,7 @@
 
          # Create temporary ObsErrorData
          - filter: Variable Assignment
+           apply at iterations: 0,1
            assignments:
            - name: TempObsErrorData/airTemperature
              type: float
@@ -109,9 +113,11 @@
                options:
                  variables:
                  - name: ObsErrorData/airTemperature
+           defer to post: true
 
          # Set ObsError set "error parameter" if < "max value"
          - filter: Perform Action
+           apply at iterations: 0,1
            filter variables:
            - name: airTemperature
            action:
@@ -124,9 +130,11 @@
            - variable:
                name: ObsErrorData/airTemperature
              value: is_valid
+           defer to post: true
 
          # Set ObsError set "error parameter" if > "min value"
          - filter: Perform Action
+           apply at iterations: 0,1
            filter variables:
            - name: airTemperature
            action:
@@ -139,9 +147,11 @@
            - variable:
                name: ObsErrorData/airTemperature
              value: is_valid
+           defer to post: true
 
          # Gross Error Check
          - filter: Background Check
+           apply at iterations: 0,1
            filter variables:
            - name: airTemperature
            threshold: 5.0
@@ -152,9 +162,11 @@
              is_not_in: 3
            action:
              name: reject
+           defer to post: true
 
          # Gross Error Check
          - filter: Background Check
+           apply at iterations: 0,1
            filter variables:
            - name: airTemperature
            threshold: 3.5
@@ -165,10 +177,12 @@
              is_in: 3
            action:
              name: reject
+           defer to post: true
 
          # Re-assign err ObsErrorData <--- TempObsErrorData
          # after gross error check.
          - filter: Perform Action
+           apply at iterations: 0,1
            filter variables:
            - name: airTemperature
            action:
@@ -178,6 +192,7 @@
            - variable:
                name: TempObsErrorData/airTemperature
              value: is_valid
+           defer to post: true
 
          #- filter: Print Filter Data
          #  message: Printing filter data
--- a/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
@@ -187,8 +187,7 @@
              name: reject
            defer to post: true
 
-         # Re-assign err ObsErrorData <--- TempObsErrorData
-         # after gross error check.
+         # Re-assign err ObsErrorData <--- TempObsErrorData after gross error check.
          - filter: Perform Action
            apply at iterations: 0,1
            filter variables:
@@ -202,7 +201,7 @@
              value: is_valid
            defer to post: true
 
-         # Temporal thinning filter
+         # Duplicate Check
          - filter: Temporal Thinning
            apply at iterations: 0,1
            min_spacing: PT30M
@@ -210,17 +209,6 @@
            category_variable:
              name: MetaData/stationIdentification
            defer to post: true
-
-         #- filter: Print Filter Data
-         #  message: Printing filter data
-         #  summary: true
-         #  variables:
-         #  - variable: MetaData/latitude
-         #  - variable: MetaData/longitude
-         #  - variable: MetaData/pressure
-         #  - variable: ObsType/airTemperature
-         #  - variable: ObsValue/airTemperature
-         #  - variable: QCflagsData/airTemperature
 
          #- filter: GOMsaver
          #  filename: ./data/geovals/msonet_geovals_rrfs.nc4
--- a/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
@@ -1,0 +1,194 @@
+     - obs space:
+         name: adpsfc_airTemperature_187
+         distribution:
+           name: "@DISTRIBUTION@"
+           halo size: 100e3
+         obsdatain:
+           engine:
+             type: H5File
+             obsfile: "@OBSFILE@"
+         obsdataout:
+           engine:
+             type: H5File
+             obsfile: hofx_adpsfc_airTemperature_187.nc4
+             allow overwrite: true
+         io pool:
+           max pool size: 1
+         observed variables: [airTemperature]
+         simulated variables: [airTemperature]
+
+       obs operator:
+         name: Composite
+         components:
+         - name: VertInterp
+           variables:
+           - name: airTemperature
+           vertical coordinate: air_pressure
+           observation vertical coordinate: pressure
+           observation vertical coordinate group: MetaData
+           interpolation method: log-linear
+
+       obs error:
+         covariance model: diagonal
+
+       obs localizations:
+         - localization method: Horizontal Gaspari-Cohn
+           lengthscale: 300e3 # orig
+
+       obs filters:
+         # ------------------
+         # airTemperature (187)
+         # ------------------
+         # Reject all obs with ObsType /= 187, QualityMarker = 4 -15
+         - filter: RejectList
+           where:
+           - variable: ObsType/airTemperature
+             is_in: 187
+           - variable: QualityMarker/airTemperature
+             is_in: 4-15
+
+         # Time window filter
+         - filter: Domain Check
+           where:
+           - variable:
+               name: MetaData/timeOffset
+             minvalue: -900
+             maxvalue:  900
+
+         # Temporal thinning filter
+         - filter: Temporal Thinning
+           min_spacing: PT30M
+           seed_time: *analysisDate
+           category_variable:
+             name: MetaData/stationIdentification
+
+         # Initial error assignment
+         - filter: Perform Action
+           filter variables:
+           - name: airTemperature
+           where:
+           - variable:
+               name: ObsType/airTemperature
+             is_in: 187
+           action:
+             name: assign error
+             error function:
+               name: ObsFunction/ObsErrorModelStepwiseLinear
+               options:
+                 xvar:
+                   name: MetaData/pressure
+                 xvals: [110000, 105000, 100000, 95000, 90000, 85000, 80000, 75000, 70000, 65000, 60000, 55000, 50000, 45000, 40000, 35000, 30000, 25000, 20000, 15000, 10000, 7500, 5000, 4000, 3000, 2000, 1000, 500, 400, 300, 200, 100, 0]
+                 errors: [2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585]
+
+         # error inflation based on pressure check (setupt.f90)
+         - filter: Perform Action
+           filter variables:
+           - name: airTemperature
+           where:
+           - variable: ObsType/airTemperature
+             is_in: 187
+           action:
+             name: inflate error
+             inflation variable:
+               name: ObsFunction/ObsErrorFactorPressureCheck
+               options:
+                 variable: airTemperature
+                 inflation factor: 8.0
+                 # The new feature "surface observation error ramp"
+                 # needs to be added to UFO to align with the ramp
+                 # options for temperature and humidity in GSI.
+                 #surface observation error ramp: true
+
+         # Create temporary ObsErrorData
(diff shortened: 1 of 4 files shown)


rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,194 @@
...
+               name: MetaData/timeOffset
+             minvalue: -900
+             maxvalue:  900
+
+         # Temporal thinning filter
+         - filter: Temporal Thinning
+           min_spacing: PT30M
+           seed_time: *analysisDate
+           category_variable:
+             name: MetaData/stationIdentification
```
delippi: @spanNOAA, are you finding that this is needed to match ob counts and increments? I found that for mesonet that this was actually detrimental for my results. Are you mainly using this to remove duplicate observations? If so, there could be other things we could try to handle that. I don't think the data is temporally thinned in GSI, so this "might" not be the best. Just something to consider.

spanNOAA: I found that removing duplicate METAR observations actually improved my results. In my case, eliminating them improved the analysis, as the duplicates may add additional value at the same location even if the information was outdated or from a significantly later time. 

I believe the detrimental results when applying temporal thinning to mesonet observations may be due to the smaller observation window used. For mesonet data, a ±6 minute window is applied, while for METAR data, a larger ±15 minute window is used. This is because some METAR stations have less frequent observations. As a result, applying a smaller window may cause some stations to be dropped, while others might have multiple observations within the ±15 minute window. 

Furthermore, GSI applies a form of "temporal thinning" when the l_closeobs option is enabled. This method retains only the observation closest to the analysis time and discards the others. Below is the relevant code:
```f90
if (l_closeobs) then
    if(abs(data(itime,k)-hr_offset)<abs(data(itime,l)-hr_offset)) then
        muse(l)=.false.
    else
        muse(k)=.false.
    endif
else
    tfact=min(one,abs(data(itime,k)-data(itime,l))/dfact1)
    dup(k)=dup(k)+one-tfact*tfact*(one-dfact)
    dup(l)=dup(l)+one-tfact*tfact*(one-dfact)
endif
```

Temporal thinning works well for temperature, humidity, and wind observations. However, I've had an issue with temporal thinning for surface pressure in my case. Specifically, surface observations in Canada were being dropped when temporal thinning was enabled. Theoretically, there should always be at least one observation at each station. I'm still investigating the cause of this issue.

30M spacing
image
0M spacing
image

hu5970: We should try to remove the duplicated observations from surface stations. @delippi Could you provide more details on how this filter detriment your MESONET results?

delippi: Here are some runs I just made for mesonet using the temporal filter. The time window is +/-6 min. "No filter" gives better observation counts (counts are in the image; lower left corner), but increments look largely the same. The difference (between no filter and PT01M) is hard to spot but mainly just over west Texas. No temporal filter just seems better to me. PT30M definitely seems not good for this case. It could be argued that PT00M or PT01M would be fine.

**No Temporal filter**: (This gives different result from PT00M)
fv3jedi_vs_gsi_increment_airTemperature_msonet_airTemperature_188_None
**PT01M**:
fv3jedi_vs_gsi_increment_airTemperature_msonet_airTemperature_188_PT01M
**PT30M**:
fv3jedi_vs_gsi_increment_airTemperature_msonet_airTemperature_188_PT30M

SamuelDegelia-NOAA:
Thanks @spanNOAA! One thing I noticed is that the filter order is slightly different from the mesonet yaml files (e.g., [here](https://github.com/NOAA-EMC/RDASApp/blob/develop/rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml)). The mesonet obs have temporal thinning (duplicate check) at the end while these yamls have it towards the beginning. Do you know if that order matters? 

Also, could we add the online domain check here? That filter is commented out of the q/uv yamls but we should add it to all of them.

delippi:
> Thanks @spanNOAA! One thing I noticed is that the filter order is slightly different from the mesonet yaml files (e.g., [here](https://github.com/NOAA-EMC/RDASApp/blob/develop/rrfs-test/validated_yamls/templates/obtype_config/msonet_airTemperature_188.yaml)). The mesonet obs have temporal thinning (duplicate check) at the end while these yamls have it towards the beginning. Do you know if that order matters?
> 
> Also, could we add the online domain check here? That filter is commented out of the q/uv yamls but we should add it to all of them.

I think the order does matter. I'm not 100% sure what the correct order should be. It seems to me that GSI has the duplicate checks earlier in the code, so I did try putting the duplicate filter toward the top of the yaml. The problem I was seeing there is you could be throwing out obs via the duplicate filter that are better quality and the lesser quality fail the gross checks. If it is near the bottom, then you allow all obs to go through the gross checks first and then pick from those which duplicates to keep. While writing this, I'm starting to think maybe we do want the more restrictive GSI method?

spanNOAA:
Thanks @SamuelDegelia-NOAA @delippi pointing that out! I hadn't noticed this difference before, probably because of the limited number of METAR observations, but it's definitely a good catch regarding the order issue. I’ve changed the order, and the result hasn't changed—the cost function at the last step is exactly the same. I'll update the order to match yours for consistency.
I still have some questions about the online domain check. Is it meant to replace the offline domain check? How should I properly set the min/max values for the observable_domain_mask? Do you have any instructions on how to use this function?

delippi:
When I updated the submodules in PR #227, that should have fixed issues we were having with the online domain check which should now be able to be used alone. There should no longer be a need to use the offline domain check... Although it can still be helpful for debugging.

This is how you use the online domain check according to @HuiLiu-NOAA. I'm not sure exactly what the min/max values are doing.
```
         # Online domain check
         - filter: Bounds Check
           filter variables:
           - name: airTemperature
           test variables:
           - name:  GeoVaLs/observable_domain_mask
           minvalue: 0.0
           maxvalue: 0.5
```

I've placed this filter as one of the first filters and right after the "Time window filter" (the one that uses timeOffset). The main thing is that it just needs to be used before ObsErrorFactorPressureCheck.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
```diff
@@ -0,0 +1,213 @@
...
+           where:
+           - variable: ObsType/windEastward
+             is_in: 287
+           action:
+             name: inflate error
+             inflation variable:
+               name: ObsFunction/ObsErrorFactorPressureCheck
+               options:
+                 variable: windEastward
+                 inflation factor: 4.0
```
delippi: Sorry if I missed it, but there should be two ObsErrorFactorPressureCheck's for the winds yamls: one for windEastward and one for windNorthward. There are also some new options that I recently added. The section should look like this:
```
         # Error inflation based on pressure check (setupw.f90)
         - filter: Perform Action
           filter variables:
           - name: windEastward
           where:
           - variable: ObsType/windEastward
             is_in: 287
           action:
             name: inflate error
             inflation variable:
               name: ObsFunction/ObsErrorFactorPressureCheck
               options:
                 variable: windEastward
                 inflation factor: 4.0
                 SetSfcWndObsHeight: true
                 AddObsHeightToStationElevation: true
                 AssumedSfcWndObsHeight: 10

         # Error inflation based on pressure check (setupw.f90)
         - filter: Perform Action
           filter variables:
           - name: windNorthward
           where:
           - variable: ObsType/windNorthward
             is_in: 287
           action:
             name: inflate error
             inflation variable:
               name: ObsFunction/ObsErrorFactorPressureCheck
               options:
                 variable: windNorthward
                 inflation factor: 4.0
                 SetSfcWndObsHeight: true
                 AddObsHeightToStationElevation: true
                 AssumedSfcWndObsHeight: 10

```

spanNOAA: Are these new options part of your PR for UFO?

spanNOAA:
@HuiLiu-NOAA Could you provide more information about the observable_domain_mask? Specifically, what do the minimum and maximum values represent? Thanks.

HuiLiu-NOAA:
> @HuiLiu-NOAA Could you provide more information about the observable_domain_mask? Specifically, what do the minimum and maximum values represent? Thanks.

Yes. Here are the details:
For MPAS, the domain mask has values of 0-7; where 0=interior, 1-4=relax zone points, and 5-7=prescribed zone.  Observations in these different zones could be used/evaluated in DA. Current, the interpolated value of the mask at OBS location with max=0.5 include only the OBS inside domain interior for DA. The max value can be relaxed to any values between 1.0 and 7.0  to consider/include OBS in the appropriate zones for DA.

For FV3, its domain mask is relative simple right now, 0=interior, 1=outside of domain.  So, the interpolated mask value at OBS location with max=0.5 consider only the OBS inside the FV3 domain. This could be changed in future development.

spanNOAA:
@HuiLiu-NOAA Thanks for the details. @SamuelDegelia-NOAA When applying the online check, we do have more data points at the edge of the domain compared to the offline check.
image

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
delippi: Yes the options `SetSfcWndObsHeight`, `AddObsHeightToStationElevation`, and `AssumedSfcWndObsHeight` were what I added in one of my recent UFO PRs. There was a fourth option that doesn't apply to mesonet or metar. Mesonet and metar, I believe, are calculated the same way: `obs_height = stationElevation + 10` which is what these settings as I have provided would do.

SamuelDegelia-NOAA:
> @HuiLiu-NOAA Thanks for the details. @SamuelDegelia-NOAA When applying the online check, we do have more data points at the edge of the domain compared to the offline check. 

That's great! Even if its a small difference, glad that we don't have to remove any obs inside the domain.

SamuelDegelia-NOAA (dismissed):

delippi:
> > @HuiLiu-NOAA Could you provide more information about the observable_domain_mask? Specifically, what do the minimum and maximum values represent? Thanks.
> 
> Yes. Here are the details: For MPAS, the domain mask has values of 0-7; where 0=interior, 1-4=relax zone points, and 5-7=prescribed zone. Observations in these different zones could be used/evaluated in DA. Current, the interpolated value of the mask at OBS location with max=0.5 include only the OBS inside domain interior for DA. The max value can be relaxed to any values between 1.0 and 7.0 to consider/include OBS in the appropriate zones for DA.
> 
> For FV3, its domain mask is relative simple right now, 0=interior, 1=outside of domain. So, the interpolated mask value at OBS location with max=0.5 consider only the OBS inside the FV3 domain. This could be changed in future development.

@HuiLiu-NOAA, thanks for this info. Does this update the `QCflagData` group after it is run? My understanding is that `QCflagData` are up-to-date values (set by one of the preceding filters). `EffectiveQC` are final values obtained after execution of all filters; this group does not exist while filters are running, but is present in the output `ioda` file.

HuiLiu-NOAA:
> > > @HuiLiu-NOAA Could you provide more information about the observable_domain_mask? Specifically, what do the minimum and maximum values represent? Thanks.
> > 
> > 
> > Yes. Here are the details: For MPAS, the domain mask has values of 0-7; where 0=interior, 1-4=relax zone points, and 5-7=prescribed zone. Observations in these different zones could be used/evaluated in DA. Current, the interpolated value of the mask at OBS location with max=0.5 include only the OBS inside domain interior for DA. The max value can be relaxed to any values between 1.0 and 7.0 to consider/include OBS in the appropriate zones for DA.
> > For FV3, its domain mask is relative simple right now, 0=interior, 1=outside of domain. So, the interpolated mask value at OBS location with max=0.5 consider only the OBS inside the FV3 domain. This could be changed in future development.
> 
> @HuiLiu-NOAA, thanks for this info. Does this update the `QCflagData` group after it is run? My understanding is that `QCflagData` are up-to-date values (set by one of the preceding filters). `EffectiveQC` are final values obtained after execution of all filters; this group does not exist while filters are running, but is present in the output `ioda` file.

I believe this updates the QCflagData group after it is run. This can be verified by check the QCflags in the diag file.

ShunLiu-NOAA:
@delippi could you please review the updates of this PR?

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+
+       obs localizations:
+         - localization method: Horizontal Gaspari-Cohn
+           lengthscale: 300e3 # orig
+
+       obs filters:
+         # ------------------
+         # airTemperature (187)
+         # ------------------
+         # Reject all obs with ObsType /= 187, QualityMarker = 4 -15
```
delippi: Sorry for the picky comment. This comment states, "Reject all obs with ObsType <ins>not equal to</ins> 187...". I just used "Reject all obs with QualityMarker > 3" or something like that is fine.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+           maxvalue: 0.5
+
+         # Initial error assignment
+         - filter: Perform Action
+           filter variables:
+           - name: airTemperature
+           where:
+           - variable:
+               name: ObsType/airTemperature
+             is_in: 187
```
delippi: I sense I'm going to be leaving several picky comments. Sorry in advance!!! I leave them just for better consistency with the mesonet yamls. Could these 3 lines be changed to a 2-liner:
```
           - variable: ObsType/airTemperature
             is_in: 187 
```

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+             name: assign error
+             error function:
+               name: ObsFunction/ObsErrorModelStepwiseLinear
+               options:
+                 xvar:
+                   name: MetaData/pressure
+                 xvals: [110000, 105000, 100000, 95000, 90000, 85000, 80000, 75000, 70000, 65000, 60000, 55000, 50000, 45000, 40000, 35000, 30000, 25000, 20000, 15000, 10000, 7500, 5000, 4000, 3000, 2000, 1000, 500, 400, 300, 200, 100, 0]
+                 errors: [2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585, 2.2585]
+
+         # error inflation based on pressure check (setupt.f90)
```
delippi: Capitalize the "e" in  "error".

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+               name: ObsFunction/ObsErrorFactorPressureCheck
+               options:
+                 variable: airTemperature
+                 inflation factor: 8.0
+                 # The new feature "surface observation error ramp"
+                 # needs to be added to UFO to align with the ramp
+                 # options for temperature and humidity in GSI.
+                 #surface observation error ramp: true
+
+         # Create temporary ObsErrorData
```
delippi: I put a "Bounds Check" here as a sanity check for bad values. This is just what I used.
```
         # Bounds Check
         - filter: Bounds Check
           apply at iterations: 0,1
           filter variables:
           - name: airTemperature
           minvalue: 100
           maxvalue: 400
```

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+           where:
+           - variable: ObsType/airTemperature
+             is_in: 187
+           - variable: QualityMarker/airTemperature
+             is_not_in: 3
+           action:
+             name: reject
+           defer to post: true
+
+         # Gross Error Check
```
delippi: Can we improve this comment to be a little more specific? I used "Gross Error Check: cgross*0.7 if QualityMarker=3"

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+           - variable: ObsType/airTemperature
+             is_in: 187
+           - variable: QualityMarker/airTemperature
+             is_in: 3
+           action:
+             name: reject
+           defer to post: true
+
+         # Re-assign err ObsErrorData <--- TempObsErrorData
+         # after gross error check.
```
delippi: Combine these into a single line.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+           action:
+             name: assign error
+             error function: TempObsErrorData/airTemperature
+           where:
+           - variable:
+               name: TempObsErrorData/airTemperature
+             value: is_valid
+           defer to post: true
+
+         # Temporal thinning filter
```
delippi: Instead of "Temporal thinning filter", can we be more descriptive in the comment and note this as "Duplicate Check"?

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+         - filter: Temporal Thinning
+           apply at iterations: 0,1
+           min_spacing: PT30M
```
delippi, line 207: No change needed here, but I just wanted to say that I'm still unsure about using PT30M for this data. The time window is 15M. We both have found this filter does some odd things. If PT30M gives you best results then I am OK.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -0,0 +1,219 @@
...
+         #  message: Printing filter data
+         #  summary: true
+         #  variables:
+         #  - variable: MetaData/latitude
+         #  - variable: MetaData/longitude
+         #  - variable: MetaData/pressure
+         #  - variable: ObsType/airTemperature
+         #  - variable: ObsValue/airTemperature
+         #  - variable: QCflagsData/airTemperature
+
```
delippi: Can we just remove these lines?

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_stationPressure_187.yaml
```diff
@@ -0,0 +1,227 @@
...
+           filter variables:
+           - name: stationPressure
+           action:
+             name: inflate error
+             inflation factor: 1.2
+           where:
+           - variable: ObsType/stationPressure
+             is_in: 187
+           - variable: QualityMarker/stationPressure
+             is_in: 3
```
delippi: Do we need this? I'm not sure.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
```diff
@@ -0,0 +1,235 @@
...
+                 AssumedSfcWndObsHeight: 10
+
+         - filter: Perform Action
+           filter variables:
+           - name: windNorthward
+           where:
+           - variable: ObsType/windNorthward
+             is_in: 287
+           action:
+             name: inflate error
```
delippi: This filter is missing a comment. "Error inflation based on pressure check (setupw.f90)"

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
```diff
@@ -0,0 +1,235 @@
...
+           action:
+             name: reject
+           defer to post: true
+
+         - filter: Background Check
+           apply at iterations: 0,1
+           filter variables:
+           - name: windNorthward
+           function absolute threshold:
+           - name: ObsFunction/WindsSPDBCheck
```
delippi, line 170: Missing a comment. "Gross Error Check"

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
```diff
@@ -0,0 +1,235 @@
...
+           action:
+             name: reject
+           defer to post: true
+
+         - filter: Background Check
+           apply at iterations: 0,1
+           filter variables:
+           - name: windEastward
+           function absolute threshold:
+           - name: ObsFunction/WindsSPDBCheck
```
delippi, line 190: Missing a comment. "Gross Error Check: cgross*0.7 if QualityMarker=3"

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
```diff
@@ -0,0 +1,235 @@
...
+           action:
+             name: reject
+           defer to post: true
+
+         - filter: Background Check
+           apply at iterations: 0,1
+           filter variables:
+           - name: windNorthward
+           function absolute threshold:
+           - name: ObsFunction/WindsSPDBCheck
```
delippi: Missing comment. "Gross Error Check: cgross*0.7 if QualityMarker=3"

delippi (commented):
Hi @spanNOAA, I'm so sorry for all the picky comments. There were many comments that I made in the airTemperature yaml that also apply to the other yamls but <ins>I did not</ins> add those comments to the other yamls; I was trying to keep picky, repetitive comments to a minimum. So when you make a change for the airTemperature yaml, can you also check to see if it applies to the others? You can also use `vi -d msonet_stationPressure_188.yaml adpsfc_stationPressure_187.yaml` (for example) to compare. It would be nice if differences were kept as few as possible! This makes debugging and maintenance of these yamls much much easier!

Overall, everything looks very nice! Thanks for putting these together and mostly following the mesonet yamls. I think the differences here were just based on older versions of the mesonet yamls which recent updates never got propagated into these yamls.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
spanNOAA: That's a great question. Previously, I only applied a filter to reject obs with QualityMarker > 3, but I noticed that even after assigning errors to just 187 observations, there were still some obs with types other than 187. This didn’t make sense to me, so I added this filter to remove any observations that weren’t of type 187.

spanNOAA: Oh, are you suggesting changing the comment to something like "Reject all observations with QualityMarker > 3"?

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_stationPressure_187.yaml
spanNOAA: I'll remove this part, but we should keep in mind that my result shows a slightly larger increment compared to the GSI output even with this inflation applied.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
spanNOAA: I thought this comment was redundant, similar to the one for windEastward. To me, windEastward and windNorthward should be treated together as a whole. Do we really need separate comments for each of them?

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
delippi: Yes, just the comment. But you are right, adpsfc contains more than 187 which I forgot. I think it is fine with the update since the obs space is for only adpsfc 187.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
delippi: In my opinion, adding a second comment before the second `- filter: Background Check` might feel redundant, but it does improve readability, especially for someone less familiar with the YAML file. Comments act as visual markers, helping to quickly navigate through sections and maintain structure if you’re consistently labeling functions or blocks. They also clarify that the second block is part of the same conceptual grouping. To avoid redundancy, you could tweak the second comment to make it more specific. For example:
```
# Gross Error Check (windEastward)
```

I won't hold up this PR if you ultimately choose not to include the comment.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
spanNOAA: The 30M interval gives better results for METAR temperature, humidity, and wind, so I'll leave it as is.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_winds_287.yaml
spanNOAA: I think it's a better idea to include "windEastward" or "windNorthward".

delippi: Thanks! I'm good with that. I will have to update my mesonet yamls to this standard next time that I make an update to them. I was thinking about trying to test `ObsErrorFactorConventional` with the mesonet obs soon.

delippi (approved):

delippi:
@spanNOAA, thanks for making all those changes!

ShunLiu-NOAA (approved):

PR #222, opened 2024-11-07
Hotfix for offline domain check longitude range

SamuelDegelia-NOAA:
A quick fix for the problem described [here](https://github.com/NOAA-EMC/RDASApp/pull/215#issuecomment-2456248208) by @HuiLiu-NOAA. The conventional version of the offline domain check was not configured to work with obs whose longitudes range from (-180, 180). This line exists in `offline_domain_check_satrad.py` but needs to be added to `offline_domain_check.py`.

### DIFF
--- a/rrfs-test/IODA/offline_domain_check.py
+++ b/rrfs-test/IODA/offline_domain_check.py
@@ -223,6 +223,7 @@ domain_path = Path(edge_points)
 # Extract observation latitudes and longitudes
 obs_lat = obs_ds.groups['MetaData'].variables['latitude'][:]
 obs_lon = obs_ds.groups['MetaData'].variables['longitude'][:]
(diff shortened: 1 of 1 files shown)


delippi (approved):
LGTM! This is an easy one. Thanks for making the PR for this.

HuiLiu-NOAA (approved):
Thanks! I think this offline tool would remain very useful even after the online one is fully functional.

guoqing-noaa (approved):

delippi:
> Thanks! I think this offline tool would remain very useful even after the online one is fully functional.

I agree with this statement. Just the other day, it came up about matching ob counts when we do FV3 vs MPAS comparisons since grids wont match 100%. We can use the smaller of the two grids using the offline domain check to get the exact number of obs when doing our comparisons.

hu5970 (approved):

Issue #240, opened 2024-11-27
`ombg, oman, innov1, innov2`, etc


guoqing-noaa:
@Junjun-NOAA @spanNOAA and I did experiments and confirmed that:
since this PR https://github.com/JCSDA-internal/oops/pull/2749 which was merged on Sept 24th,
`ombg` always refers to the O-B before any outerloops
`oman` always refers to the O-A after all outerloops
`innov1` always refers to the O-F after the 1st outerloop and before the 2nd outerloop
`innov2` always refers to the O-F after the 2nd outerloop and before the 3rd outerloop
....

These diagnostics are always generated in the latest RDASApp and we no longer need to specify the `diagnostics.departures` keys in the yaml file.
i.e. let's remove the following lines in our RDASApp yaml files.
```
diagnostics:
  departures: ombg
```
Here is a snapshot of the relevant codes:
https://github.com/JCSDA/oops/commit/f099b356a89c93312556b99c75cab57861465351#diff-6f90c48cceb31841da068f513e80c30efb49d27800cd1d0ac988931f8a62e37fL210

![image](https://github.com/user-attachments/assets/c3ba783f-9173-4e6a-ae05-75c5fa95527c)

I have tested the new yaml file without the `diagnostics.departures` keys and it worked as expected.
Here is my test directory and yaml file:
`/scratch1/BMC/wrfruc/gge/RDASApp/expr/mpas_2024052700_save/rrfs_mpasjedi_2024052700_Ens3Dvar.yaml`
and the jdiag file:
`/scratch1/BMC/wrfruc/gge/RDASApp/expr/mpas_2024052700_save/jdiag.nc4`

Tag @HaidaoLin-NOAA @SamuelDegelia-NOAA as this also addresses issue #229

guoqing-noaa:
Based on the above code change, I did tests and also confirmed that we can remove this section and still get the `oman` in the jdiag files.
```
final:
  diagnostics:
    departures: oman
```

guoqing-noaa:
Also for information, RDASApp included the aforementioned OOPS PR https://github.com/JCSDA-internal/oops/pull/2749 since Nov. 6th through Sam's PR #217 which updated OOPS from `d7721732` (Aug 29, 2024, OOPS PR #2719) to `b2e6d273` (Oct 22, 2024, OOPS PR #2760)

PR #238, opened 2024-11-27
Update FV3-JEDI ctest case to match MPAS-JEDI case 

SamuelDegelia-NOAA:
## Description

This PR unifies the FV3-JEDI and MPAS-JEDI test cases in RDASApp. Now the FV3-JEDI test case and ctests are valid at the same time (2024-05-27, 00:00 UTC) which allows for better comparison between GSI (RRFSv1), FV3-JEDI, and MPAS-JEDI. Thanks to @chunhuazhou for providing the background files for this case! 

The case data are now linked in to `RDASApp/fix/expr_data/fv3_2024052700` through `init.sh` (similar to the MPAS-JEDI test data). The `link_fv3jedi_expr.sh` script is then added to create a new experiment directory during the build called `RDASApp/expr/fv3_2024052700`. New bumploc files and template jobs for running fv3-jedi and bump are also added for running this new case. The reference data are also updated for the new case.  

Staged data and tools are also added for running GSI-based EnVar for this test case. The GSI configuration is taken from the RRFS_A retros on WCOSS2 with some minor changes including disabling dBZ state variables, disabling the hybrid (now using pure EnVar), a different mesonetuselist, and turning off `ext_sonde`. 

Given that some developers are still using the old FV3-JEDI test for validation, the old case is still linked into `RDASApp/fix/rrfs-test-data/rrfs-data_fv3jedi_2022052619` and `RDASApp/bundle/rrfs-test-data/rrfs-data_fv3jedi_2022052619`. As such, the old test data can still be accessed for validation experiments using `RDASApp/scripts/setup_experiment.sh`. But only the new test data is linked into `RDASAppp/fix/expr_data` and used for the ctests. 

## Issue(s) addressed
No issue was opened for this PR. 

## Dependencies (if applicable)
None. 

## Checklist
- [x] I have performed a self-review of my own code.
- [x] I have run rrfs tests before creating the PR (if applicable).
- [x] I have staged the new FV3-LAM test data on `RDAS_DATA`
- [x] I have staged the GSI test data on `RDAS_DATA`

PR #241, opened 2024-11-28
remove no-longer-needed `diagnostics.departures` and rename jedi diagnostic files as jdiag*

guoqing-noaa:
As discussed in issue #240 , we no longer need to specify `diagnostics.departures` in YAML files and JEDI will always output `ombg, oman, innov1` for two outer loop minimization

Also, the JEDI diagnostic file contains much more information than `hofx`. It is better to rename the JEDI diagnostic files as `jdiag*`, for example `jdiag_adpupa.nc4` to be consistent with the reality (the word `jdiag` has been used for a long time in the python plotting scripts).


rrfs-test/testinput/rrfs_fv3jedi_hyb_2022052619.yaml
```diff
@@ -109,7 +109,7 @@ cost function:
          obsdataout:
            engine:
              type: H5File
-             obsfile: rass_tsen_obs_2022052619_hofx.nc4
+             obsfile: jdiag_rass_tsen_obs_2022052619.nc4
```
guoqing-noaa: Copy lines 101-112 here:
```
  observations:
     observers:
     - obs space:
         name: Aircraft
         obsdatain:
           engine:
             type: H5File
             obsfile: Data/obs/rass_tsen_obs_2022052619.nc4
         obsdataout:
           engine:
             type: H5File
             obsfile: jdiag_rass_tsen_obs_2022052619.nc4
```

RASS is the ground-based "Radio Acoustic Sounding System". Just curious why Line 104 sets the obs_space.name as "Aircraft"?

SamuelDegelia-NOAA (dismissed):
Good changes, thanks for doing this @guoqing-noaa. Also glad to see more verification that the `ombg` and `oman` diagnostics are working correctly now.

ShunLiu-NOAA (dismissed):

rrfs-test/testinput/rrfs_fv3jedi_hyb_2022052619.yaml
delippi: @guoqing-noaa, probably just a mistake. I think we should try to clean up those obs space names to match the obs. There were a few other places where multiple obs spaces had the same name in the same yaml. Will changing those names affect ctest results. I would think not, but I'm not sure how the ctest references work and if they use the obs space names or not.

SamuelDegelia-NOAA: The fv3-jedi yamls were restructured in #238 to use the same (or very close) observations to mpas-jedi, so we shouldn't need to worry about this particular bug anymore. But the new fv3-jedi yamls will need to have the corresponding changes made (renaming `hofxs` to `jdiag` and removing the `ombg` and `oman` diagnostics) as part of this PR. Sorry @guoqing-noaa!

guoqing-noaa: Thanks, @delippi and @SamuelDegelia-NOAA !

Yes, this issue is not related to this PR. But since I noticed this, I think I can bring this up for awareness. 
Thanks for addressing this issue in #238! No worries, I can resolve the conflicts once #238 is merged.

guoqing-noaa:
@ShunLiu-NOAA @SamuelDegelia-NOAA @delippi 

I have merged the latest develop branch, resolved the conflicts, and updated the newly-added fv3jedi yaml files accordingly.

SamuelDegelia-NOAA (approved):

SamuelDegelia-NOAA:
On a sort of unrelated note, do we know if anyone is using the yamls contained in `rrfs-test/testinput_expr`? This directory contains our older yamls from before updating the ctests to use more obs types. These yamls are copied into the run directory created by `setup_experiment.sh`, but I do not know if anyone is actually using the yamls (or just running `setup_experiment.sh` solely to create the run directory). If not, it would be best to remove `testinput_expr` to avoid having to keep updating these yamls for developments like this PR.

delippi (approved):

delippi:
> On a sort of unrelated note, do we know if anyone is using the yamls contained in `rrfs-test/testinput_expr`? This directory contains our older yamls from before updating the ctests to use more obs types. These yamls are copied into the run directory created by `setup_experiment.sh`, but I do not know if anyone is actually using the yamls (or just running `setup_experiment.sh` solely to create the run directory). If not, it would be best to remove `testinput_expr` to avoid having to keep updating these yamls for developments like this PR.

That is a good point @SamuelDegelia-NOAA. We need to open an issue to start cleaning out and cleaning up older yamls.

guoqing-noaa:
> On a sort of unrelated note, do we know if anyone is using the yamls contained in `rrfs-test/testinput_expr`? This directory contains our older yamls from before updating the ctests to use more obs types. These yamls are copied into the run directory created by `setup_experiment.sh`, but I do not know if anyone is actually using the yamls (or just running `setup_experiment.sh` solely to create the run directory). If not, it would be best to remove `testinput_expr` to avoid having to keep updating these yamls for developments like this PR.

@SamuelDegelia-NOAA The `rrfs-test/testinput_expr` directory contains the single obs test which is a very useful experiment to test lots of functionalities. The directory was renamed to its current form in PR #184.  The directory is used by `link_mpasjedi_expr.sh` and [this wiki](https://github.com/NOAA-EMC/RDASApp/wiki/A-Quick-Start-Guide-for-Running-the-MPASJEDI-2024052700-case-on-Jet-or-Hera). We need to keep this single obs test but we can combine them into the `testinput` directory if appropriate.

SamuelDegelia-NOAA:
@guoqing-noaa It is worth keeping that directory since it contains the single obs test as you mention. Maybe it would be be better to rename `testinput_expr` to something else since "testinput" implies use for the ctests. But that is not a priority right now and I am fine with keeping it the way it is.

PR #242, opened 2024-11-29
Add bufr2ioda and ens3dvar yaml for ZTD obs

HuiLiu-NOAA:
This PR is to add a bufr2ioda yaml and an ens3dvar yaml for ZTD obs.
1. The bufr2ioda yaml converts RAP gpsipw dump hourly Bufr data into the IODA files.
2. The initial en3dvar yaml perform assimilation of ideal or real ZTD observations with MPAS-JEDI using the existing UKMO ZTD local operator.


SamuelDegelia-NOAA:
@HuiLiu-NOAA Instead of adding a new en3dvar yaml in `RDASApp/rrfs-test/testinput`, could you instead just create a new obs-space yaml in `RDASApp/rrfs-test/validated_yamls/templates/obtype_config`? It could just be called something like `ztd.yaml`. And then in that file all you need to add is the section of your original yaml starting with `- obs space: `. You can see the files in that directory for examples. You'll also need to add `distribution: ` and `obs localizations: ` headers so that the ztd yaml can work with EnKF.

SamuelDegelia-NOAA:
Also @HuiLiu-NOAA do you know how the time window is set for the ZTD obs? In our current bufr2ioda yamls we create a `timeOffset` variable and then use a pre-filter in the obs-space yamls such as: 

```
         - filter: Domain Check
           apply at iterations: 0,1
           where:
             - variable:
                 name: MetaData/timeOffset
               minvalue: -0.75
               maxvalue: 0.75
```

HuiLiu-NOAA:
> Also @HuiLiu-NOAA do you know how the time window is set for the ZTD obs? In our current bufr2ioda yamls we create a `timeOffset` variable and then use a pre-filter in the obs-space yamls such as:
> 
> ```
>          - filter: Domain Check
>            apply at iterations: 0,1
>            where:
>              - variable:
>                  name: MetaData/timeOffset
>                minvalue: -0.75
>                maxvalue: 0.75
> ```

The "timeOffset" seems only available in prepbufr data, not in the bufr data? The ZTD bufr data seems does not have the variable.

SamuelDegelia-NOAA:
> The "timeOffset" seems only available in prepbufr data, not in the bufr data? The ZTD bufr data seems does not have the variable.

Yes the `timeOffset` is only in prepbufr. So if the ZTD obs are bufr then this method would not work. What is the time range of the obs in the bufr and IODA files? Is it small enough that we do not need a time window filter?

HuiLiu-NOAA:
> > The "timeOffset" seems only available in prepbufr data, not in the bufr data? The ZTD bufr data seems does not have the variable.
> 
> Yes the `timeOffset` is only in prepbufr. So if the ZTD obs are bufr then this method would not work. What is the time range of the obs in the bufr and IODA files? Is it small enough that we do not need a time window filter?

The time of ZTD is right on the hour of every hourly bufr data. So, time window filter might not be needed?

SamuelDegelia-NOAA:
Okay yeah then we probably do not need a time window filter right now. Thanks for confirming!

HuiLiu-NOAA:
> @HuiLiu-NOAA Instead of adding a new en3dvar yaml in `RDASApp/rrfs-test/testinput`, could you instead just create a new obs-space yaml in `RDASApp/rrfs-test/validated_yamls/templates/obtype_config`? It could just be called something like `ztd.yaml`. And then in that file all you need to add is the section of your original yaml starting with `- obs space: `. You can see the files in that directory for examples. You'll also need to add `distribution: ` and `obs localizations: ` headers so that the ztd yaml can work with EnKF.

Thanks! done as suggested.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
```diff
@@ -0,0 +1,70 @@
...
+       obs error:
+         covariance model: diagonal
+
+       obs localizations:
+         - localization method: Horizontal Gaspari-Cohn
+           lengthscale: 300e3 # orig
+
+       obs filters:
+       - filter: PreQC
+         maxvalue: 10
```
delippi: 1. For the entire `obs filters` section, it would be helpful to have some useful comments before each QC filter. For example, this first one could be like
```yaml
       # Reject all obs with QualityMarker > 10
       - filter: PreQC
         maxvalue: 10
```

The comments add visual breaks and just make it easier for someone to know what is going on. For example, the "online domain check" sometimes uses the `BoundsCheck` which can be confusing. Feel free to look at the mesonet and metar yamls for how that has been done.

2. Is `PreQC` a filter? What is it checking? To my knowledge, `PreQC` has been replaced with `QualityMarker` (although there are a lot of places in JEDI that still use `PreQC`). This section could likely also be done like:
```yaml
         - filter: RejectList
           apply at iterations: 0,1
           where:
           - variable: QualityMarker/zenithTotalDelay
             is_in: 4-15
```
3. I think we typically reject obs with `PreQC`/`QualityMarker` > 3. Is there any reason for setting 10 as the maxvalue? I'll caveat this with: I'm not sure how robust the `QualityMarker` values are as I think those are very crude QC mainly for a time when there wasn't as robust QC checks as there are today.

rrfs-test/IODA/yaml/bufr_ncep_ztd.yaml
```diff
@@ -0,0 +1,159 @@
...
+            transforms:
+              - scale: 1000.0
+
+          # ObsValue
+          zenithTotalDelay:
+            query: "*/GNSSRPSQ{1}/APDS"
+          
+          # QualityMarker 
+          zenithTotalDelayQM:
+            query: "*/QFGN"
```
delippi: I thought QualityMarker was the term meant to replace PreQC? What is the difference here?

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
```diff
@@ -0,0 +1,70 @@
...
+       - filter: Bounds Check
+         filter variables:
+         - name: zenithTotalDelay
+         test variables:
+         - name:  GeoVaLs/observable_domain_mask
+         minvalue: 0.0
+         maxvalue: 0.5   # keep OBS in MPAS domain interior
+
+       - filter: GOMsaver
+         filename: geoval_ztd_single.nc
```
delippi: Since these are mainly for debugging, we have been commenting these lines out in the validated_yamls.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
```diff
@@ -0,0 +1,70 @@
+     - obs space:
+         name: Groundgnss
```
delippi: Could this be named something like:
```yaml
         name: gnss_zenithTotalDelay_123
```

To try to match what is in the mesonet and metar yamls. Does ztd have a observation type number? For example mesonet would be:
```yaml
         name: msonet_airTemperature_188
```

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
```diff
@@ -0,0 +1,70 @@
...
+           halo size: 100e3
+         obsdatain:
+           engine:
+             type: H5File
+             obsfile: single_ztd_obs_2024052700.nc
+#            obsfile: ioda/2024052700.rap.t00z.ztd.tm00.bufr_d.nc
+         obsdataout:
+           engine:
+             type: H5File
+             obsfile: ztd_single_diag_out_2024052700.nc4
```
delippi: This one would need to also be updated with the current norms. This would be something like:
```yaml
             obsfile: jdiag_gnss_zenithTotalDelay_123.nc4
```
or whatever you decide on the `obs space` name.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
```diff
@@ -0,0 +1,70 @@
...
+#-------------------------------------------------------
+#         io pool:
+## to be consistent with the ntask number
+#           max pool size: 120
+#           write multiple files: true
+#-------------------------------------------------------
```
delippi: Do we need this here? Seems like that part might have been removed.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
```diff
@@ -0,0 +1,70 @@
...
+         name: GroundgnssMetOffice
+#        variables:
+#        - name: zenithTotalDelay
+
+         observation alias file: obsop_name_map.yaml
+#------------------------------------------------
+         vertical coordinate: height
+         observation vertical coordinate: height
+         interpolation method: linear
+#or ---------------------------------------------
```
delippi: Why is this enclosed in
```yaml
#------
#or ------
```

Can those be removed if you don't think they are necessary?

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
HuiLiu-NOAA: Thanks for the comments. 
1. Will add these comments. 
2. The PreQC filter exists and works well as described in JEDI doc 8.0, which is similar to GSI. 
3.  The actual QC values are not established yet (still missing) yet in the Bufr dataset. Here only a trial value (0-10) is used temporarily for testing.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
HuiLiu-NOAA: These outputs are for debugging purpose. Commented for now.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
HuiLiu-NOAA: Commented these.

rrfs-test/IODA/yaml/bufr_ncep_ztd.yaml
HuiLiu-NOAA: The PreQC is kept here just for historical reason and backward compatibility since many current yamls still use it.  It is the same as QualityMarke.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
HuiLiu-NOAA: updated.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
HuiLiu-NOAA: updated.

rrfs-test/validated_yamls/templates/obtype_config/ztd_ukmo.yaml
HuiLiu-NOAA: removed "#" and commented out these lines, which are for MPAS-JEDI only.

guoqing-noaa:
@HuiLiu-NOAA Sorry this crash is NOT due to your PR. It is because the latest RDASApp cannot build on Hercules.

guoqing-noaa:
@HuiLiu-NOAA Sorry that I used your PR to test the latest RDASApp on Hercules. The failure is NOT due to this PR.

@SamuelDegelia-NOAA 
Some ctests failed on Hercules for the latest RDASApp.  Do you have time to take a look? Thanks!

SamuelDegelia-NOAA:
@guoqing-noaa Looks like it was a memory issue. We will probably need a PR to change the ctest configuration on Hercules specifically. 

```
DRPCG end of iteration 18
  Residual norm (18) = 39.64704425434255
  Norm reduction (18) = 0.03950584712510658

  Quadratic cost function: J   (18) = 17693.9201842808
  Quadratic cost function: Jb  (18) = 2989.616654909762
  Quadratic cost function: JoJc(18) = 14704.30352937103

 DRPCG Starting Iteration 19
slurmstepd: error: Detected 1 oom_kill event in StepId=4595349.0. Some of the step tasks have been OOM Killed.
srun: error: hercules-02-53: task 20: Out Of Memory
srun: Terminating StepId=4595349.0
slurmstepd: error: *** STEP 4595349.0 ON hercules-02-53 CANCELLED AT 2025-02-19T14:23:12 ***
```

guoqing-noaa:
@SamuelDegelia-NOAA Thanks a lot for the quick response!

HuiLiu-NOAA:
This PR is closed and a new one will be created with updated ZTD yaml and bufr2ioda tool for RRFSv2 workflow.

Issue #258, opened 2025-01-06
Duplicate Checking in RDAS


delippi:
## Description
In GSI, RRFS uses `l_closeobs` in the setup routines to perform duplicate checks. This option retains the observation at a given lat/lon/pressure that is closest to the analysis time. Currently, we have been using `Temporal Thinning` filter in JEDI to perform duplicate checks; however, we have been specifying the category variable (i.e., how the observations are grouped) as `MetaData/stationIdentification` which will not always work. For example, if you have a profiler where there may be many observations at the same lat/lon/stationID but different pressure, only a single observation will be retained from that profile. The category variable must be a string or integer and it can't [currently] be a list of variables. I thought my proposed solution below would be easier than updating the code to be able to use more than one category variable.

**Proposed fix:**
We can add an ad hoc variable such as a string like "longitude_latitude_pressure" to the IODA file to be used as the category variable. This would be best suited to be placed in a python bufr2ioda converter; however, we are currently using yaml based converters. Therefore, I am proposing to add this to the offline_domain check code and we will just run that offline tool for the early stages of development.

## Acceptance Criteria (Definition of Done)
The Temporal Thinning filter can use longitude latitude and pressure (not just station ID) as category variable either by adding the ad hoc variable to the offline domain check, a separate tool, python converter, or updating the Temporal Thinning code.
- Link any relevant pull requests here:
  - PR #
  - PR #

## Dependencies
None

delippi:
Are there any objections to adding the variable to the offline domain check and just using that for early cycling experiments? That is likely the easiest way forward until a more permanent solution can be made.

delippi:
Here is an example for mesonet. I can share more if interested. Really just looking at observation counts.

The old way
![msonet_airTemperature_188](https://github.com/user-attachments/assets/63f03d69-580c-4524-becf-cbc2b9c15d70)

The new way
![msonet_airTemperature_188](https://github.com/user-attachments/assets/700e2f35-d271-4358-936f-aefa9895b9c9)

PR #259, opened 2025-01-09
add the offline tool to add lon_lat_ps to IODA and updated each jedi yaml to use that as category variable

delippi:
## Bug Fix Description

1. The Temporal Thinning filter had previously been configured using `category_variable.name: MetaData/stationIdentification`.
2. It actually isn't clear what that filter was doing and it often produced strange results.

1. A new variable is added to IODA in an offline tool which is provided in this PR. The offline tool is only a temporary solution and the best place to put this will likely be in a python converter. This tool is based on the offline domain check.
2. The new variable `MetaData/longitude_latitude_pressure` is now used as the `category variable` to check for duplicate obs matching the functionality that is in GSI using `l_closeobs` option.

## Issue(s) addressed

- Resolves Issue https://github.com/NOAA-EMC/RDASApp/issues/258

## Additional Info
To run the offline add variable to ioda tool use the command below.
```bash
python ../offline_add_var_to_ioda.py -o $ioda_file
```
The code creates a copy of the IODA with the additional variable. The final ioda will end in `_llp.nc` or `_llp.nc4`.

## Checklist

### DIFF
--- a/rrfs-test/IODA/offline_add_var_to_ioda.py
+++ b/rrfs-test/IODA/offline_add_var_to_ioda.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python
+import netCDF4 as nc
+import numpy as np
+from timeit import default_timer as timer
+import argparse
+import warnings
+import os
+
+"""
+This program makes a copy of an original IODA file and can add additional
+adhoc variables. This program was originally written to add the
+MetaData/longitude_latitude_pressure for use in l_closeobs duplicate checking
+but others can be added as necessary.
+"""
+
+# Disable warnings
+warnings.filterwarnings('ignore')
+
+# Functions for calculating run times.
+def tic():
+    return timer()
+
+def toc(tic=tic, label=""):
+    toc = timer()
+    elapsed = toc-tic
+    hrs = int(elapsed // 3600)
+    mins = int((elapsed % 3600) // 60)
+    secs = int(elapsed % 3600 % 60)
+    print(f"{label}({elapsed:.2f}s), {hrs:02}:{mins:02}:{secs:02}")
+
+tic1 = tic()
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-o', '--obs', type=str, help='ioda observation file', required=True)
+args = parser.parse_args()
+
+# Assign filenames
+obs_filename = args.obs
+
+obs_ds = nc.Dataset(obs_filename, 'r')
+
+# Extract observation latitudes and longitudes
+obs_lat = obs_ds.groups['MetaData'].variables['latitude'][:]
+obs_lon = obs_ds.groups['MetaData'].variables['longitude'][:]
+obs_lon = np.where(obs_lon < 0, obs_lon + 360, obs_lon)
+obs_prs = obs_ds.groups['MetaData'].variables['pressure'][:]
+
+# Create a new NetCDF file to store the selected data using the more efficient method
+try:
+    outfile = obs_filename.replace('.nc', '_llp.nc')
+except:
+    outfile = obs_filename.replace('.nc4', '_llp.nc4')
+fout = nc.Dataset(outfile, 'w')
+
+# Create dimensions and variables in the new file
+fout.createDimension('Location', len(obs_lat))
+fout.createVariable('Location', 'int64', 'Location')
+fout.variables['Location'][:] = 0
+for attr in obs_ds.variables['Location'].ncattrs():  # Attributes for Location variable
+    fout.variables['Location'].setncattr(attr, obs_ds.variables['Location'].getncattr(attr))
+
+# Copy all non-grouped attributes into the new file
+for attr in obs_ds.ncattrs():  # Attributes for the main file
+    fout.setncattr(attr, obs_ds.getncattr(attr))
+
+# Copy all groups and variables into the new file, keeping only the variables in range
+groups = obs_ds.groups
+for group in groups:
+    g = fout.createGroup(group)
+    for var in obs_ds.groups[group].variables:
+        invar = obs_ds.groups[group].variables[var]
+        try:  # Non-string variables
+            vartype = invar.dtype
+            fill = invar.getncattr('_FillValue')
+            g.createVariable(var, vartype, 'Location', fill_value=fill)
+        except:  # String variables
+            g.createVariable(var, 'str', 'Location')
+        g.variables[var][:] = invar[:][:]
+        # Copy attributes for this variable
+        for attr in invar.ncattrs():
+            if '_FillValue' in attr: continue
+            g.variables[var].setncattr(attr, invar.getncattr(attr))
+
+# Generate longitude_latitude_pressure location strings (for dup checking)
+longitude_latitude_pressure = [f"{lon}_{lat}_{pres}" for lon, lat, pres in zip(obs_lon, obs_lat, obs_prs)]
+longitude_latitude_pressure = np.array(longitude_latitude_pressure)
+
+# Add the longitude_latitude_pressure variable to the file
+var = "longitude_latitude_pressure"
+data = longitude_latitude_pressure
+metadata_group = fout.groups['MetaData']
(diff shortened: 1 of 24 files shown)


SamuelDegelia-NOAA:
Thanks for adding this tool @delippi. Did you recreate the obs files with the new variable so that we can use these updated yamls? If so we should go ahead and stage those and then relink them under `fix/expr_data`.

SamuelDegelia-NOAA (approved):

ShunLiu-NOAA:
@guoqing-noaa or @spanNOAA If you have a chance, please review this PR. Thanks.

guoqing-noaa:
I have three comments:

1. Does this tool include the online domain check?

2. METAR and MSONET sometimes report slightly different pressures for close observations. In GSI, we only choose the close one based on lat/lon and don't consider the pressure differences. Although there are different views on how to handle this, I think the current GSI method is acceptable.

3. The wind profiler data is different, we do want to consider pressure differences when choosing the closet observations. 

So I would think as this moment, it is a safe bet to apply the change to wind profiler data only.

SamuelDegelia-NOAA:
@guoqing-noaa Based on the code [here](https://github.com/NOAA-EMC/GSI/blob/27c03e83a66fa2ba3b32781dab7cd6a8c0fc497d/src/gsi/setupt.f90#L357) and [here](https://github.com/NOAA-EMC/GSI/blob/27c03e83a66fa2ba3b32781dab7cd6a8c0fc497d/src/gsi/setupt.f90#L453-L464), it looks like the pressure is always considered for duplicate surface obs when `i_use_2mt4b > 0` in GSI (no matter the obtype). I'm not sure if this is the intended behavior, though.

guoqing-noaa:
> @guoqing-noaa Based on the code [here](https://github.com/NOAA-EMC/GSI/blob/27c03e83a66fa2ba3b32781dab7cd6a8c0fc497d/src/gsi/setupt.f90#L357) and [here](https://github.com/NOAA-EMC/GSI/blob/27c03e83a66fa2ba3b32781dab7cd6a8c0fc497d/src/gsi/setupt.f90#L453-L464), it looks like the pressure is always considered for duplicate surface obs when `i_use_2mt4b > 0` in GSI (no matter the obtype). I'm not sure if this is the intended behavior, though.

@SamuelDegelia-NOAA You are right about that part. But for selecting the closest obs, theoretically, we only need to pick the closest one for METAR and MESONET despite pressure differences.

guoqing-noaa (changes requested):
I think we need more discussions on this.

For METAR, MSONET, we may not need to consider pressure differences.

For aircraft,  sounding, I am not sure whether we need the "l_closest_obs" option. It will be good to double-check this.

For wind profilers, it will be good to check how many profilers are available in the current operational observation stream and what's their observing frequencies.

Thanks!

SamuelDegelia-NOAA:
@guoqing-noaa I guess the question then is at this stage are we trying to emulate what GSI does for METAR, MSONET (considers pressure for duplicate obs)? Or are we okay making some changes if we think they are more reasonable.

guoqing-noaa:
I understand this tool is trying to mimic the "l_closet_obs" option (using temporal thinning) instead of avoiding duplicate obs, right?

delippi:
> I have three comments:
> 
> 1. Does this tool include the online domain check?

No. I made a completely different offline tool just meant for adding extra variables. It seemed best to keep it separate rather than overloading the domain check tool. It just modeled this tool off of the offline domain check tool, so there are a lot of similarities between them.

> 2. METAR and MSONET sometimes report slightly different pressures for close observations. In GSI, we only choose the close one based on lat/lon and don't consider the pressure differences. Although there are different views on how to handle this, I think the current GSI method is acceptable.

Thanks @SamuelDegelia-NOAA for answering this. I want to add one more point to this and @spanNOAA would probably agree that how we were currently using this Temporal Thinning filter was giving very strange results. I think grouping via stationIdentification was (part of) the problem. Let me give a few examples (and feel free to try playing around with these on your own). For simplicity all these will be with msonet data.
**example 1:**
This is what I think is the "most correct" way to specify this filter. In plain language, as configured below, this will group observations that match exactly with the category_variable.MetaData/longitude_latitude_pressure and only keep the closest observation to the seed_time.2024-05-27T00:00:00Z within a 90 minute period.  The result is
nobs: 30693 (GSI.nobs: 30661)
```
         # Duplicate Check
         - filter: Temporal Thinning
           apply at iterations: 0,1
           min_spacing: PT90M
           seed_time: "2024-05-27T00:00:00Z"
           category_variable:
             name: MetaData/longitude_latitude_pressure
```
**example 2:**
The only difference here is using category_variable.MetaData/stationIdentification. What we had previously done to get nobs counts close was specify min_spacing to something like PT00M - PT30M. But that doesn't make sense because it would potentially keep duplicate observations that are only spaced 0M - 30M. min_spacing.PT90M is the correct number to use and when specified like below the result is
nobs: 14130
```
         # Duplicate Check
         - filter: Temporal Thinning
           apply at iterations: 0,1
           min_spacing: PT90M
           seed_time: "2024-05-27T00:00:00Z"
           category_variable:
             name: MetaData/stationIdentification
```
**example 3:**
In this example, this duplicate check is commented out in the yaml because I don't trust that min_spacing.PT00M was the same as turning off this filter. The result is
nobs: 61923

**example 4 & 5:**
Using min_spacing.PT00M in experiments grouping via MetaData/longitude_latitude_pressure and MetaData/stationIdentification respectively gives the following results
nobs: 61919
nobs: 61557

nobs for the new grouping was still off by a few counts, compared to example 3, but much closer when grouped with the new category variable.

> 3. The wind profiler data is different, we do want to consider pressure differences when choosing the closet observations.
> 
> So I would think as this moment, it is a safe bet to apply the change to wind profiler data only.

Yes, we definitely want this for profiler data. Working on this data is how I came to this solution and then tested with other data types. But I'm confident that we also need to consider this for other data types including soundings. Let me give another two examples:
**example 1:**
This is what I think is "most correct". This gives the result
nobs: 21444 (GSI.nobs: 21444)
```
         # Duplicate Check
         - filter: Temporal Thinning
           apply at iterations: 0,1
           min_spacing: PT90M
           seed_time: "2024-05-27T00:00:00Z"
           category_variable:
             name: MetaData/longitude_latitude_pressure
```
**example 2:**
The only difference here is changing the category variable back to stationIdentification. In plain language, as configured below, this filter will group observations that match exactly  category_variable.MetaData/stationIdentification and only keep the single closest observation (from the entire profile) to the seed_time.2024-05-27T00:00:00Z within a 90 minute period.  The result is
nobs: 70 (I checked, there are only 70 stations)
```
         # Duplicate Check
         - filter: Temporal Thinning
           apply at iterations: 0,1
           min_spacing: PT90M
           seed_time: "2024-05-27T00:00:00Z"
           category_variable:
             name: MetaData/stationIdentification
```

delippi:
> I understand this tool is trying to mimic the "l_closet_obs" option (using temporal thinning) instead of avoiding duplicate obs, right?

Yes. I think it is best while still in "validation mode" that we first seek to emulate capabilities in GSI. Then later if there is something we want to improve, we will have the baseline already configured.

Edit: in general

ShunLiu-NOAA:
@guoqing-noaa could you please review this again? Let us know if we need further discussion.

guoqing-noaa (approved):
Thanks for the discussion today!

delippi:
Thanks @guoqing-noaa for approving! Discussion is always welcome!

Just to recap so it is documented here: We plan to proceed with the PR as-is because it brings the use of observations to how it is currently being done in GSI (the baseline). That might not be how we end up implementing in v2 as there are known issues with GSI (e.g., too many surface obs retained due to slight differences in surface pressure). That requires a science change which we will do in the future after we establish our baseline. Those changes are also likely very easy because it might just be search and replace `longitude_latitude_pressure` with `stationIdentitifaction` for surface observations (effectively grouping by longitude_latitude). The PR as-is is what we want for most (if not all) other ob types.

PR #262, opened 2025-01-22
Plot increments at pressure levels in addition to model levels

keenaneure:
Traditionally, mpasjedi_increment_fulldom.py, fv3jedi_increment_fulldom.py and other scripts plot increments based on only a model level. This update requires the user to specify a model level OR a pressure level. When a pressure level is specified, the script does a 1d linear interpolation.

This will be helpful for needing increments at specific pressure levels as well as comparisons across FV3-JEDI and MPAS-JEDI tests, in which model levels are defined differently in the output.

Here, I also attach an example of plotting temperature increments for a GOES AMV test at both model level 24 in MPAS, which was approximately close to 500 hPa based on checks of pressure values at all model levels. These comparisons are not expected to be identical, but they should be relatively close.

model level
pressure level

As of now, mpasjedi_increment_fulldom.py and fv3jedi_increment_fulldom.py are modified.

> **NOTE:** Please go to the `Preview` tab and select the appropriate sub-template.

* [Feature](?expand=1&template=feature_pr.md)
* [Bugfix](?expand=1&template=bugfix_pr.md)
* [Sync Submodules](?expand=1&template=sync_submodule_pr.md)


delippi:
I think this is a good idea to have this capability. My only thought is that we might be "overloading" these scripts. Would there be any objection to splitting this into two different scripts? So keep the original scripts plotting on model levels and add two more to do plotting on pressure levels.

Let's see what other's think about this.

guoqing-noaa:
> I think this is a good idea to have this capability. My only thought is that we might be "overloading" these scripts. Would there be any objection to splitting this into two different scripts? So keep the original scripts plotting on model levels and add two more to do plotting on pressure levels.
> 
> Let's see what other's think about this.

Either way would work for me. But if it is combined, we should keep the vertical levels as the default options (i.e. keep what is in current repo as the default and one can make edits to plot pressure levels).

keenaneure:
I would be totally fine creating two separate scripts. Especially with the design now, it will not work if BOTH a model and pressure level are chosen.

SamuelDegelia-NOAA:
Thank you @keenaneure for adding this functionality! It will be great to validate at a constant pressure surface between FV3 and MPAS. 

I am fine with splitting this up into multiple utilities. It would keep the script cleaner (though at the expense of having to make any changes in both scripts). Perhaps we could have `fv3jedi_increment_fulldom.py` and `fv3jedi_increment_fulldom_pres.py`. 

The scripts seem to work well though! I was a little worried about looping over the individual grid points for the interpolation but it actually runs really fast. I have a few very minor comments below.

rrfs-test/ush/fv3jedi_increment_fulldom.py
```diff
@@ -35,16 +39,18 @@
...
+#jbackgrnd = f"{datapath}/letkf-meanprior-fv3_lam-C775.fv_core.res.nc"
 
 # FOR HYBRID (or ENVAR)
-#janalysis = f"{datapath}/hybens3dvar-fv3_lam-C775.fv_core.res.nc" #Ens3dvar-fv3_lam-C775.fv_core.res.nc"
-#jbackgrnd = f"{datapath}/Data/bkg/fv3_dynvars.nc"
-
+janalysis = f"./ens3dvar-fv3_lam-C775.fv_core.res.nc"
+jbackgrnd = f"./Data/bkg/20240527.000000.fv_core.res.tile1.nc"
+#added for pressure calculations
+jbackgrnd2 = f"./Data/bkg/20240527.000000.fv_core.res.nc"
```
SamuelDegelia-NOAA: Could you add back in the `{datapath}` variable to the definitions of `janalysis`, `jbackgrnd`, `jbackgrnd2`, and `jgrid` at the top of the script? That would be helpful if we want to run the script in another location.

rrfs-test/ush/fv3jedi_increment_fulldom.py
```diff
@@ -35,16 +39,18 @@
 
 # JEDI data
 datapath = "./"
-jgrid = f"{datapath}/Data/bkg/fv3_grid_spec.nc"
+#jgrid = f"{datapath}/Data/bkg/fv3_grid_spec.nc"
+jgrid = f"./fv3_grid_spec.nc"
```
SamuelDegelia-NOAA: For `jgrid`, I do not have a `fv3_grid_spec.nc` file in `expr/fv3_2024052700` by default. Could we change it to `{datapath}/Data/bkg/grid_spec.nc`?

rrfs-test/ush/fv3jedi_increment_fulldom.py
```diff
@@ -17,12 +17,16 @@
...
+from scipy.interpolate import interp1d
 
 warnings.filterwarnings('ignore')
 
 ############ USER INPUT ##########################################################
 plot_var = "Increment"
-lev = 60                # 60=sfc; 1=toa
+##choose lev OR pres_lev by commenting/uncommenting out which is desired
+##code will break otherwise
+#lev = 35                # 60=sfc; 1=toa
```
SamuelDegelia-NOAA: I agree with keeping `lev` as the default in this main script and then using `lev_pres` in a corresponding pressure script.

rrfs-test/ush/mpasjedi_increment_fulldom.py
```diff
@@ -137,4 +166,4 @@ def plot_T_inc(var_n, clevmax):
 # Print some final stats
 print(f"Stats:")
 print(f" {longname} max: {np.around(np.max(jedi_inc), decimals)}")
-print(f" {longname} min: {np.around(np.min(jedi_inc), decimals)}")
```
SamuelDegelia-NOAA: Could we add this back in? It is useful to see the min/max stats printed at runtime just to make sure the increments are reasonable.

keenaneure:
Hi all,

Here's a short summary of the changes I made. At the moment, I am still having trouble pushing these updates:

1) mpasjedi_increment_fulldom_pres.py and fv3jedi_increment_fulldom_pres.py are now their own scripts, in the same ush/ directory. I got rid of the if statements later on in each script based on whether 'lev' or 'pres_lev' was provided.

2) The pathways for jgrid, janalysis, jbackgrnd, and jbackgrnd2 are changed to include {datapath}, while jgrid's filename is changed to grid_spec.nc instead of fv3_grid_spec.nc. For what it is worth, I actually did not see {datapath} in my recent-ish clone of RDASApp for janalysis and jbackgrnd (not including jbackgrnd2 since I added that one with this PR).

SamuelDegelia-NOAA (approved):
LGTM, thanks for adding this utility!

delippi (approved):
Looks good. Thanks.

ShunLiu-NOAA (approved):

Issue #270, opened 2025-02-04
Incorporate proflr DA into JEDI for RRFS


delippi:
## Description
The UFO in JEDI is the component that not only computes model-simulated observations, but also houses filters and methods for observation QC, ob errors, and bias correction. The GSI observer is the equivalent. Many forward operators for various observations have been developed for the UFO. These operators can be utilized in RDAS. However these operators still must be tested for RDAS. The steps for transition by observation platform are as follows:

1. Establish IODA processing
2. Establish operator in UFO
3. Establish YAML input file
4. Validate with test assimilation experiments
5. Create ctest, if necessary
6. proflr data is bufr obtype=227

## Requirements
To create yaml configurations for assimilating profiler data.

## Acceptance Criteria (Definition of Done)
- [x] Phase 1 validation (hofx validation using GSI-IODA and GSI-geovals as necessary; GSI vs FV3-JEDI)
- [ ] Phase 2 validation (QC validation; no reliance on GSI except to be used as the baseline; GSI vs FV3-JEDI)
- [ ] Phase 3 validation (FV3-JEDI and GSI vs MPAS-JEDI)

- Link any relevant pull requests here:
  - 

## Dependencies
None

delippi:
**Phase 1 update:**
These results were created using GSI for IODA, geovals, QC, and final ob error which passes the standard for phase 1 testing.

![Image](https://github.com/user-attachments/assets/438b6317-e0d8-4116-8bff-55660921ec0c)
![Image](https://github.com/user-attachments/assets/f00829c9-7abf-4edd-b1ea-fbba89d97993)

Here are the results after using only GSI-IODA and allowing JEDI to handle the rest (geovals, QC, and oberror inflation). This results also used ObsErrorFactorConventional (OEFC) which seems to play a role in the differences we are seeing here in final ob errors. I think the differences in hofx are simply due to differences in geovals calculations since in the previous plot we saw an exact one-to-one match using GSI-IODA and GSI-geovals.

![Image](https://github.com/user-attachments/assets/f125cb13-93ac-4070-bd67-e7581bb6c285)
![Image](https://github.com/user-attachments/assets/278796c3-3650-4c51-b4b9-a9e83110079a)

delippi:
**Phase 2/3 update**
[...]

PR #271, opened 2025-02-04
Phase 1 validation: proflr (227) yaml

delippi:
## Description
Proflr yaml passed phase 1 validation. There were issues with ObsErrorFactorConventional (OEFC) filter for this data similar to most other data types, but results were not bad.

## Issue(s) addressed
- https://github.com/NOAA-EMC/RDASApp/issues/270

## Checklist

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/proflr_winds_227.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/proflr_winds_227.yaml
@@ -36,6 +36,9 @@
            - name: windEastward
            - name: windNorthward
 
+       linear obs operator:
+         name: VertInterp
+
        obs error:
          covariance model: diagonal
 
@@ -86,7 +89,7 @@
            action:
              name: reduce obs space
 
-         # Online domain check
+         # Online regional domain check
          - filter: Bounds Check
            filter variables:
            - name: windEastward
--- a/rrfs-test/validated_yamls/templates/obtype_config/proflr_winds_227.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/proflr_winds_227.yaml
@@ -1,0 +1,359 @@
+     - obs space:
+         name: proflr_winds_227
+         distribution:
+           name: "@DISTRIBUTION@"
+           halo size: 100e3
+         obsdatain:
+           engine:
+             type: H5File
+             obsfile: "@OBSFILE@"
+             #obsfile: "obsout/windprof_obs_2024052700.nc4"
+             #obsfile: "obsout/windprof_obs_2024052700_dc.nc4"
+           obsgrouping:
+             group variables: ["stationIdentification"]
+             sort variable: "pressure"
+             sort order: "descending"
+         obsdataout:
+           engine:
+             type: H5File
+             obsfile: jdiag_proflr_winds_227.nc4
+             allow overwrite: true
+         io pool:
+           max pool size: 1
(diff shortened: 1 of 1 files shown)


HuiLiu-NOAA:
one minor question: 
  why the  linear obs operator uses "Identity", rather than the "VertInterp" operator?

also, could the next 2 lines removed from the obs operator?
        name: Composite
         components:

delippi:
> one minor question: why the linear obs operator uses "Identity", rather than the "VertInterp" operator?
> 
> also, could the next 2 lines removed from the obs operator? name: Composite components:

@HuiLiu-NOAA, good questions. I was trying to recall why I needed the linear obs operator in the yaml, and I can't find any other examples that use it. In fact, I just tried taking it out and it didn't affect hofx results. It did oddly affect final analysis result though (in a slightly better way). I think I may just take this out of this and each of the yamls that we have that use it. I need to do a little more testing for other ob types just to be certain. As for the Composite and components lines, yes I think that can be removed as long as I also remove the "-" before name: VertInterp as well. I think I till also do this too for this and each of the yamls that we currently have as well. These changes will just make the yamls a little more clean. Thanks for these suggestions.

HuiLiu-NOAA:
Thanks @delippi.  
My understanding is that for actual assimilation with En3DVar etc, the linear operator is needed and better set to "VertInterp" to be consistent to the nonlinear operator, which would be more accurate and optimal. If only for HofX, that does not matter as you mentioned.

Another thought (may be too picky): 
Would it be more clear to add "regional" to the next comment line, since the "domain" in JEDI may have other meanings :
#Online regional domain check

SamuelDegelia-NOAA:
Thanks @HuiLiu-NOAA for catching that. We use the linear operator in LETKF/GETKF so setting it to VertInterp would be best

delippi:
> Thanks @delippi. My understanding is that for actual assimilation with En3DVar etc, the linear operator is needed and better set to "VertInterp" to be consistent to the nonlinear operator, which would be more accurate and optimal. If only for HofX, that does not matter as you mentioned.
> 
> Another thought (may be too picky): Would it be more clear to add "regional" to the next comment line, since the "domain" in JEDI may have other meanings : #Online regional domain check

Hmm, I also tested it with VertInterp and it didn't change results in my current configuration. I will add it back in and change it to VertInterp based on your findings and since it didn't change my results. 

I have no problem updating the comment. The only problem is I would want to update that comment in all the yamls too for consistency. I will eventually need to go back through all the yamls that we've built up and do this any way for several items.

ShunLiu-NOAA (approved):

ShunLiu-NOAA:
@guoqing-noaa or @spanNOAA do you have a change to review this PR?

HuiLiu-NOAA:
@delippi : Has this yaml been tested with MPAS-JEDI with the pressure coordinates listed (MPAS is basically height coordinate)?

delippi:
> @delippi : Has this yaml been tested with MPAS-JEDI with the pressure coordinates listed (MPAS is basically height coordinate)?

@HuiLiu-NOAA, no not yet. Phase 1 means FV3-GSI vs FV3-JEDI both using GSI-IODA. However, I have been using pressure coordinates in my other Phase 3 (MPAS) testing and it works. I will remember to test height coordinate in future tests to see what the difference may be.

ShunLiu-NOAA:
@guoqing-noaa, @spanNOAA and @chunhuazhou can someone from GSL review this PR?

PR #272, opened 2025-02-04
Add `reduce obs space` filter action to speed up solvers

SamuelDegelia-NOAA:
## Description
This PR adds the filter action `reduce obs space` to pre filters for all observation yamls in RDASApp. This filter action removes observations from memory if they would be rejected by the prefilters. This saves computation time on hofx calculations and in the solver. Here, this action is added to the Bounds Check, Domain Check (really our time window check), Temporal Thinning, and RejectList prefilters.

I should note that `reduce obs space` did not appear to work with any of the satellite radiance filters. I assume those are all considered post-filters.  

This filter action significantly speeds up the runtime for the solvers (especially for GETKF). Here are the ctest runtimes before this change: 

```
(eva) [Samuel.Degelia@hfe11 ush]$ ./print_ctest_runtime.py
#1: rrfs_fv3jedi_2024052700_Ens3Dvar = 00:05:52
#2: rrfs_fv3jedi_2024052700_getkf_observer = 00:05:21
#3: rrfs_fv3jedi_2024052700_getkf_solver = 00:10:48
#4: rrfs_mpasjedi_2024052700_Ens3Dvar = 00:08:06
#5: rrfs_mpasjedi_2024052700_getkf_observer = 00:05:06
#6: rrfs_mpasjedi_2024052700_getkf_solver = 00:18:03
#7: rrfs_mpasjedi_2024052700_bumploc = 00:05:15
#8: rrfs_bufr2ioda_msonet = 00:00:34
```

And here are the runtimes after the change: 

```
(eva) [Samuel.Degelia@hfe11 ush]$ ./print_ctest_runtime.py
#1: rrfs_fv3jedi_2024052700_Ens3Dvar = 00:04:49
#2: rrfs_fv3jedi_2024052700_getkf_observer = 00:02:13
#3: rrfs_fv3jedi_2024052700_getkf_solver = 00:03:11
#4: rrfs_mpasjedi_2024052700_Ens3Dvar = 00:06:11
#5: rrfs_mpasjedi_2024052700_getkf_observer = 00:02:33
#6: rrfs_mpasjedi_2024052700_getkf_solver = 00:05:56
#7: rrfs_mpasjedi_2024052700_bumploc = 00:05:23
#8: rrfs_bufr2ioda_msonet = 00:00:33
```

The Ens3Dvar ctest runs 1-2 minutes faster, and the GETKF solver now runs 2-3x faster. Overall these are great improvements. 

This PR also increases the halo radius the GETKF solver from 100 to 300 km. The earlier value was found to be too low and did not include all obs within the localization radius for grid points near the boundaries of each PE.  

## Issue(s) addressed
Partly addresses #122 

None

## Checklist


SamuelDegelia-NOAA:
Ctest results from Hera: 

```
[Samuel.Degelia@hfe01 rrfs-test]$ ctest -j8
Test project /scratch2/BMC/zrtrr/Samuel.Degelia/RDASApp_reduceobspace/RDASApp/build/rrfs-test
1/8 Test #8: rrfs_bufr2ioda_msonet .....................   Passed   33.76 sec
2/8 Test #2: rrfs_fv3jedi_2024052700_getkf_observer ....   Passed  133.41 sec
3/8 Test #5: rrfs_mpasjedi_2024052700_getkf_observer ...   Passed  153.18 sec
4/8 Test #1: rrfs_fv3jedi_2024052700_Ens3Dvar ..........   Passed  289.66 sec
5/8 Test #7: rrfs_mpasjedi_2024052700_bumploc ..........   Passed  323.19 sec
6/8 Test #3: rrfs_fv3jedi_2024052700_getkf_solver ......   Passed  191.70 sec
7/8 Test #4: rrfs_mpasjedi_2024052700_Ens3Dvar .........   Passed  371.62 sec
8/8 Test #6: rrfs_mpasjedi_2024052700_getkf_solver .....   Passed  356.82 sec

Label Time Summary:
mpi            = 1853.33 sec*proc (8 tests)
rdas-bundle    = 1853.33 sec*proc (8 tests)
script         = 1853.33 sec*proc (8 tests)

Total Test time (real) = 510.03 sec
```

delippi (dismissed):
Thanks for updating this Sam. I've been hoping this was coming soon. Looks like a lot of repetitive updates to each yaml. Depending on who's PR is merged first, there will likely be merge conflicts since I moved the Temporal Thinning filter toward the top of the yamls--just be aware. I would still advocate for building the large ctest yamls on the fly instead of having to maintain them, but it is fine. Other than that it looks good.

guoqing-noaa:
>I should note that reduce obs space did not appear to work with any of the satellite radiance filters. I assume those are all considered post-filters.

Does this mean we will keep all radiance data in memory? Why are prefilters regarded as postfilters for radiance data?

guoqing-noaa (dismissed):

SamuelDegelia-NOAA:
> > I should note that reduce obs space did not appear to work with any of the satellite radiance filters. I assume those are all considered post-filters.
> 
> Does this mean we will keep all radiance data in memory? Why are prefilters regarded as postfilters for radiance data?

It looks like our radiance yamls (e.g., [ATMS](https://github.com/NOAA-EMC/RDASApp/blob/develop/rrfs-test/validated_yamls/templates/obtype_config/atms_npp_qc_bc.yaml)) include split sections for pre-filters and post-filters. Adding `reduce obs space` to any filters listed under the `obs post filters:` header throws an error. However, some of these filters are actually pre-filters that can likely take advantage of `reduce obs space`. For example, the Domain check to remove obs from the edge of the scan, and the Gaussian thinning.  

@xyzemc Do you know why we split these yamls into `obs pre filters` and `obs post filters` sections? Would there be any harm in just using a general `obs filters` section like we do for the conventional obs types?

delippi:
@SamuelDegelia-NOAA, quick question, lets consider `ScanEdgeRemoval` for example. Did you add the `reduce obs space` around line 72 or line 283?

Edit: in the example ATMS yamls you just shared

SamuelDegelia-NOAA:
@delippi I originally tried adding it at L283 which caused the error because that filter is listed under obs post filters. I just tried adding it to L72 but it did not seem to have any impact (same hofx file as without `reduce obs space`). I also tried this with the Gaussian thinning filter and it had the same effect. 

I am wondering if we can just use a general `obs filters:` section instead of both `obs pre filters: ` and `obs post filters:`. I am testing that now.

HuiLiu-NOAA:
@SamuelDegelia-NOAA : It seems that @XuLu-NOAA had the "reduce obs space" function working with his radiance DA tests.  Might check with him?

rrfs-test/testinput/rrfs_fv3jedi_2024052700_Ens3Dvar.yaml
delippi, line 1: Did you want to update each of these large yamls with the thinning filter updates. There were two main things 1) the location in the yaml moved; instead of being at the end of each obsspace it is now 3rd-4th filter to be used, 2) use of lon_lat_pres grouping variable instead of stationID.

This will be very tedious. You'll also have to change ctest results. I still think we should be building these on-the-fly instead of maintaining them. The whole spirit of ctests would be making any changes such as to these yamls with new results should be flagged by a failing ctest, updating those references, and moving on. As of now, it seems the spirit we're embracing is that it is okay to use slightly misconfigured yamls in ctest and no need to run ctests for updates to yamls because they're hard coded.

delippi (dismissed):

XuLu-NOAA:
> @delippi I originally tried adding it at L283 which caused the error because that filter is listed under obs post filters. I just tried adding it to L72 but it did not seem to have any impact (same hofx file as without `reduce obs space`). I also tried this with the Gaussian thinning filter and it had the same effect.
> 
> I am wondering if we can just use a general `obs filters:` section instead of both `obs pre filters: ` and `obs post filters:`. I am testing that now.

Hey, Sam, I'm currently using three filters in the same yaml. It works fine for me with ATMS at the reduce obs space. In fact, I put everything under reduce obs space. An example is here on Hera: /scratch2/NCEPDEV/hwrf/scrub/Xu.Lu/JEDI_Inter/analysis_60_10/jedi.yaml
I don't think you can simply use obs filters to replace all those sections as I tried. Certain filters only works under post/pre/prior sections specifically. Also, in the satellite radiance situation, some post filters will reply on the variable transformed/defined from the prior filters.

XuLu-NOAA:
> > @delippi I originally tried adding it at L283 which caused the error because that filter is listed under obs post filters. I just tried adding it to L72 but it did not seem to have any impact (same hofx file as without `reduce obs space`). I also tried this with the Gaussian thinning filter and it had the same effect.
> > I am wondering if we can just use a general `obs filters:` section instead of both `obs pre filters: ` and `obs post filters:`. I am testing that now.
> 
> Hey, Sam, I'm currently using three filters in the same yaml. It works fine for me with ATMS at the reduce obs space. In fact, I put everything under reduce obs space. An example is here on Hera: /scratch2/NCEPDEV/hwrf/scrub/Xu.Lu/JEDI_Inter/analysis_60_10/jedi.yaml I don't think you can simply use obs filters to replace all those sections as I tried. Certain filters only works under post/pre/prior sections specifically. Also, in the satellite radiance situation, some post filters will reply on the variable transformed/defined from the prior filters.

It looks I only used the reduce obs space at the beginning of each data type for the domain check with lat/lon. Is it necessary to repeat it for every filter? Shouldn't it be already reduced after the pre filters? What is the issue you saw?

rrfs-test/testinput/rrfs_fv3jedi_2024052700_Ens3Dvar.yaml
SamuelDegelia-NOAA: You are right that there are lots of differences now between the ctest yamls in `testinput` and those added to `validated_yamls`. I think maybe we should set aside some time for this tomorrow to discuss whether we should build the ctests on the fly. The main downsides are that the ctests will have to be run and updated each time that `validated_yamls` are updated. But if others are willing to do that extra work then it would not be hard to make this change.

SamuelDegelia-NOAA:
@XuLu-NOAA Thank you for helping out here. The domain check developed for MPAS-JEDI is a post filter right now since it relies on Geovals. So we cannot use reduce obs space there. Instead I would primarily like to use reduce obs space with the Gaussian Thinning filter since that removes a decent number of obs. 

You are correct that simply using one big `obs filters` section did not work. I then tried moving just the Gaussian Thinning filter into `obs pre filters` but that changed the increment results. So it seems that the order of the filters here is important. 

I think it worth discussing with @xyzemc whether it is okay to move any of these filters into pre-filters. So we can hold off on merging this PR until she returns from travel.

SamuelDegelia-NOAA:
After some discussion with @xyzemc, we have decided not to implement `reduce obs space` for the polar-orbiting satellites at this time. Shifting the Gaussian Thinning filter into pre-filters changes the order at which these filters are applied and actually removes some obs near the edge of the scan/domain. Those results are [here](https://docs.google.com/presentation/d/1PS3vW9Lqu1dxvueROBMiREDzJAG2n5-tvVh9faTtcug/edit?usp=sharing).

So this PR should be good to continue moving forward as is.

HuiLiu-NOAA:
Thanks @SamuelDegelia-NOAA for the careful comparison.
Actually, I like the results from moving the gaussian thinning to pre-filter, which shows more effective impact of the "thinning" of OBS at the border location.  For the original order, the thinning filter seem showing little impact , that is, the OBS density remains very high at that particular location in the original order.

SamuelDegelia-NOAA:
That is a good point @HuiLiu-NOAA. I was looking at it from the context of removing obs, but if the thinning as a pre-filter works more as intended then maybe it is okay. We should discuss this further tomorrow.

delippi (approved):

PR #280, opened 2025-02-13
Several consistency fixes for all the conventional yamls

delippi:
## Description
This PR fixes several consistency issues between all current conventional validated yamls. The files will now be more easily comparable when opening side by side (e.g. using `vi -d adpsfc_airTemperature_187.yaml msonet_airTemperature_188.yaml`). New yamls are often modeled from previously built yamls, so this keeps new development clean and consistent. This PR also adds some filters that were previously missing from the earliest developed yamls in addition to slight reorganization of some filters. These changes do affect final results. The current baseline was run for every single case and a test was rerun for each updated yaml showing improvements across the board.

Summary of changes:

- Added 1.2x inflation from read_prepbufr to all conventional obs at QM == 3 and specified pressure level.
- Added ObsErrorFactorConventional for all obs but commented out for most. There are still issues for some obs types. Other obs types seem okay (the ones where it is turned on). Ultimately, it is supposed to be on for every type so just establishing the option in the yaml now.
- Updates to some comments
- Linear obs operator to match that used by the non-linear operator OR added when not explicitly defined.
- Filter reorganization
- Various other consistency items to make vi -d happy

## Issue(s) addressed
No issue

None

## Checklist


delippi:
Just for reference, here is a link to download a zip file with all the plots confirming no issues and showing improvements for each type. Note there is no aircft*35 obs, so there is not plot for that.
[yaml_consistency.zip](https://github.com/user-attachments/files/18787642/yaml_consistency.zip)

SamuelDegelia-NOAA (approved):
Looks good to me, results seem either positive or neutral. Just curious, does adding the `obsgrouping` section do anything for surface obs or is that mainly for consistency with other obs types?

delippi:
> Looks good to me, results seem either positive or neutral. Just curious, does adding the `obsgrouping` section do anything for surface obs or is that mainly for consistency with other obs types?

Well, mainly for consistency right now. The obsgrouping at the top is actually required for the ObsErrorFactorConventional (OEFC) function, but the OEFC is commented out for most of the yamls right now. OEFC will run without the obsgrouping, but I believe it will do nothing, so it is good to have it in there now. I think there is still a bug in the OEFC that I need to better understand and fix before I can turn it back on for the remainder to the obtypes.

SamuelDegelia-NOAA:
Ahh gotcha, I did not know that was required for OEFC. Thanks!

delippi:
@guoqing-noaa are there any questions about this on GSL's side?

hu5970 (approved):
Let's merge this PR and move forward based on those updated YMAL files.

delippi:
@chunhuazhou here is an example of the improvement I was talking about at today's meeting. Again, the aircar yaml being used in the rrfs-workflow doesn't have any QC filters.
Baseline
aircar_winds_233

Updated
aircar_winds_233

chunhuazhou:
@delippi Thank you for sharing these plots! Very big difference in the increments. I will see what I can do in the real-time runs. Thanks again!

guoqing-noaa:
@delippi Thanks for the comparison figures. 
I will test @HuiLiu-NOAA's changes. If the geoval related error is fixed, we will add all validated QC filters to the workflow.

PR #284, opened 2025-02-18
adjust observer yaml files for rrfs-workflow and downstream utilities

guoqing-noaa:
~~Use both short names and long names in the `obs space.name`. The short names are used as keywords in the downstream applications.~~
~~use short jdiag file names to match rrfs-workflow (downstream applications will also look for short jdiag file names)~~

**Use not-too-long and not-too-short names for `obs space.name` and `jdiag` file names**


delippi:
- I can probably live with the short and long names for obs space, but can you help me understand the exact use case for this? Is it just for the yaml parsing tools and to not have to type the long name fully? Why does one need to use those yaml tools in the first place? What information are you trying to get from those tools that you couldn't get (probably easier) by opening up the file?

- What downstream applications are going to be looking for the short jdiag file names? Couldn't they just use the names as they are now? All of my scripts right now assume the jdiag has the same name convention as the jedi yaml itself (e.g., `jdiag_${yamlname}.nc4`). I find that much easier. Plus, I feel more verbose names is the way of the JEDI and is just more descriptive for all users.

- What is the NCO standard for the observation file names? I was thinking they needed to have a date string appended to the filename. If that is the case, we would need to keep the `@OBSFILE@` placeholder.

guoqing-noaa:
Let's use the limited jdiag files to demonstrate what it will look like for short names and long names:   
(I think the short names method is simpler, consistent, clean, concise without losing anything and should be preferred. It is also consistent with the rrfs-workflow style, i.e. if there is a simple way to do things, let's go that way)
**short names**
jdiag short

**long names**
jdiag long

guoqing-noaa:
A list of only jdiag files:
**short names**
jdiag short2

**long names**
jdiag long2

guoqing-noaa:
Here is another example of how the short names will be used:
image

delippi:
Where does the `jquery` tool come from? Do I need to install something to use that or can I use something in your path? What other relevant commands like `jquery` do you suspect we will use?

SamuelDegelia-NOAA:
Is the main benefit of changing these that they can be used with the `jquery` tool? I typically prefer the longer names because it makes it easier to find certain jdiag files while running experiments. For example, that way I don't have to remember that 133/233 corresponds to aircar. Plus it is nice to have consistency between the obspace output in the JEDI stdout and the yaml and jdiag filenames.

guoqing-noaa:
> Is the main benefit of changing these that they can be used with the `jquery` tool? I typically prefer the longer names because it makes it easier to find certain jdiag files while running experiments. For example, that way I don't have to remember that 133/233 corresponds to aircar. Plus it is nice to have consistency between the obspace output in the JEDI stdout and the yaml and jdiag filenames.

@SamuelDegelia-NOAA Thanks for the feedback.
There are two parts in this PR:

One is to use both short names and long names in the `obs space.name` part. The `jquery` screenshot is just to show that the short names may be used by some downstream applications. To me, this `obs space.name` is like a comment in a code. Adding a short name does not affect JEDI functionalities as it is a comment and not used by any other codes. But it can help some downstream users.

The major point is the two posts of screenshots about the jdiag file names. 
The very verbose naming convection is NOT consistent with the rrfs-workflow style.
`jdiag_aircrft_airTemperature_130.nc4` is not consistent with `jdiag_atms_n20.nc4` either.

If keeping `aircft` is preferred, we may use `jdiag_aircft_t130.nc4` which is acceptable to me. Thanks!

SamuelDegelia-NOAA:
> The very verbose naming convection is NOT consistent with the rrfs-workflow style. 

Could you explain that briefly? Does rrfs-workflow try to use short file names? 

> If keeping `aircft` is preferred, we may use `jdiag_aircft_t130.nc4` which is acceptable to me. Thanks!

That would be preferable to me! If we change it though I would also to change the yaml names so they are consistent with the jdiag names. What about this @delippi and @guoqing-noaa? 

yaml filename = `aircft_t130.yaml`
obs space name = `aircft_t130`
jdiag name = `jdiag_aircft_t130.nc4`

guoqing-noaa:
> > The very verbose naming convection is NOT consistent with the rrfs-workflow style.
> 
> Could you explain that briefly? Does rrfs-workflow try to use short file names?

Yes. I mean for all possible situations. If a simple form is enough without any confusion, we will use that. For example, the forecast task is named `fcst`, and the lateral boundary condition task is named `lbc`. A lot of such examples in rrfs-workflow.

> > If keeping `aircft` is preferred, we may use `jdiag_aircft_t130.nc4` which is acceptable to me. Thanks!
> 
> That would be preferable to me! If we change it though I would also to change the yaml names so they are consistent with the jdiag names. What about this @delippi and @guoqing-noaa?
> 
> yaml filename = `aircft_t130.yaml` obs space name = `aircft_t130` jdiag name = `jdiag_aircft_t130.nc4`

I like this style. But I will NOT ask for changing the yaml file names at the moment as that will involve more effort.
Using them for obs space name and jdiag name is already very good to me. Thanks!

guoqing-noaa:
@delippi @SamuelDegelia-NOAA 
I think using the not-too-long and not-too-short names proposed by Sam is a good solution. I have modified this PR based on that. Let me know if you have any comments. Thanks!

hu5970:
I agree. The file name needs to be understandable by most of the people. t133 is too short as few people know what is 133.
At least add aircraft in the file name to give user a clue the data platform of this kind of observations.

xyzemc:
> @delippi @SamuelDegelia-NOAA I think using the not-too-long and not-too-short names proposed by Sam is a good solution. I have modified this PR based on that. Let me know if you have any comments. Thanks!

Thanks @guoqing-noaa to modify the name to balance the more understandable and fit the workflow. This change looks good to me.

rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
```diff
@@ -14,7 +14,7 @@
          obsdataout:
            engine:
              type: H5File
-             obsfile: jdiag_msonet_stationPressure_188.nc4
+             obsfile: jdiag_msonet_p188.nc4
```
hu5970, line 17: Should we change p188 to ps_188?

rrfs-test/validated_yamls/templates/obtype_config/msonet_specificHumidity_188.yaml
```diff
@@ -14,7 +14,7 @@
          obsdataout:
            engine:
              type: H5File
-             obsfile: jdiag_msonet_specificHumidity_188.nc4
+             obsfile: jdiag_msonet_q188.nc4
```
hu5970, line 17: Should we change q188 to q_188?

rrfs-test/validated_yamls/templates/obtype_config/msonet_winds_288.yaml
```diff
@@ -14,7 +14,7 @@
          obsdataout:
            engine:
              type: H5File
-             obsfile: jdiag_msonet_winds_288.nc4
+             obsfile: jdiag_msonet_uv288.nc4
```
hu5970, line 17: Should we change uv288 to uv_288?

rrfs-test/validated_yamls/templates/obtype_config/adpupa_specificHumidity_120.yaml
```diff
@@ -14,7 +14,7 @@
          obsdataout:
            engine:
              type: H5File
-             obsfile: jdiag_adpupa_specificHumidity_120.nc4
+             obsfile: jdiag_adpupa_q120.nc4
```
hu5970, line 17: should we change q120 to q_120? Also, do we have a better short way for specificHumidity? Moisture has several units. It is better to make it clear which unit is used here.

delippi: I still prefer using the longer name `jdiag_adpupa_specificHumidity_120.nc`. Then we keep the jedi yaml, obs space name, and the jdiag name all consistent. Keeping them aligned should simplify things from an engineering perspective, IMO.

hu5970:
Hi All, Let's wait for @ShunLiu-NOAA  coming back and coordinate a discuss on how we should move forward.  Thanks, Ming

rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
guoqing-noaa: Sure, I will change to ps188

rrfs-test/validated_yamls/templates/obtype_config/msonet_specificHumidity_188.yaml
guoqing-noaa: Do we really need to add an extra `_`? 
`jdiag_msonet_q188`
vs
`jdiag_msonet_q_188`

rrfs-test/validated_yamls/templates/obtype_config/adpupa_specificHumidity_120.yaml
guoqing-noaa: > should we change q120 to q_120? Also, do we have a better short way for specificHumidity? Moisture has several units. It is better to make it clear which unit is used here.

Could you elaborate on what you mean by "better short way for specific humidity"?

guoqing-noaa: If you check this line 59 of this page:
https://github.com/NOAA-EMC/rrfs-workflow/pull/661/files
`specific_humidity` has been changed to `water_vapor_mixing_ratio_wrt_moist_air`

I think at the workflow level, using `q120` is far enough, `adpupa_q120` already overkills.

ShunLiu-NOAA:
I think we need to check GDAS as well. Eventfully, we need handle the diagnostic files using the same way as GDAS so that we can use the same utilities to further process those files for monitoring and diagnostic jobs.

guoqing-noaa:
I've been reading the PEP 8 style guide today:
See the screenshot below, we can see that short, concise forms are always preferred. If we can do things in a simple way without ambiguity, don't make them complicated. 

In terms of consistency, I think `aircar_t133` will provide the best consistency no matter how JEDI changes its internal naming system (which has happened a few times already) and how many sets of naming systems are inside JEDI.

Even in RDASApp, we have used as many concise forms as possible:  

We use `offline_add_var_to_ioda.py` instead of `offline_add_variables_to_ioda.py`,   
`offline_domain_check_satrad.py` instead of `offline_domain_check_satellite_radiance.py`
`obsdatain` instead of `observation_data_input`,    .....................
   
image

guoqing-noaa:
Another example, for people who have used GrADS, I think most people will type `d` instead of `display` to make plots. 
Short, concise forms are adopted by almost all industrial products, including the Linux System.

image
_(copied from [Training_GrADS_exercises.pdf](https://www.cpc.ncep.noaa.gov/products/international/grads/Training_GrADS_exercises.pdf))_

guoqing-noaa:
I would like to propose a middle-ground solution:

1. individual filenames use long names as is (except satellite files uses short names as is)
2. `obs space.name` use a combination of short name and long names since it is just a comment thing in JEDI,   
eg:
```
t133=aircar_airTemperature_133
ztd=GNSS_Zenith_Total_Delay
atms_npp=Advanced_Technology_Microwave_Sounder_National_Polar-orbiting_Partnership
t126=rassda_airTemperature_126_Radio_Acoustic_Sounding_System # to be frank, I don't know what rassda is until I google a few times).  
```
3. use the intermediate names for jdiag file names
eg: `jdiag_aircar_t133.nc`, `jdiag_gnss_ztd.nc`, `jdiag_atms_npp.nc`

@hu5970 @ShunLiu-NOAA @delippi @SamuelDegelia-NOAA @xyzemc @HuiLiu-NOAA

HuiLiu-NOAA:
The short names look reasonable and convenient to use in the workflow.

guoqing-noaa:
Leave a comment here: It was not a nice way to do things like this today.

hu5970:
We should not keep arguing on this. Let's use airTemperature, relativeHumidity, winds, surfacePressure instead of t, q, us, ps. Thanks, Ming

PR #305, opened 2025-03-02
add yaml_cat_together.py to generate MPASJEDI super YAML files for `YAML_GEN_METHOD=1` in rrfs-workflow

guoqing-noaa:
This PR is to generate MPASJEDI super YAML files for `YAML_GEN_METHOD=1` in rrfs-workflow

Currently, there are a few discrepancies between the super YAML files used in rrfs-workflow and those generated in RDASApp:
1.  workkflow uses `date: &analysisDate '@analysisDate@'`
2. workflow uses `filename: mpasin.nc` for the `output` section in `jedvar.yaml`
3. workflow uses `begin: '@beginDate@'` and `length: PT4H ` for `time window`
4. workflow uses `filename: ./data/ens/mem%iMember%.nc` for ensembles
5. workflow uses short names for jdiag files
6. workflow uses `seed_time: "@analysisDate@`
7. workflow uses `prior_mean.nc` to store the prior ensemble mean
8. workflow uses `filename: ./data/ens/mem%{member}%.nc` for the `output` section in `getkf_solver.yaml` for updating all ensembles
9. workflow assuems jdiag files coming from `data/jdiag/jdiag_*.nc` for the getkf solver.
10. wokflow does not need the `test` section

To avoid manual editing, I developed a Python utility that can generate MPASJEDI super YAML files that are ready for rrfs-workflow "out-of-the-box" for `YAML_GEN_METHOD=1`. I will illustrate its use in the next post.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -160,3 +160,6 @@ install*/
 expr
 fix/INIT_DONE
 fix/.agent
+rrfs-test/validated_yamls/templates/jedivar.yaml
+rrfs-test/validated_yamls/templates/getkf_observer.yaml
+rrfs-test/validated_yamls/templates/getkf_solver.yaml
--- a/rrfs-test/validated_yamls/templates/basic_config/mpasjedi_getkf_observer.yaml
+++ b/rrfs-test/validated_yamls/templates/basic_config/mpasjedi_getkf_observer.yaml
@@ -14,7 +14,7 @@ output: # for outputting mean posterior
   stream name: analysis
 
 time window:
-  begin: 2024-05-26T21:00:00Z
+  begin: '2024-05-26T21:00:00Z'
   length: PT6H
 
 geometry:
--- a/rrfs-test/validated_yamls/templates/basic_config/mpasjedi_getkf_observer.yaml
+++ b/rrfs-test/validated_yamls/templates/basic_config/mpasjedi_getkf_observer.yaml
@@ -3,6 +3,20 @@ _member: &memberConfig
   state variables: [spechum,surface_pressure,temperature,uReconstructMeridional,uReconstructZonal,theta,rho,u,qv,pressure,landmask,xice,snowc,skintemp,ivgtyp,isltyp,snowh,vegfra,u10,v10,lai,smois,tslb,pressure_p,qc,qi,qg,qr,qs,cldfrac]
   stream name: ensemble
 
+increment variables: [temperature, spechum, uReconstructZonal, uReconstructMeridional, surface_pressure]
+
+output mean prior:
+  filename: ./bkg.$Y-$M-$D_$h.$m.$s.nc
+  stream name: background
+
+output: # for outputting mean posterior
+  filename: ./ana.$Y-$M-$D_$h.$m.$s.nc
+  stream name: analysis
+
+time window:
+  begin: 2024-05-26T21:00:00Z
+  length: PT6H
+
 geometry:
   nml_file: ./namelist.atmosphere
   streams_file: ./streams.atmosphere
@@ -10,29 +24,19 @@ geometry:
   interpolation type: unstructured
   iterator dimension: 2
 
-increment variables: [temperature, spechum, uReconstructZonal, uReconstructMeridional, surface_pressure]
-
 background: 
-    members from template:
-      template:
-        <<: *memberConfig
-        filename: ./data/ens/mem%iMember%/mpasout.2024-05-27_00.00.00.nc
-      pattern: "%iMember%"
-      start: 1
-      zero padding: 3
-      nmembers: 30
-
-observations:
-     observers:
-        "@OBSERVATIONS@"
+  members from template:
+    template:
+      <<: *memberConfig
+      filename: ./data/ens/mem%iMember%/mpasout.2024-05-27_00.00.00.nc
+    pattern: "%iMember%"
+    start: 1
+    zero padding: 3
+    nmembers: 30
 
 driver:
   run as observer only: true
   update obs config with geometry info: false
-
-time window:
-       begin: 2024-05-26T21:00:00Z
-       length: PT6H
 
 local ensemble DA:
   solver: GETKF
@@ -46,13 +50,9 @@ local ensemble DA:
     rtpp: 0.6
     mult: 1.1
 
-output mean prior:
-  filename: ./bkg.$Y-$M-$D_$h.$m.$s.nc
-  stream name: background
-
-output: # for outputting mean posterior
-  filename: ./ana.$Y-$M-$D_$h.$m.$s.nc
-  stream name: analysis 
+observations:
+     observers:
+        "@OBSERVATIONS@"
 
 test:
   reference filename: testoutput/rrfs-mpasjedi-getkf-observer.ref
--- a/rrfs-test/validated_yamls/templates/basic_config/mpasjedi_getkf_solver.yaml
+++ b/rrfs-test/validated_yamls/templates/basic_config/mpasjedi_getkf_solver.yaml
@@ -14,7 +14,7 @@ output: # for outputting mean posterior
   stream name: analysis
 
 time window:
-  begin: 2024-05-26T21:00:00Z
+  begin: '2024-05-26T21:00:00Z'
   length: PT6H
 
 geometry:
--- a/rrfs-test/validated_yamls/templates/basic_config/mpasjedi_getkf_solver.yaml
+++ b/rrfs-test/validated_yamls/templates/basic_config/mpasjedi_getkf_solver.yaml
@@ -3,6 +3,20 @@ _member: &memberConfig
   state variables: [spechum,surface_pressure,temperature,uReconstructMeridional,uReconstructZonal,theta,rho,u,qv,pressure,landmask,xice,snowc,skintemp,ivgtyp,isltyp,snowh,vegfra,u10,v10,lai,smois,tslb,pressure_p,qc,qi,qg,qr,qs,cldfrac]
   stream name: ensemble
 
+increment variables: [temperature, spechum, uReconstructZonal, uReconstructMeridional, surface_pressure]
+
+output mean prior:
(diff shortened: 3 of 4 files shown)


guoqing-noaa:
Examples of using `yaml_cat_together.py` :
    
Clone the PR branch `workflow`  from guoqing-noaa's fork:
```
git clone -b workflow https://github.com/guoqing-noaa/RDASApp
cd RDASApp/rrfs-test/validated_yamls/templates
```
1. Generate `jedivar.yaml`, `getkf_observer.yaml` and `getkf_solver.yaml` with all available observers under `obtype_config/`:
```
yaml_cat_together.py all
```
2.  Generate super YAML files using only a subset of observers:
```
yaml_cat_together.py t133,t187
```
3. Check current available observers:
```
yaml_cat_together.py ?
```

delippi:
What was wrong with the gen_jedivar_yaml script in https://github.com/NOAA-EMC/rrfs-workflow/pull/630/files? As @HuiLiu-NOAA also mentioned, we would change the static list to a dynamic list from of all available ioda_bufr.

HuiLiu-NOAA:
A check of available ioda files and adding corresponding "obs space" in th esuper-yaml would be nice.

Also, it would be desirable to move all of the jdiag files into a single location under com/, (rather than currently these jdiag files are separated located in many sub-directories under stmp/

guoqing-noaa:
> A check of available ioda files and adding corresponding "obs space" in th esuper-yaml would be nice.
> 
> Also, it would be desirable to move all of the jdiag files into a single location under com/, (rather than currently these jdiag files are separated located in many sub-directories under stmp/

Thanks, Hui!  Yes, you are right; we need to copy jdiag files into com/.  We will add that soon. Thanks for the suggestion!

guoqing-noaa:
Before the merge of this PR, one may use the `yaml_cat_togther.py` as follows (assuming you have the latest RDASApp)
```
cd ${HOMErdasapp}/rrfs-test/validated_yamls/templates
wget https://raw.githubusercontent.com/guoqing-noaa/jtools/refs/heads/main/misc/yaml_cat_together.py
./yaml_cat_together.py t133,t187  # or all or ?
```

HuiLiu-NOAA:
> Before the merge of this PR, one may use the `yaml_cat_togther.py` as follows (assuming you have the latest RDASApp)
> 
> ```
> cd ${HOMErdasapp}/rrfs-test/validated_yamls/templates
> wget https://raw.githubusercontent.com/guoqing-noaa/jtools/refs/heads/main/misc/yaml_cat_together.py
> ./yaml_cat_together.py t133,t187  # or all or ?
> ```

Thanks. I guess the script for jedivar would also be updated?

guoqing-noaa:
> Thanks. I guess the script for jedivar would also be updated?

Yes, it generates "jedivar.yaml", "getkf_observer.yaml," and "getkf_solver.yaml" at the same time. Let me know if you find any issues. Thanks!

delippi:
> What was wrong with the gen_jedivar_yaml script in https://github.com/NOAA-EMC/rrfs-workflow/pull/630/files? As @HuiLiu-NOAA also mentioned, we would change the static list to a dynamic list from of all available ioda_bufr.

@guoqing-noaa did you see I already made this capability in a bash utility (workflow/ush/gen_jedivar_yaml_nonjcb.sh) in rrfs-workflow 3 weeks ago? The bash version is much simpler btw. Your python version also assumes we will use your short names. Which I don't see a need for.

I don't understand why to reinvent the wheel when we already have a decent system in place and the goal is JCB anyways.

HuiLiu-NOAA:
> > What was wrong with the gen_jedivar_yaml script in https://github.com/NOAA-EMC/rrfs-workflow/pull/630/files? As @HuiLiu-NOAA also mentioned, we would change the static list to a dynamic list from of all available ioda_bufr.
> 
> @guoqing-noaa did you see I already made this capability in a bash utility (workflow/ush/gen_jedivar_yaml_nonjcb.sh) in rrfs-workflow 3 weeks ago? The bash version is much simpler btw. Your python version also assumes we will use your short names. Which I don't see a need for.
> 
> I don't understand why to reinvent the wheel when we already have a decent system in place and the goal is JCB anyways.

Maybe at this point, we just get the workflow working first. We may refine the workflow at a later point.

guoqing-noaa:
@delippi 
There are three "YAML_GEN_METHOD"s in rrfs-workflow as discussed:
https://github.com/NOAA-EMC/rrfs-workflow/blob/09dde8f9d717c2c7cb9dce65ea778c8d463e6505/scripts/exrrfs_jedivar.sh#L110

Your bash utility and the future jcb method are still there. 

This PR is for method 1, where super yaml files come for `${PARMrrfs}`

This reminds me that I should make the PR title and content clearer.

ShunLiu-NOAA:
@guoqing-noaa Is it possible that we combine method1 and method2 to reduce the future maintenance work for the two scripts?

guoqing-noaa:
> @guoqing-noaa Is it possible that we combine method1 and method2 to reduce the future maintenance work for the two scripts?

I think it will be possible in the future.  Right now, we still need to figure out how to make aircraft observation correctly assimilated in the cycling system. We will test a few important observation types first, both in real time and in retros. We should be in better shape after a few months.

ShunLiu-NOAA:
@guoqing-noaa I think this practice might ruin the effort of ctest in RDASApp and probably lead to diverse the inconsistent test results between RDASApp and workflow. Let us discuss how to better handle this.

guoqing-noaa:
@ShunLiu-NOAA Could you clarify how this PR will ruin the ctest in RDASApp?
All YAML contents come from the validated templates.

guoqing-noaa:
> @ShunLiu-NOAA Could you clarify how this PR will ruin the ctest in RDASApp?

ALL YAML contents come from the "validated" templates.

ShunLiu-NOAA:
It is possible that ctest uses a different YMAL than that the workflow. Even if ctest passes, there is no guarantee JEDI will work in the workflow. You must keep track of the difference between ctest YMAL and workflow YAML. As you are aware, it might not be easy to do this until we have better way to ensure that method1 and method2 are able to generate identical YAML. Otherwise, it doesn't make sense to do ctest in RDASApp.

guoqing-noaa:
> It is possible that ctest uses a different YMAL than that the workflow. Even if ctest passes, there is no guarantee JEDI will work in the workflow. You must keep track of the difference between ctest YMAL and workflow YAML. As you are aware, it might not be easy to do this until we have better way to ensure that method1 and method2 are able to generate identical YAML. Otherwise, it doesn't make sense to do ctest in RDASApp.

@ShunLiu-NOAA Thanks for the input. The ctest is to make sure all algorithms, all observers, and all obs filters work as expected with new updates. To my knowledge, this PR does not change any of that. All algorithms, all observers and all obs filters come from the same `rrfs-test/validated_yamls/templates`.  

It will be good if you can give specific examples about what changes of this PR will cause a YAML to work in RDASApp but fail in rrfs-workflow. Thanks!

ShunLiu-NOAA (approved):

delippi:
> > It is possible that ctest uses a different YMAL than that the workflow. Even if ctest passes, there is no guarantee JEDI will work in the workflow. You must keep track of the difference between ctest YMAL and workflow YAML. As you are aware, it might not be easy to do this until we have better way to ensure that method1 and method2 are able to generate identical YAML. Otherwise, it doesn't make sense to do ctest in RDASApp.
> 
> @ShunLiu-NOAA Thanks for the input. The ctest is to make sure all algorithms, all observers, and all obs filters work as expected with new updates. To my knowledge, this PR does not change any of that. All algorithms, all observers and all obs filters come from the same `rrfs-test/validated_yamls/templates`.
> 
> It will be good if you can give specific examples about what changes of this PR will cause a YAML to work in RDASApp but fail in rrfs-workflow. Thanks!

I just wanted to try to clarify what I think Shun's concern is. This new method for generating the YAML *could* create a different YAML then how we make them with for the ctests. It's okay to have a few ways to generate the yamls now and we can all try out each method to see what works best. But ultimately, we need to use the same method for both the ctest in RDASApp and for rrfs-workflow. AND the other point here is that RDASApp should dictate what is used in rrfs-workflow for anything DA and not the other way around. This is not to say that we might not end up using the python version for the ctests if we can show that it is better than what we're using now.

guoqing-noaa:
@delippi Thanks for the input. I don't think there are two sets of super YAML files as all differences are cosmetic ones, which will exist anyway until we update the MPAS test case in RDASApp. We would like to use rrfs-workflow 60 vertical levels, the GSL MPAS physics, a set of new fix files, etc. It will take some time and effort to complete this. As we rush to a reasonable rrfs.v2 demonstration for this year's HWT SFE, we, or at least I, don't have enough time to update RDASApp for this before SFE. It will be great if anyone else can take lead on this. If not, current RDASApp is also good enough in terms of testing algorithms, observers and obs filters.

delippi:
> ...I, don't have enough time to update RDASApp for this before SFE. It will be great if anyone else can take lead on this.

@guoqing-noaa I'm not sure I followed. What specifically are you mentioning here?

HuiLiu-NOAA:
I agree that it would make more sense to update RDASApp settings according to the needs of realtime workflow.

guoqing-noaa:
> > ...I, don't have enough time to update RDASApp for this before SFE. It will be great if anyone else can take lead on this.
> 
> @guoqing-noaa I'm not sure I followed. What specifically are you mentioning here?

Sorry that it may not be clear here. I mean we have a need to update RDASApp to use rrfs-workflow vertical levels, physics and associated fix files but I will not able to do that before May2025 (when SFE starts). But it will be good if anyone else can make this happen and update RDASApp for that. Thanks!

ShunLiu-NOAA:
@guoqing-noaa @delippi Thank you for the good discussion. Let's merge this PR first. If any new function/features are added to python/shell version, please make sure that we update the both versions simultaneously to avoid any unnecessary confusion.

Issue #294, opened 2025-02-26
Incorporate sfcshp DA into JEDI for RRFS


delippi:
## Description
The UFO in JEDI is the component that not only computes model-simulated observations, but also houses filters and methods for observation QC, ob errors, and bias correction. The GSI observer is the equivalent. Many forward operators for various observations have been developed for the UFO. These operators can be utilized in RDAS. However these operators still must be tested for RDAS. The steps for transition by observation platform are as follows:

1. Establish IODA processing
2. Establish operator in UFO
3. Establish YAML input file
4. Validate with test assimilation experiments
5. Create ctest, if necessary
6. sfcshp data includes bufr obtypes=*80

## Requirements
To create yaml configurations for assimilating sfcshp data.

## Acceptance Criteria (Definition of Done)
- [ ] Phase 1 validation (hofx validation using GSI-IODA and GSI-geovals as necessary; GSI vs FV3-JEDI)
- [ ] Phase 2 validation (QC validation; no reliance on GSI except to be used as the baseline; GSI vs FV3-JEDI)
- [ ] Phase 3 validation (FV3-JEDI and GSI vs MPAS-JEDI)

- Link any relevant pull requests here:
  - 

## Dependencies
None

delippi:
**Phase 1 update: (for 180/280)**
Phase 1 results are in. These results were created using GSI-IODA, but allowing JEDI to handle the rest (geovals, QC, and oberror inflation). ObsErrorFactorConventional is included in these yamls, but it is turned off. Overall airTemperature and specificHumidity were pretty good. There were a few uncertainties with stationPressure and winds. However, the results when using GSI-geovals (not shown) didn't have any of the following issues. Phase 2 and 3 might be able to reveal more about these issues, so I have decided this is a good time to merge these results and establish the phase 1 yamls. Phase 1 yamls will not affect cycling results since they will not be used until passing phase 3. I detail the uncertainties below.

Uncertainties:
- winds: JEDI has too few observations. I tracked this down to slight differences in pressure in GSI thus duplicate check is less restrictive hence the higher counts in GSI. I don't understand why GSI had pressure values that varied by 1-3 hPa for these duplicate observations whereas JEDI had constant values for the corresponding lat/lon obs. When changing the grouping in GSI to not check for matching pressure, the observation counts match better. The grouping using stationIdentification in JEDI was also different but still close to this result.
[...]

PR #308, opened 2025-03-03
Phase 1 validation: sfcshp & adpsfc (183/284) yamls

delippi:
## Description
These ObTypes come from both sfcshp and adpsfc. Since we currently create IODA files into subsets like ioda_adpsfc.nc and ioda_sfcshp.nc, I've created a separate yaml for each subset type. This type passes Phase 1 validation standards, but there are still some uncertainties going into Phase 2. I've noted the results and the uncertainties in this [comment](https://github.com/NOAA-EMC/RDASApp/issues/294#issuecomment-2695331331).

The point of this PR is to establish these yamls in RDASApp. This won't affect cycling DA results because we will only use configurations that have passed phase 3 validation.

## Issue(s) addressed
- https://github.com/NOAA-EMC/RDASApp/issues/294

## Checklist

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_183.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_183.yaml
@@ -1,0 +1,275 @@
+     - obs space:
+         name: adpsfc_airTemperature_183
+         distribution:
+           name: "@DISTRIBUTION@"
+           halo size: 300e3
+         obsdatain:
+           engine:
+             type: H5File
+             obsfile: "data/obs/ioda_adpsfc.nc"
+             #obsfile: "obsout/sfcship_tsen_obs_2024052700_dc.nc4"
+           obsgrouping:
+             group variables: ["stationIdentification"]
+             sort variable: "pressure"
+             sort order: "descending"
+         obsdataout:
+           engine:
+             type: H5File
+             obsfile: jdiag_adpsfc_airTemperature_183.nc4
+             allow overwrite: true
+         io pool:
+           max pool size: 1
+         observed variables: [airTemperature]
+         simulated variables: [airTemperature]
+
+       obs operator:
+           name: VertInterp
+           vertical coordinate: air_pressure
+           observation vertical coordinate: pressure
+           observation vertical coordinate group: MetaData
+           interpolation method: log-linear
+           #gsi geovals:
+           #  filename: "obsout/sfcship_tsen_geoval_2024052700.nc4"
+           #  levels_are_top_down: False
+           variables:
+           - name: airTemperature
+
+       linear obs operator:
+         name: VertInterp
+
+       obs error:
+         covariance model: diagonal
+
+       obs localizations:
+         - localization method: Horizontal Gaspari-Cohn
+           lengthscale: 300e3 # orig
+
+       obs filters:
+         # ------------------
+         # airTemperature (183)
+         # ------------------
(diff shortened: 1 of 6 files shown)


HuiLiu-NOAA:
Is the action  " reduce obs space" really needed under every filter?

SamuelDegelia-NOAA:
> Is the action  " reduce obs space" really needed under every filter?

This has come up a couple of times recently so maybe we should have some discussion on it. My opinion has been that there is no reason not to include the filter action on every possible pre-filter. It saves computational time by removing obs from the hofx calculation. Even if it is a very small savings, why not take advantage of it? Plus, the savings are much larger for GETKF which has to compute hofx for 360 members (12 modulated members per real member).

HuiLiu-NOAA:
I guess this yaml will work only with FV3-JEDI, not MPAS-JEDI, when the option is turned on in adpsfc_specificHumidity_183.yaml:
                           request_saturation_specific_humidity_geovals: true

HuiLiu-NOAA:
> > Is the action  " reduce obs space" really needed under every filter?
> 
> This has come up a couple of times recently so maybe we should have some discussion on it. My opinion has been that there is no reason not to include the filter action on every possible pre-filter. It saves computational time by removing obs from the hofx calculation. Even if it is a very small savings, why not take advantage of it? Plus, the savings are much larger for GETKF which has to compute hofx for 360 members (12 modulated members per real member).

OK. you mean the action "reduce obs space" will work with all the QC rejected obs. This is actually very nice to have!!

HuiLiu-NOAA:
Are there any indications in the validated yamls under rrfs-test/validated_yamls/templates/obtype_config/, which have passed phase 3 validations?

delippi:
> I guess this yaml will work only with FV3-JEDI, not MPAS-JEDI, when the option is turned on in adpsfc_specificHumidity_183.yaml: request_saturation_specific_humidity_geovals: true

It will not work in MPAS-JEDI setup (yet). It also won't work in FV3-JEDI configuration without that setting true. Because the code expects qsat to scale the ob error from RH. The code doesn't automatically request the geovals be created.

> Are there any indications in the validated yamls under rrfs-test/validated_yamls/templates/obtype_config/, which have passed phase 3 validations?

There is no indication in the yamls themselves. You can refer to our spread sheet of all the obs.

delippi:
> > > Is the action  " reduce obs space" really needed under every filter?
> > 
> > 
> > This has come up a couple of times recently so maybe we should have some discussion on it. My opinion has been that there is no reason not to include the filter action on every possible pre-filter. It saves computational time by removing obs from the hofx calculation. Even if it is a very small savings, why not take advantage of it? Plus, the savings are much larger for GETKF which has to compute hofx for 360 members (12 modulated members per real member).
> 
> OK. you mean the action "reduce obs space" will work with all the QC rejected obs. This is actually very nice to have!!

Yes. The reduce obs space does it's thing only on the filter it is specified under. It has to be explicitly defined for each pre-filter for it to be enabled. There is no limit to the number of filters it can be applied to. The only limit is it must be applied to pre-filters.

SamuelDegelia-NOAA (dismissed):

delippi:
converted to draft really quick to make changes similar to #314

ShunLiu-NOAA:
@hu5970 @guoqing-noaa @spanNOAA Please review this PR if you get a chance. I will merge this PR into develop by the end of this Friday.

ShunLiu-NOAA (approved):

SamuelDegelia-NOAA (approved):

PR #311, opened 2025-03-04
hotfix for rrfs-workflow - make all missing QualityMarker to 15

guoqing-noaa:
Add the modification capability into `offline_add_var_to_ioda.py` and rename it to `offline_ioda_tweak.py`

### DIFF
--- a/rrfs-test/IODA/offline_ioda_tweak.py
+++ b/rrfs-test/IODA/offline_ioda_tweak.py
@@ -7,10 +7,11 @@ import warnings
 import os
 
 """
-This program makes a copy of an original IODA file and can add additional
+This program makes a copy of an original IODA file and can add/modify additional
 adhoc variables. This program was originally written to add the
 MetaData/longitude_latitude_pressure for use in l_closeobs duplicate checking
 but others can be added as necessary.
+3/4/2025: change all QualityMarker with missing values to 15
 """
 
 # Disable warnings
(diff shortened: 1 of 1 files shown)


delippi:
Thanks @guoqing-noaa. I'm curious. Can you provide some more details about what exactly happens when the QM is undefined? Also, what scenario are the QM's not defined? I haven't come across that case. Maybe this is something that needs handled elsewhere (either now or later)? One option *could* be to change the first QC filter of each yaml to instead of reject QMs 4-15 only accept QMs 1-3. In hind sight that might've been better since they would have to be defined to make it in.

I don't particularly care for "tweak", would it be okay to use the work "modify" instead?

And FYI, this script is called in the rrfs-workflow so be sure to change the name of the script there as well.
https://github.com/NOAA-EMC/rrfs-workflow/blob/54bdb0b5a7d6f3764ea41651f08d507d5b3d613a/scripts/exrrfs_ioda_bufr.sh#L38

guoqing-noaa:
@delippi Thanks for the input. I agree that this may be a temporary solution as we have to get rrfs-worfkow run without model crashes for real time and retros.  The acceptance of 0-3 filter had also been tried, but I did not know why it did not work (@chunhuazhou or @spanNOAA may provide more details on this). Maybe you can help us with that. 

`offline_ioda_modify.py` works for me.

guoqing-noaa:
We have a staged case from Chunhua at:
`/lfs6/BMC/wrfruc/gge/tmp/aircraftDA`

There is a `prepbufr` file and `ioda_aircar.nc` file. You can check them. 

```
./pbdump prepbufr AIRCAR dumpall | grep -A 3 'JWYLX5ZA'
```
can give you a quick look of the bad qcflag (The first 4 columns are: number, P Q T).

guoqing-noaa:
convert to draft before a close. @delippi and @chunhuazhou found a workable obs filter to deal with missing qc flags.

Issue #310, opened 2025-03-04
[bug]Observation filtering not working when assimilating t133 from aircar data


chunhuazhou:
When checking our MPAS real-time runs with JEDI on Jet, which assimilates only airTemperature from AIRCAR, we found huge temperature increments, in both 3DVar for the conus-3km and 3DEnVar for conus-12km, as shown in the figures below.

<img width="1920" alt="Image" src="https://github.com/user-attachments/assets/c8050ef3-f473-4a78-9a76-c5e0b4f60b90" />

When tracing down the cause of the huge increment, we noticed bad observation with QualityMarker being "D" (equivalent to 13) or "*" from the prepbufr file. 

```
--------------------------------------------------------------------------------
MESSAGE: AIRCAR     676    87  2025022701
STATION: FDQLX3ZA   257.72    42.09  
TIME:    2025022701      0.17   
ELV:     10363.00
TYPE:         41. ********     133.   
DATA:     
 LVL CAT        POB        SPH        TOB        ZOB        UOB        VOB    
   1   6   250.0(2) *******(*)    -5.3(D) 10363.0(2) *******(*) *******(*)
--------------------------------------------------------------------------------
```

```
--------------------------------------------------------------------------------
MESSAGE: AIRCAR      14    72  2025022701
STATION: JWYLX5ZA   240.49    46.21                                                                                                                                                    
TIME:    2025022701     -0.98
ELV:     11579.00
TYPE:         41. ********     133.
DATA:
 LVL CAT        POB        SPH        TOB        ZOB        UOB        VOB
   1   6   206.6(2) *******(*)    -6.3(*) 11579.0(2) *******(*) *******(*)
--------------------------------------------------------------------------------
```

comparing to good observations with quality marker being 1, such as this one:
```
--------------------------------------------------------------------------------
MESSAGE: AIRCAR       1     3  2025022701
STATION: 1HFRTXBA   104.40     1.16   
TIME:    2025022701     -1.00  
ELV:      3475.00
TYPE:         41. ********     133.   
DATA:    
 LVL CAT        POB        SPH        TOB        ZOB        UOB        VOB    
   1   6   659.8(2) *******(*)     7.1(1)  3475.0(2) *******(*) *******(*)
--------------------------------------------------------------------------------
```

Checking the ioda_aircar.nc, we see the quality markers look like this, by using "$ ncdump -v QualityMarker/airTemperature ioda_bufr_01/ioda_aircar.nc":

```
[...]

PR #314, opened 2025-03-05
bugfix to handle missing QualityMarkers in jediyaml and removes duplicate action key.

delippi:
## Bug Fix Description

In rrfs-workflow, GSL colleagues (Chunhua and others) found that aircar_airTemperature_133 assimilation experiments had unrealistic increments. It was found that this is due to the jediyaml not correctly handling missing QualityMarker values. Essentially, those observations were getting into the analysis causing the very large and unreasonable analysis increments.

I also snuck in the removal of the duplicate action key which wasn't necessary.

In collaboration with GSL, we have tested a solution to correctly reject observations where QualityMarker > 3 <ins>**OR**</ins> QualityMarker == missing.

## Issue(s) addressed
- https://github.com/NOAA-EMC/RDASApp/issues/310

## Checklist


delippi:
@chunhuazhou, can you confirm that we don't need that duplicate variable like like this:
```
           where:
           - variable: QualityMarker/airTemperature
             is_in: 4-15
           - variable: QualityMarker/airTemperature
             value: is_not_valid
           where operator: or
```

I made all the changes assuming the duplicate line was not needed. I will add those lines back if you find it is required.

chunhuazhou:
@delippi  I think the extra variable line is needed. I tested without it and it seems not working. Thanks!

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_181.yaml
```diff
@@ -55,9 +55,10 @@
            where:
            - variable: QualityMarker/airTemperature
              is_in: 4-15
+             value: is_not_valid
+             where operator: or
```
chunhuazhou: This "where operator: or" should use same indentation as "where:"

chunhuazhou: @delippi My earlier test somehow messed up. Actually we don't need that duplicate variable line.

delippi: Oh my... I just added it all.
Yep, my script put the where operator at the wrong indentation. I had noticed that and fixed it.

delippi: It looks cleaner without. I will make those changes tomorrow. Thanks!

chunhuazhou:
@delippi Sorry for the confusion! It does look cleaner without, even though both works.  Thanks for fixing them!

delippi:
> @delippi Sorry for the confusion! It does look cleaner without, even though both works. Thanks for fixing them!

@chunhuazhou, no worries! I wasn't manually editing anything so it wasn't much work to undo. Thank you so much for tracking down the issue and testing the fixes! I suspect this could really help all my validation tests as well. There have been things at times that I haven't been able to explain and this could certainly be (one of) the reason(s)!

rrfs-test/validated_yamls/templates/obtype_config/vadwnd_winds_224.yaml
```diff
@@ -263,7 +264,7 @@
                name: AbsObsValue/windNorthward
              minvalue: 0
              maxvalue: 1
-           where operator: and
+         where operator: and
```
delippi: This was unintended. I need to fix this.

rrfs-test/validated_yamls/templates/obtype_config/proflr_winds_227.yaml
```diff
@@ -255,7 +256,7 @@
                name: AbsObsValue/windNorthward
              minvalue: 0
              maxvalue: 1
-           where operator: and
+         where operator: and
```
delippi: This was unintended. I need to fix this

rrfs-test/validated_yamls/templates/obtype_config/vadwnd_winds_224.yaml
delippi: fixed

rrfs-test/validated_yamls/templates/obtype_config/proflr_winds_227.yaml
delippi: fixed

guoqing-noaa:
Can we replace
```
         - filter: Perform Action
           apply at iterations: 0,1
           filter variables:
           - name: airTemperature
           where:
           - variable: QualityMarker/airTemperature
             is_in: 4-15
             value: is_not_valid
           where operator: or
           action:
             name: reject
             name: reduce obs space
```
with 
```
filter: Domain Check
where:
- variable:
    name: QualityMarker/airTemperature
  minvalue: 0
  maxvalue: 3
action:
  name: reduce obs space
```
?

If the latter work, we can combine it with the lat/lat domain check as follows:
```
filter: Domain Check
where:
- variable:
    name: MetaData/latitude
  minvalue: -9.06518
  maxvalue: 11.4725
- variable:
    name: MetaData/longitude
  minvalue: 91.7506
  maxvalue: 112.268
- variable:
    name: QualityMarker/airTemperature
  minvalue: 0
  maxvalue: 3
action:
  name: reduce obs space
```

guoqing-noaa:
Ah, can we apply `domian check` to a given variable?

delippi:
@guoqing-noaa, thanks for the suggestion. While it should be possible to combine multiple filters like this, I have a few reasons why I think it *might* not be the best approach: 

1. **Clarity**: Keeping filters separate makes it clear what each one is doing--especially for users who don't look at yamls everyday.
2. **Simplicity**: While separate filters make the yaml longer, having each filter dedicated to a single function reduces the risk of unintended consequences.
4. **Separation of concerns**: The QM check is about data quality, while the domain check is about spatial validity. Combining them blurs their distinct purposes, which could lead to confusion during debugging. Keeping them separate also makes it easier to toggle them on/off as needed.
5. **Debugging**: If observations get rejected, separate filters allows you to pinpoint whether it was due to quality issues or domain constraints.

I know you like things more compact, but I think it's wise to keep one filter focused on one task. These are also my reasons for having each obs space focused on a single platform and obtype. The most extreme example: it is possible to have a single obs space that deals with every single platform and observation, but that would be VERY difficult to understand and I can guarantee there would be unintended consequences that would be practically untraceable.

That said, I do think including the pre-online domain check filter is a good idea--just not combined with another filter.

Do you know if there are any benefits to combining filters (aside from more compact yamls)? The only benefit I can think of would be performance related, but I have never tried this.

guoqing-noaa:
@delippi, Thanks for the input! I am fine with splitting them if it provides more clarity.

I think I just understood that "domain" check may mean "variable" domain; it works similarly to "bounds check" or range check. You can use it to check data/time range, lat range, lon range, qc range, etc. Just my two cents.

delippi:
@guoqing-noaa, yes. You are right that there are many filters that can be switched to do the same thing. I may have not always chosen the best function for each task, but the main thing is that they are separated (in my view). I am all for suggestions on which filter would best categorize a task. For example, "Domain Check" may not be the best to use for the "Time window filter"--category-wise, but it works perfectly well. This is something we can always come back to and very easy to change all at once.

chunhuazhou (dismissed):

SamuelDegelia-NOAA (dismissed):

guoqing-noaa:
Do we want to consider using `actions` instead of `action` for multiple actions?

The following text and examples are copied from:
https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/jedi-components/ufo/qcfilters/FilterOptions.html
`To perform multiple actions, replace the action option, which takes a single action, by actions, which takes a list of actions.`

and Example 4 - [setting a diagnostic flag at observations rejected by a filter](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/jedi-components/ufo/qcfilters/FilterOptions.html#example-4-setting-a-diagnostic-flag-at-observations-rejected-by-a-filter)
```
- filter: Domain Check
  where:
  - variable:
      name: MetaData/latitude
    minvalue: -60
    maxvalue:  60
- filter: Bounds Check
  filter variables:
  - name: airTemperature
  min value: 250
  max value: 350
  # Reject particularly hot and cold observations
  # and mark them with the ExtremeValue diagnostic flag
  actions:
  - name: set
    flag: ExtremeCheck
    ignore: rejected observations
  - name: reject
```

delippi:
> Do we want to consider using actions instead of action for multiple actions?

Thanks for the suggestion. Yes, we should consider using that. I will do a few tests to double check.

guoqing-noaa:
Based on this JEDI doc:
https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/jedi-components/ufo/qcfilters/GenericQC.html#example-2

image

So I would think `RejectList` is more straightforward in this situation.
`reduce obs space` is more like a "bonus" action that can be applied to any pre-filters without the need to specify the `Perform Action` explicitly

delippi:
I think it is reasonable to change back to RejectList at this point. I have no strong feelings about this one.

SamuelDegelia-NOAA (dismissed):

ShunLiu-NOAA (approved):

PR #298, opened 2025-02-27
Generate ctest yamls when RDASApp is built to better sync ctests with templates

SamuelDegelia-NOAA:
## Description

This PR addresses #273 to generate the ctest yamls when RDASApp is built. This will allow us to keep better sync between the ctests and updates to `rrfs-test/validated_yamls/templates`. It will also make reviewing PRs easier if we don't also need to make changes to the super yamls. 

A couple other minor changes are included here. 
* `gen_yaml_ctest.sh` is updated to make it easier for adding new observation yamls and generally follows the script added by rrfs-workflow [#630](https://github.com/NOAA-EMC/rrfs-workflow/pull/630) (with some changes needed specifically for the ctests). Currently, observation yamls are only included in this script and the ctests if they have passed "phase 3" validation. 
* The yamls in `rrfs-test/validated_yamls/templates/obtype_config` now follow the desired configuration for mpas-jedi. Additionally, the `modify_yaml_fv3.py` tool is added and run through `gen_yaml_ctest.sh`. This script will modify the yamls to be used for the FV3-JEDI ctests by 1) uncommenting the ObsErrorFactorPressureCheck for the specificHumidity yamls, and 2) commenting out the ObsErrorFactorPressureCheck for adpupa_120.  
* A new obs directory is linked in for the test case (`obs_2024052700_llp`) that includes the `latitude_longitude_pressure` variable for the new duplicate check filters. 
* `rrfs-test/testinput_expr` is renamed to `rrfs-test/example_yamls` to be more accurate. These yamls can be used as examples for running simple experiments if users do not want to use the larger super yamls.

## Issue(s) addressed
#273

## Dependencies (if applicable)
There will be merge conflicts with #284 or #287 if they are merged first. 

## Checklist
- [x] I have performed a self-review of my own code.
- [x] I have run rrfs tests before creating the PR (if applicable).

PR #320, opened 2025-03-07
Update yamls to comment out ObsErrorFactorPressureCheck for specificHumidity (MPAS-JEDI) and update halo radius

SamuelDegelia-NOAA:
## Description
I have found some additional repeatability issues with PR #298 that are causing ctest failures on Jet. However there is a need to get some of those yaml updates committed sooner for use in rrfs-workflow. So this PR contains just the yaml updates from #298 while the repeatability issues are further investigated. 

Included in these changes are: 
1) Commenting out the `ObsErrorFactorPressureCheck` filter for specificHumidity yamls used in MPAS-JEDI
2) Updating the halo radius to 300 km for some lingering yamls missed earlier
3) Changing the jdiag name for `amsua_n19.yaml` to match our new standard. 

## Issue(s) addressed
None

None

## Checklist
- [ ] I have performed a self-review of my own code.

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_specificHumidity_181.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/adpsfc_specificHumidity_181.yaml
@@ -110,21 +110,21 @@
                  xvals: [110000, 105000, 100000, 95000, 90000, 85000, 80000, 75000, 70000, 65000, 60000, 55000, 50000, 45000, 40000, 35000, 30000, 25000, 20000, 15000, 10000, 7500, 5000, 4000, 3000, 2000, 1000, 500, 400, 300, 200, 100, 0]
                  errors: [0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912, 0.05912]
 
-         # Error inflation based on pressure check (setupq.f90)
-         - filter: Perform Action
-           filter variables:
-           - name: specificHumidity
-           where:
-           - variable: ObsType/specificHumidity
-             is_in: 181
-           action:
-             name: inflate error
-             inflation variable:
-               name: ObsFunction/ObsErrorFactorPressureCheck
-               options:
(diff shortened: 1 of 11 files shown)


guoqing-noaa (approved):
Thank you, @SamuelDegelia-NOAA

delippi:
@SamuelDegelia-NOAA, I'm curious why you wouldn't want to dynamically comment the OEFPC with that python script? Was there trouble with it? Some of these yamls are only completed at Phase 1. That means I still need to test them in Phase 2 still using FV3-JEDI. I'm opposed to manually commenting/uncommenting these, if we can avoid it!

Plus, the correct setting is with the filter turned on and we have a solution in the works. I can't say the same for the OEFC which I do have turned off for about every observation type--that one doesn't work in either FV3-JEDI or MPAS-JEDI.

SamuelDegelia-NOAA:
@delippi In PR #298 we decided to just do the reverse wherein we uncomment out the OEFPC for FV3-JEDI (instead of commenting it out for MPAS-JEDI). The python utility is still used, its just it is now uncommenting instead of commenting. The goal is to have `validated_yamls` default to the MPAS-JEDI configuration, and then we can edit them further using the python code for FV3-JEDI ctests. This reduces the need for outside utilities in rrfs-workflow. 

Once we are able to use `request_saturation_specific_humidity_geovals` for MPAS-JEDI, then we could quickly edit those yamls to add the OEFPC back in. Does that work for you?

delippi:
@SamuelDegelia-NOAA, hmm, I missed that conversation. I think I disagree with that idea and here is why. The function turned on IS technically the correct configuration for MPAS-JEDI--it's just "broken". My main point is, we shouldn't be using the specificHumidity yamls AT ALL for any DA cycling tests in rrfs-workflow right now because we can't expect to get good results from them until this is fixed. Yes, commenting it out makes it work functionally, but it'll be junk results and wasted time cycling that data.

SamuelDegelia-NOAA:
That is a good point @delippi that these moisture yamls probably should not be used right now because of the lack of a working OEFPC filter. Apparently the results when using the specificHumidity yamls without this filter are very poor. 

@guoqing-noaa Would you be okay with holding off on this PR for now? I was under the impression that we wanted these merged earlier to use in rrfs-workflow. But if the specificHumidity yamls are not unsable for mpas-jedi then I do not think there is a need to merge this now.

guoqing-noaa:
> @guoqing-noaa Would you be okay with holding off on this PR for now? I was under the impression that we wanted these merged earlier to use in rrfs-workflow. But if the specificHumidity yamls are not unsable for mpas-jedi then I do not think there is a need to merge this now.

It is fine with me. Thanks!

SamuelDegelia-NOAA:
Okay, closing this issue for now since #298 is still up.

Issue #232, opened 2024-11-21
Incorporate Sonde DA into JEDI for RRFS


delippi:
## Description
The UFO in JEDI is the component that not only computes model-simulated observations, but also houses filters and methods for observation QC, ob errors, and bias correction. The GSI observer is the equivalent. Many forward operators for various observations have been developed for the UFO. These operators can be utilized in RDAS. However these operators still must be tested for RDAS. The steps for transition by observation platform are as follows:

1. Establish IODA processing
2. Establish operator in UFO
3. Establish YAML input file
4. Validate with test assimilation experiments
5. Create ctest, if necessary

Sonde data is bufr obtype=*20

## Requirements
To create yaml configurations for assimilating sonde data.

## Acceptance Criteria (Definition of Done)

- [x] Phase 1 validation (hofx validation using GSI-IODA and GSI-geovals as necessary; GSI vs FV3-JEDI)
- [x] Phase 2 validation (QC validation; no reliance on GSI except to be used as the baseline; GSI vs FV3-JEDI)
- [ ] Phase 3 validation (FV3-JEDI and GSI vs MPAS-JEDI)

- Link any relevant pull requests here:
  - #245 
  - PR for Phase 3 (to be added later)

## Dependencies
- #188 is useful to be used as a building block for this work.
- This is one of the more important ob types to start cycling MPAS-JEDI so we need this ready ASAP.

delippi:
**Phase 2 update**
I did Phase 1 validation awhile ago for these types (unofficially; not documented). They use the exact same operator as mesonet, so I feel confident about it. I will likely still do Phase 1 again just to have it documented.

Here are some preliminary results with some sonde yamls mostly based on the mesonet yamls in PR #188. Ob counts are definitely incorrect. Even when removing all QC the ob counts are badt (for example, only 96 airTemp obs used with/without any QC). This has been noticed before; need to investigate. Increments look very different as well; probably due to mismatch in ob errors? Omb look pretty good (which should mean hofx in Phase 1 is fine).

airTemperature
![fv3jedi_vs_gsi_increment_airTemperature_adpupa_airTemperature_120](https://github.com/user-attachments/assets/5b6e971a-b482-479c-bcd2-7a2141f22fef)
![hist_adpupa_airTemperature_120](https://github.com/user-attachments/assets/fa441483-fc2d-408f-bfd7-c36e644d9c11)

specificHumidity
[...]

PR #323, opened 2025-03-10
Phase 1 validation: adpupa (132/232) yamls

delippi:
## Description
Adding phase 1 adpupa (132/232) yamls. There were no obs in the ctest case to fully validate. This isn't a surprise since these ob types were marked to be pretty sparse in the observation spreadsheet. These yamls were developed following the (120/220) yamls exactly with the exception of the ObsErrors and gross error checks for winds. The point of this PR is to establish these yamls in RDASApp. This won't affect cycling DA results because we will only use configurations that have passed phase 3 validation.

## Issue(s) addressed
- https://github.com/NOAA-EMC/RDASApp/issues/232

## Checklist
- [ ] I have performed a self-review of my own code.

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/adpupa_airTemperature_132.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/adpupa_airTemperature_132.yaml
@@ -1,0 +1,274 @@
+     - obs space:
+         name: adpupa_airTemperature_132
+         distribution:
+           name: "@DISTRIBUTION@"
+           halo size: 300e3
+         obsdatain:
+           engine:
+             type: H5File
+             obsfile: data/obs/ioda_adpupa.nc
+           obsgrouping:
+             group variables: ["stationIdentification"]
+             sort variable: "pressure"
+             sort order: "descending"
+         obsdataout:
+           engine:
+             type: H5File
+             obsfile: jdiag_adpupa_airTemperature_132.nc4
+             allow overwrite: true
+         io pool:
+           max pool size: 1
+         observed variables: [airTemperature]
+         simulated variables: [airTemperature]
+
+       obs operator:
+           name: VertInterp
+           vertical coordinate: air_pressure
+           observation vertical coordinate: pressure
+           observation vertical coordinate group: MetaData
+           interpolation method: log-linear
+           #gsi geovals:
+           #  filename: "obsout/sondes_tsen_geoval_2022052619.nc4"
+           #  levels_are_top_down: False
+           variables:
+           - name: airTemperature
+
+       linear obs operator:
+         name: VertInterp
+
+       obs error:
+         covariance model: diagonal
+
+       obs localizations:
+         - localization method: Horizontal Gaspari-Cohn
+           lengthscale: 300e3 # orig
+
+       obs filters:
+         # ------------------
+         # airTemperature (132)
+         # ------------------
+         # Reject all obs with QualityMarker > 3
+         - filter: RejectList
+           apply at iterations: 0,1
+           filter variables:
+           - name: airTemperature
+           where:
+           - variable: QualityMarker/airTemperature
+             is_in: 4-15
+             value: is_not_valid
+           where operator: or
+           action:
+             name: reduce obs space
+
+         # Time window filter
+         - filter: Domain Check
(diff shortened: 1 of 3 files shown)


delippi:
This is the last of the Phase 1 jedi yamls (from prepbufr) that are marked as used by RRFS in the spreadsheet.

ShunLiu-NOAA (approved):

HuiLiu-NOAA:
Just wondering if this check would be enough applied only once before any outloops, that is, the line "apply at iterations: 0,1" could be removed? If so, this would save some runtime resources.  Same issue might also apply to some other filters in this yaml.

        # Time window filter
         - filter: Domain Check
           apply at iterations: 0,1
           where:
             - variable:
                 name: MetaData/timeOffset # units: s

delippi:
> Just wondering if this check would be enough applied only once before any outloops, that is, the line "apply at iterations: 0,1" could be removed? If so, this would save some runtime resources. Same issue might also apply to some other filters in this yaml.
> 
> ```
>     # Time window filter
>      - filter: Domain Check
>        apply at iterations: 0,1
>        where:
>          - variable:
>              name: MetaData/timeOffset # units: s
> ```

It might be. Have you tested how much time or resources it would save? My guess is it is negligible, but I don't know that for sure! Instead of changing it here. Let's test it and confirm if it saves time then make those changes in a follow up. This way if it needs changed, I can change it for all the yamls all at once! I have a few other things I need to do to fix some more consistency issues again anyway since I have multiple PRs open at once and things get switch around.

HuiLiu-NOAA:
> > Just wondering if this check would be enough applied only once before any outloops, that is, the line "apply at iterations: 0,1" could be removed? If so, this would save some runtime resources. Same issue might also apply to some other filters in this yaml.
> > ```
> >     # Time window filter
> >      - filter: Domain Check
> >        apply at iterations: 0,1
> >        where:
> >          - variable:
> >              name: MetaData/timeOffset # units: s
> > ```
> 
> It might be. Have you tested how much time or resources it would save? My guess is it is negligible, but I don't know that for sure! Instead of changing it here. Let's test it and confirm if it saves time then make those changes in a follow up. This way if it needs changed, I can change it for all the yamls all at once! I have a few other things I need to do to fix some more consistency issues again anyway since I have multiple PRs open at once and things get switch around.

Not sure the actual saving could be. Maybe for more clear logically, this check is only needed applied only once? It seems make no sense to apply this check for  more than 1 time.

delippi:
> > > Just wondering if this check would be enough applied only once before any outloops, that is, the line "apply at iterations: 0,1" could be removed? If so, this would save some runtime resources. Same issue might also apply to some other filters in this yaml.
> > > ```
> > >     # Time window filter
> > >      - filter: Domain Check
> > >        apply at iterations: 0,1
> > >        where:
> > >          - variable:
> > >              name: MetaData/timeOffset # units: s
> > > ```
> > 
> > 
> > It might be. Have you tested how much time or resources it would save? My guess is it is negligible, but I don't know that for sure! Instead of changing it here. Let's test it and confirm if it saves time then make those changes in a follow up. This way if it needs changed, I can change it for all the yamls all at once! I have a few other things I need to do to fix some more consistency issues again anyway since I have multiple PRs open at once and things get switch around.
> 
> Not sure the actual saving could be. Maybe for more clear logically, this check is only needed applied only once? It seems make no sense to apply this check for more than 1 time.

Yeah, I can likely safely remove this line from this filter and from a few other filters as well. I'll double check this in a later PR which can also allow me to check if there is any time savings. For now, let's keep it in and I will add this to my list of things to do to all yamls in a single PR.

Thanks for the suggestion. It is something I had considered before, but just for consistency reasons between all yamls I just never changed it.

SamuelDegelia-NOAA (approved):
Looks good to me. The only thing is that I think the title of the PR is a typo (it says sfcshp instead of adpupa).

delippi:
> Looks good to me. The only thing is that I think the title of the PR is a typo (it says sfcshp instead of adpupa).

HA! dang copy and paste. Thanks. I will fix it.

ShunLiu-NOAA:
@hu5970 and @guoqing-noaa Do you have a chance to review this PR? We will merge this PR by Friday.

PR #324, opened 2025-03-11
ADD ZTD bufr2ioda python tool and jedi yaml (updated)

HuiLiu-NOAA:
This PR adds a bufr2ioda python tool and a jedi yaml for ZTD obs to RDASApp.

    The ZTD bufr2ioda python tool converts RAP hourly gpsipw Bufr data into the ZTD IODA files for use in the RDASApp and RRFSv2 workflow.  The .json file is generic with analysis time only and works easier for all bufr2ioda python converters in the workflow.
    The initial ZTD "obs space" yaml is for assimilation of ZTD observations using the existing UKMO local operator only. More QC filters would be added later as evaluation continues.

  The yaml also has an option to do "monitoring only mode" of observations.


delippi:
@HuiLiu-NOAA I can't see where you had the jedi.yaml before in the old PR, but I think it should go in the `validated_yamls/` path. That name might be a little misleading--it doesn't really mean that those yamls are ready for cycling in MPAS-JEDI. I currently have Phase 1-3 yamls there and I note which phase each are on in that observation spread sheet. I think the main thing is that they should all just be in the same location regardless of the level of "validation" actually done. My point is that I don't think we need to have things scattered about. This type isn't going to follow the phased testing, so I think it is definitely fine. If it were another ob type that we use in GSI, I would say it needs to be at least Phase 1 before committing at all.

HuiLiu-NOAA:
The ZTD jedi yaml is added to the RDASApp ctests. The En3Dvar of FV3-JEDI and MPAS-JEDI were passed successfully. 

The yaml also is also updated with the option "monitoring obs" or "passivate mode".

HuiLiu-NOAA:
@SamuelDegelia-NOAA : Can you help to start the ctests for the ZTD jedi yaml from this PR? I did not find the clue/link here.

SamuelDegelia-NOAA:
@HuiLiu-NOAA To add the ZTD yaml to the ctest, you can modify `RDASApp/rrfs-test/validated_yamls/gen_yaml_ctest.sh` by adding your new yaml to the list of `obtype_configs`. Then you can run the ctests as normal by going into `RDASApp/build/rrfs-test` and running `ctest -j8`. 

We expect the ctests to fail the first time until the test references are updated. You can check why the ctests fail in `RDASApp/build/rrfs-test/Testing/Temporary/LastTest.log` and make sure the failures are along the lines of a "reference mismatch error". Then, you can copy the test output files to overwrite the reference files. For example: 

```
cp rundir-rrfs_fv3jedi_2024052700_Ens3Dvar/rrfs-fv3jedi-ens3dvar.out ../../rrfs-test/testoutput/rrfs-fv3jedi-ens3dvar.ref
cp rundir-rrfs_fv3jedi_2024052700_getkf_observer/rrfs-fv3jedi-getkf-observer.out ../../rrfs-test/testoutput/rrfs-fv3jedi-getkf-observer.ref
cp rundir-rrfs_fv3jedi_2024052700_getkf_solver/rrfs-fv3jedi-getkf-solver.out ../../rrfs-test/testoutput/rrfs-fv3jedi-getkf-solver.ref
cp rundir-rrfs_mpasjedi_2024052700_Ens3Dvar/rrfs-mpasjedi-ens3dvar.out ../../rrfs-test/testoutput/rrfs-mpasjedi-ens3dvar.ref
cp rundir-rrfs_mpasjedi_2024052700_getkf_observer/rrfs-mpasjedi-getkf-observer.out ../../rrfs-test/testoutput/rrfs-mpasjedi-getkf-observer.ref
cp rundir-rrfs_mpasjedi_2024052700_getkf_solver/rrfs-mpasjedi-getkf-solver.out ../../rrfs-test/testoutput/rrfs-mpasjedi-getkf-solver.ref
```

Finally, you should be able to rerun the ctests once more and expect them to pass. If so, you can then commit these updated test references files with this PR. We can then run the ctests on the other machines using the "labels" options on Github. 

I should probably update the wiki with this information now that the method for adding obs types to the ctests have changed.

HuiLiu-NOAA:
Thanks Sam!. It seems also need to copy the yamls generated by the gen_yaml_ctest.sh to the individual RDASApp/build/rrfs-test/rundir*/ manually?

SamuelDegelia-NOAA:
@HuiLiu-NOAA Oops, I forgot to mention that. After you add your ZTD yaml to `gen_yaml_ctest.sh`, you will need to run that script. And then you can run `RDASApp/rrfs-test/ush/update_ctest_inputs.sh` to copy the newly-generated super yamls into the run directories. 

We should probably integrate `gen_yaml_ctest.sh` into `update_ctest_inputs.sh` to make this into once step. 

Thanks for testing this process, by the way. This will help when I update the wiki instructions for how to do this more easily.

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
```diff
@@ -0,0 +1,100 @@
+     - obs space:
+         name: gnss_ztd_zenithTotalDelay
```
delippi, line 2: Can we match the obs space name with the yaml name? `gnss_zenithTotalDelay`. There isn't an obtype number associated with ztd, correct?

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
```diff
@@ -0,0 +1,100 @@
...
+             obsfile: "data/obs/ioda_ztd.nc"
+
+         obsgrouping:
+           group variables: ["stationIdentification"]
+           sort variable: "pressure"
+           sort order: "descending"
+         obsdataout:
+           engine:
+             type: H5File
+             obsfile: jdiag_ztd.nc
```
delippi: Can we also match the jdiag with the yaml name? `jdiag_gnss_zenithTotalDelay.nc`

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
```diff
@@ -0,0 +1,100 @@
...
+         name: gnss_ztd_zenithTotalDelay
+
+         distribution:   # for LETKF
+           name: "@DISTRIBUTION@"
+           halo size: 300e3
+
+         obsdatain:
+           engine:
+             type: H5File
+             obsfile: "data/obs/ioda_ztd.nc"
```
delippi: For the other conventional obs I've used the 6-character mnemonic, but this is from gnss. Would it make sense to use `ioda_gnss.nc`? I'm actually not sure what to suggest here and I'm just trying to keep things consistent. For example, ztd is the obtype and we don't have `ioda_airTemperature.nc`, we have `ioda_adpsfc.nc`. Are there other gnss ob types that would conflict with this suggestion?

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
HuiLiu-NOAA: > Can we match the obs space name with the yaml name? `gnss_zenithTotalDelay`. There isn't an obtype number associated with ztd, correct?

The workflow prepfer a short name for the included obs type list. So, it is kept here.

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
HuiLiu-NOAA: > For the other conventional obs I've used the 6-character mnemonic, but this is from gnss. Would it make sense to use `ioda_gnss.nc`? I'm actually not sure what to suggest here and I'm just trying to keep things consistent. For example, ztd is the obtype and we don't have `ioda_airTemperature.nc`, we have `ioda_adpsfc.nc`. Are there other gnss ob types that would conflict with this suggestion?

The "gnss" is too general for this specific yaml, which could include ztd, pw, RO obs type etc

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
HuiLiu-NOAA: > zenithTotalDelay

changed the name to "jdiag_gnss_ztd_zenithTotalDelay.nc

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
ShunLiu-NOAA: Could we use "gnssztd" or "gnssZtd" ? Do you think you can start a CTEST after adjusting naming convention?

HuiLiu-NOAA: Yes, we can use gnssztd. working on the ctest of letkf.

HuiLiu-NOAA:
@SamuelDegelia-NOAA: The ctest reference files are loaded and updated. 
Could you help to rename the following ioda files and then start related ctests?  Thanks. 

ioda_ztd.nc  -> ioda_gnss_ztd.nc 
ioda_ztd_dc.nc  -> ioda_gnss_ztd_dc.nc 
ioda_ztd_dc.png  -> ioda_gnss_ztd_dc.png

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
HuiLiu-NOAA: changed to ioda_gnss_ztd.nc

SamuelDegelia-NOAA:
Thanks for updating the ctests @HuiLiu-NOAA. I hope the process wasn't too annoying. The utilities added in #331 and the updates to the wiki page [here](https://github.com/NOAA-EMC/RDASApp/wiki/Updating-the-RDASApp-ctests-(adding-new-obs-or-updating-test-reference-files)) will hopefully make this easier in the future. 

I just updated the IODA file names across all of our supported machines. I'll start the CI ctests now.

HuiLiu-NOAA:
Thanks @SamuelDegelia-NOAA !!  
It is good to learn more about getkf observer and solver (e.g., which need to be run in the correct order)

SamuelDegelia-NOAA:
@HuiLiu-NOAA I just noticed that this PR did not add the ZTD yaml to `rrfs-test/validated_yamls/gen_yaml_ctest.sh`. Could you commit that change?

HuiLiu-NOAA:
> @HuiLiu-NOAA I just noticed that this PR did not add the ZTD yaml to `rrfs-test/validated_yamls/gen_yaml_ctest.sh`. Could you commit that change?

done. Thanks.

rrfs-test/validated_yamls/templates/obtype_config/gnss_zenithTotalDelay.yaml
```diff
@@ -0,0 +1,102 @@
...
+       - filter: Domain Check
+         where:
+           - variable:
+               name: MetaData/latitude
+             minvalue:  20.0
+             maxvalue:  60.0
+           - variable:
+               name: MetaData/longitude
+             minvalue: -135.0
+             maxvalue: -55.0
```
delippi: Can we double check these are correct for the current MPAS case? I'm finding
```
Max/Min Lat: 61.212459564208984, 19.301267623901367
Max/Min Lon: -48.77972412109375, -146.12791442871094
```
This check should be a box that is slightly larger than the domain itself. This will also change as the domain size changes. It seems our ctest case won't update but the size of the domain could become smaller soon. We should use the values for the ctest domain size. These values are also only for CONUS domain; we'll need to update again when we start working on NA domain. I suggest we use:
```
               name: MetaData/latitude
             minvalue:  19.0
             maxvalue:  62.0
           - variable:
               name: MetaData/longitude
             minvalue: -147.0
             maxvalue: -48.0
```

HuiLiu-NOAA: > Can we double check these are correct for the current MPAS case? I'm finding
> 
> ```
> Max/Min Lat: 61.212459564208984, 19.301267623901367
> Max/Min Lon: -48.77972412109375, -146.12791442871094
> ```
> 
> This check should be a box that is slightly larger than the domain itself. This will also change as the domain size changes. It seems our ctest case won't update but the size of the domain could become smaller soon. We should use the values for the ctest domain size. These values are also only for CONUS domain; we'll need to update again when we start working on NA domain. I suggest we use:
> 
> ```
>                name: MetaData/latitude
>              minvalue:  19.0
>              maxvalue:  62.0
>            - variable:
>                name: MetaData/longitude
>              minvalue: -147.0
>              maxvalue: -48.0
> ```

Thanks @delippi for piont out this!
 For the ZTD data, the current lat/lon ranges are good/large enough.  You can use your larger domain lat/lon in the other yamls to be updated.

delippi: I think it should to be updated even if it works for your current case. The point of the pre-online domain check is to get a rough estimate of the obs to be used. These bounds shouldn't be defined based on the data or observing network; it should be defined based on the domain. These bounds should never be smaller than the model domain. And it should be consistent across all yamls.

delippi: Also, do the longitude values need to be in [0, 360] format instead of [-180, 180]? I just tested and the [-180, 180] actually had zero observations...

Oh! wait, I know why this is. It is because we haven't been following a standard for these values in the ioda file. There are some ioda files with [0, 360] and some with [-180, 180]. This is a problem. We need to come to an agreement on that underlying problem and then apply that standard here. This way all yamls are the same.

HuiLiu-NOAA: @delippi : You are right. The lat/lon ranges are updated.  I do not expect any changes to the ctest references.  
Maybe these ranges can be argumented in future for other domains.  

For the (-180, 180) of lon, I think the related codes should be able to handle it?

delippi: I tested with my mesonet data, which is in the [0, 360] range. Using the values you provided removed every observation from the obs space at that filter. The domain check (and similar filters) aren't smart enough to recognize that MetaData/longitude could be in either format—they simply assume the data is already correctly formatted.

Some parts of JEDI can handle different longitude formats, but only when they explicitly expect longitude data. This is just a generic filter that can process any type of data, so it relies on the user to ensure proper formatting beforehand.

I think you're fine to include it in this PR. I'm just saying I'm not going to do it for all the prepbufr obs at the moment because we need to sort this out better first.

HuiLiu-NOAA: Just changed the lon range to (0, 360) range in the bufr2ioda python tool.

HuiLiu-NOAA: Also, updated the jedi yaml as below. Please double check if this works.

               name: MetaData/longitude
             minvalue: 213.0
             maxvalue: 312.0

delippi: Yep, these were what worked for my mesonet data when I tested it. Thanks for switching this already. None of this should affect your results, but it would be good to check with the ctests again. Since it looks like you were planning to add them to the ctests already. I guess you also have to update the staged data too?

SamuelDegelia-NOAA:
Looking at the ctest failure on Hercules, it looks similar to the reproducibility issues I document in #327. It is a small difference in the max analysis `cloud_liquid_water` values. Maybe we should consider increasing the ctest tolerance now that we are adding so many obs types. I'll look and see if that would allow the tests to pass on Hercules. 

```
terminate called after throwing an instance of 'oops::TestReferenceFloatMismatchError'
  what():  Test reference Float mismatch @ Line:1970
Test Val : 9.9343203786583919e-04
Ref  Val : 9.9069968655447738e-04
Delta    : 2.7323513113618105e-06
Relative tolerance: 9.9206586221015823e-07
Absolute tolerance: 9.9999999999999995e-07
Test Line: 'cloud_liquid_water                           | Min:+0.0000000000000000e+00 Max:+9.9343203786583919e-04 RMS:+1.5179189356771924e-05'
Ref Line : 'cloud_liquid_water                           | Min:+0.0000000000000000e+00 Max:+9.9069968655447738e-04 RMS:+1.5184603405441123e-05'
```

SamuelDegelia-NOAA:
@HuiLiu-NOAA Do the staged IODA files need to change for your updates to lat/long? I just tried running the ctests from your fork and they fail because there are no ZTD obs passing QC

HuiLiu-NOAA:
> @HuiLiu-NOAA Do the staged IODA files need to change for your updates to lat/long? I just tried running the ctests from your fork and they fail because there are no ZTD obs passing QC

I will send you the updated ioda files shortly. Thanks.

ShunLiu-NOAA (dismissed):

SamuelDegelia-NOAA:
@HuiLiu-NOAA It looks like the changes to the lat/long and the generated domain check files have slightly changed the ob counts (actually more obs now). I am seeing some ctest failures after this update: 

```
Test Line: 'gnss_ztd_zenithTotalDelay nobs= 648 Min=1.7270777266953357e+00, Max=2.5801955171543041e+00, RMS=2.3171327236157495e+00'
Ref Line : 'gnss_ztd_zenithTotalDelay nobs= 638 Min=1.7270777266953357e+00, Max=2.5801955171543041e+00, RMS=2.3163845745293292e+00'
```

Would you be able to regenerate the test references?

HuiLiu-NOAA:
> @HuiLiu-NOAA It looks like the changes to the lat/long and the generated domain check files have slightly changed the ob counts (actually more obs now). I am seeing some ctest failures after this update:
> 
> ```
> Test Line: 'gnss_ztd_zenithTotalDelay nobs= 648 Min=1.7270777266953357e+00, Max=2.5801955171543041e+00, RMS=2.3171327236157495e+00'
> Ref Line : 'gnss_ztd_zenithTotalDelay nobs= 638 Min=1.7270777266953357e+00, Max=2.5801955171543041e+00, RMS=2.3163845745293292e+00'
> ```
> 
> Would you be able to regenerate the test references?

OK. will do.

SamuelDegelia-NOAA:
Thanks @HuiLiu-NOAA, everything is passing again now except for `rrfs_fv3jedi_2024052700_getkf_solver` on hercules (as from before).

HuiLiu-NOAA:
> Thanks @HuiLiu-NOAA, everything is passing again now except for `rrfs_fv3jedi_2024052700_getkf_solver` on hercules (as from before).

Thanks @SamuelDegelia-NOAA !!

ShunLiu-NOAA:
Since we will use the different test tolerance on hercules, ctests passed on hera and jet, we merge this PR to develop.

ShunLiu-NOAA:
@HuiLiu-NOAA and @SamuelDegelia-NOAA could you please resolve the conflicts so that we can move forward.

HuiLiu-NOAA:
> @HuiLiu-NOAA and @SamuelDegelia-NOAA could you please resolve the conflicts so that we can move forward.

@SamuelDegelia-NOAA : do you know how to resolve the reference conflict? maybe remove the previous references?

SamuelDegelia-NOAA:
I just resolved the conflicts using the Github text editor, so it should be good to go now.

ShunLiu-NOAA (approved):

ShunLiu-NOAA:
Thank you, Sam.

Issue #282, opened 2025-02-14
Incorporate adpsfc DA into JEDI for RRFS


delippi:
## Description
The UFO in JEDI is the component that not only computes model-simulated observations, but also houses filters and methods for observation QC, ob errors, and bias correction. The GSI observer is the equivalent. Many forward operators for various observations have been developed for the UFO. These operators can be utilized in RDAS. However these operators still must be tested for RDAS. The steps for transition by observation platform are as follows:

1. Establish IODA processing
2. Establish operator in UFO
3. Establish YAML input file
4. Validate with test assimilation experiments
5. Create ctest, if necessary
6. adpsfc data includes bufr obtypes=*81

## Requirements
To create yaml configurations for assimilating adpsfc data.

## Acceptance Criteria (Definition of Done)
- [x] Phase 1 validation (hofx validation using GSI-IODA and GSI-geovals as necessary; GSI vs FV3-JEDI)
- [ ] Phase 2 validation (QC validation; no reliance on GSI except to be used as the baseline; GSI vs FV3-JEDI)
- [ ] Phase 3 validation (FV3-JEDI and GSI vs MPAS-JEDI)

- Link any relevant pull requests here:
  - 

## Dependencies
None

delippi:
**Phase 1 update: (for 181/281)**
Phase 1 results are in. These results were created using GSI-IODA, but allowing JEDI to handle the rest (geovals, QC, and oberror inflation). ObsErrorFactorConventional is included in these yamls, but it is turned off.

Uncertainties:

- For airTemperature: I had to use 1/2 obs error provided in errtable and commented out the ObsErrorPressureCheck to match ObsError in GSI.
- For stationPressure: Ob counts different a lot since JEDI rejects from the gross checks. When using GSI-QC, increments look really bad. The final result (with lower ob counts) allowing JEDI to handle the QC checks seems okay.


airTemperature
![Image](https://github.com/user-attachments/assets/ab177228-007e-4b72-8f80-6138d1a0be49)
![Image](https://github.com/user-attachments/assets/53340c55-f711-4ac3-b685-851b292e1497)

specificHumidity
![Image](https://github.com/user-attachments/assets/69fc49ec-f468-463f-b1e0-c46a0b3c22af)
![Image](https://github.com/user-attachments/assets/c7f5ee77-5011-4558-8b78-a81668a3b425)

stationPressure
[...]

Issue #268, opened 2025-02-03
Incorporate RASS DA into JEDI for RRFS


delippi:
## Description
The UFO in JEDI is the component that not only computes model-simulated observations, but also houses filters and methods for observation QC, ob errors, and bias correction. The GSI observer is the equivalent. Many forward operators for various observations have been developed for the UFO. These operators can be utilized in RDAS. However these operators still must be tested for RDAS. The steps for transition by observation platform are as follows:

1. Establish IODA processing
2. Establish operator in UFO
3. Establish YAML input file
4. Validate with test assimilation experiments
5. Create ctest, if necessary
6. rass data is bufr obtype=126

## Requirements
To create yaml configurations for assimilating RASS data.

## Acceptance Criteria (Definition of Done)
- [x] Phase 1 validation (hofx validation using GSI-IODA and GSI-geovals as necessary; GSI vs FV3-JEDI)
- [ ] Phase 2 validation (QC validation; no reliance on GSI except to be used as the baseline; GSI vs FV3-JEDI)
- [ ] Phase 3 validation (FV3-JEDI and GSI vs MPAS-JEDI)

- Link any relevant pull requests here:
  - 

## Dependencies
None

delippi:
**Phase 1 update:**
Phase 1 results are in. These results were created using GSI-IODA, but allowing JEDI to handle the rest (geovals, QC, and oberror inflation). This results also used ObsErrorFactorConventional (OEFC) which had been previously noted to have some issues for other ob platforms, but seems to be working fine here.

![Image](https://github.com/user-attachments/assets/6176b3d6-41f8-4877-a67c-b272dc913715)
![Image](https://github.com/user-attachments/assets/5bdea9ed-39f4-4606-a762-1ee9ecdfa621)

delippi:
**Phase 2/3 update**
Complete phase 3 testing with an updated RDASApp that can assimilate specificHumidity in MPAS-JEDI. All results will be in the following [slide deck](https://docs.google.com/presentation/d/1lcJMdWRKIbF_CSfjh4G7PtJY8L_bxenCyJhJt2dGL9w/edit?usp=sharing).

Issue #106, opened 2024-07-18
Incorporate Aircraft DA into JEDI for RRFS


delippi:
The UFO in JEDI is the component that not only computes model-simulated observations, but also houses filters and methods for observation QC, ob errors, and bias correction. The GSI observer is the equivalent. Many forward operators for various observations have been developed for the UFO. These operators can be utilized in RDAS. However these operators still must be tested for RDAS. The steps for transition by observation platform are as follows:

1. Establish IODA processing
2. Establish operator in UFO
3. Establish YAML input file
4. Validate with test assimilation experiments
5. Create ctest, if necessary

Aircraft data is bufr obtype=*30, *31, *33-*35

guoqing-noaa:
@delippi Thanks for doing this!

FYI, @spanNOAA submitted an ioda task to rrfs-workflow earlier this year https://github.com/NOAA-EMC/rrfs-workflow/pull/199.
Those yaml files may be re-used here:
https://github.com/NOAA-EMC/rrfs-workflow/tree/dev-sci/parm/iodaconv

delippi:
Hofx and oberrors for 130, 131, 134, and 135 (from AIRCFT) are validated. Followed the following documentation for what is assimilated in RAP: https://www.emc.ncep.noaa.gov/mmb/data_processing/prepbufr.doc/table_5.htm

The following tests were completed by feeding the GSI-geovals into vertInterp. There were a few places where the JEDI ob errors don't match the GSI ob errors, but that doesn't seem to be a worry. I didn't look into it beyond what I have shown here, but I would guess it is due to obs being closer to the surface relative to most of the other obs. I have previously shown that ob errors can be different when an ob is below the top 20 or so model levels. This is related to hybrid vcoord and pressure level geoval calculations.

ObType=130 T
![fv3jedi_vs_gsi_validation_130_airTemperature](https://github.com/user-attachments/assets/0e971878-d134-41fa-90fa-ad30d5dcbeff)

ObType=131 T
![fv3jedi_vs_gsi_validation_131_airTemperature](https://github.com/user-attachments/assets/a6988e62-0864-4405-a545-1f5203720dbd)

ObType=134 T
![fv3jedi_vs_gsi_validation_134_airTemperature](https://github.com/user-attachments/assets/9bd3de7b-9f1f-411b-9089-5976279c9ffd)

ObType=135 T
[...]

PR #326, opened 2025-03-14
Phase 2&3 validation: prepbufr

delippi:
## Description
Finished most of Phase 3 testing for obs from prepbufr. Phase 2&3 were combined similarly to other PRs before. All results can be found in this [slide deck](https://docs.google.com/presentation/d/1lcJMdWRKIbF_CSfjh4G7PtJY8L_bxenCyJhJt2dGL9w/edit?usp=sharing). After this PR, all prepbufr obs except for VAD winds will be ready for DA cycling in rrfs-workflow. Also note that there were several observation types without observations to test; those should be okay since the yaml structure follows that of all the other similar types.

## Issue(s) addressed
- https://github.com/NOAA-EMC/RDASApp/issues/80
- https://github.com/NOAA-EMC/RDASApp/issues/106
- https://github.com/NOAA-EMC/RDASApp/issues/232
- https://github.com/NOAA-EMC/RDASApp/issues/268
- https://github.com/NOAA-EMC/RDASApp/issues/270
- https://github.com/NOAA-EMC/RDASApp/issues/282
- https://github.com/NOAA-EMC/RDASApp/issues/294

- https://github.com/JCSDA-internal/ufo/pull/3645

## Checklist
- [x] Unit tests added/updated (if applicable).


HuiLiu-NOAA:
Any plan to add Ps 120 for validation and to the workflow?

delippi:
> Any plan to add Ps 120 for validation and to the workflow?

Yes, that is one of two tasks I still need to complete for prepbufr observations. I have already finished Phase 1 for Ps 120 using the GSI-IODA approach. At that stage, it should be nearly ready for integration into the workflow. The next step is to formally test it with the python bufr2ioda converter, which also needs to be incorporated into the workflow. It'll take just a little more effort to finalize that one.

HuiLiu-NOAA:
Thanks @delippi.  These look better to me.

Another thing is: 
 would it be worth to add the pre-online domain check to these yamls as well? It seems that this check can reduce runtime memory due to rejected OBS and run faster according to tests of Sam.  A few arguments of lat/lon range could be added for the workflow for various domain sizes.

delippi:
> Thanks @delippi. These look better to me.
> 
> Another thing is: would it be worth to add the pre-online domain check to these yamls as well? It seems that this check can reduce runtime memory due to rejected OBS and run faster according to tests of Sam. A few arguments of lat/lon range could be added for the workflow for various domain sizes.

Thanks for the suggestion @HuiLiu-NOAA. I will work on adding this to each yaml here. I have your lat/lon settings from your ztd yaml.

HuiLiu-NOAA:
Thanks @delippi! 
 Maybe also add the "monitoring obs" option as well?  I guess there are some obs types needed to be in the "monitoring" mode in the workflow.

delippi:
> Thanks @delippi! Maybe also add the "monitoring obs" option as well? I guess there are some obs types needed to be in the "monitoring" mode in the workflow.

Do you mean a block that is commented out, like this:
```
       # Monitoring mode (no assimilation, set QCflag=1)
#      - filter: Perform Action
#        filter variables:
#        - name: zenithTotalDelay
#        action:
#          name: passivate
```

HuiLiu-NOAA:
> > Thanks @delippi! Maybe also add the "monitoring obs" option as well? I guess there are some obs types needed to be in the "monitoring" mode in the workflow.
> 
> Do you mean a block that is commented out, like this:
> 
> ```
>        # Monitoring mode (no assimilation, set QCflag=1)
> #      - filter: Perform Action
> #        filter variables:
> #        - name: zenithTotalDelay
> #        action:
> #          name: passivate
> ```

Yes, users can uncomment these lines if needed.

SamuelDegelia-NOAA:
Since we are updating the yamls used in the ctest, we will need to run the ctests and update test references as part of this PR. I am working on updates to the wiki page to better describe how to do this. 

Also, since we are now able to use the OEFPC for the moisture yamls in mpas-jedi, we can remove the line that calls `python commentQC.py` in `gen_yaml_ctest.sh`.

SamuelDegelia-NOAA:
I updated the wiki page [here](https://github.com/NOAA-EMC/RDASApp/wiki/Updating-the-RDASApp-ctests-(adding-new-obs-or-updating-test-reference-files)) for updating the ctest output for a PR such as this. You can just skip to step 4 since you are not adding new observation types to the ctests. Let me know if there are any issues or if it is confusing.

delippi:
> I updated the wiki page [here](https://github.com/NOAA-EMC/RDASApp/wiki/Updating-the-RDASApp-ctests-(adding-new-obs-or-updating-test-reference-files)) for updating the ctest output for a PR such as this. You can just skip to step 4 since you are not adding new observation types to the ctests. Let me know if there are any issues or if it is confusing.

Thanks @SamuelDegelia-NOAA, I will work on this too.

HuiLiu-NOAA:
> I updated the wiki page [here](https://github.com/NOAA-EMC/RDASApp/wiki/Updating-the-RDASApp-ctests-(adding-new-obs-or-updating-test-reference-files)) for updating the ctest output for a PR such as this. You can just skip to step 4 since you are not adding new observation types to the ctests. Let me know if there are any issues or if it is confusing.

Thanks to @SamiraArdani-NOAA. The instructions indeed look more clear!

HuiLiu-NOAA:
> I updated the wiki page [here](https://github.com/NOAA-EMC/RDASApp/wiki/Updating-the-RDASApp-ctests-(adding-new-obs-or-updating-test-reference-files)) for updating the ctest output for a PR such as this. You can just skip to step 4 since you are not adding new observation types to the ctests. Let me know if there are any issues or if it is confusing.

Thanks to @SamuelDegelia-NOAA!  These instructions do look more clear to me.

delippi:
> > Thanks @delippi. These look better to me.
> > Another thing is: would it be worth to add the pre-online domain check to these yamls as well? It seems that this check can reduce runtime memory due to rejected OBS and run faster according to tests of Sam. A few arguments of lat/lon range could be added for the workflow for various domain sizes.
> 
> Thanks for the suggestion @HuiLiu-NOAA. I will work on adding this to each yaml here. I have your lat/lon settings from your ztd yaml.

@HuiLiu-NOAA After some testing, I think I'm not going to include the pre-online domain check in this PR. We have some things to sort out first. We don't have a standard for the range of longitude values in the ioda file. Some ioda files use [-180, 180] and some are [0, 360]. I think once we fix that we can certainly add this functionality into the yamls.

delippi:
> > > Thanks @delippi! Maybe also add the "monitoring obs" option as well? I guess there are some obs types needed to be in the "monitoring" mode in the workflow.
> > 
> > 
> > Do you mean a block that is commented out, like this:
> > ```
> >        # Monitoring mode (no assimilation, set QCflag=1)
> > #      - filter: Perform Action
> > #        filter variables:
> > #        - name: zenithTotalDelay
> > #        action:
> > #          name: passivate
> > ```
> 
> Yes, users can uncomment these lines if needed.

@HuiLiu-NOAA, this was another good suggestion. Since I'm holding off on adding the pre-online domain check, I think I will hold off on this one too. We can always add these in later. I know the yamls as-is are working as expected and there are a lot to edit and a lot of room for mistakes. It is best to do this in a few steps.

HuiLiu-NOAA:
> > > Thanks @delippi. These look better to me.
> > > Another thing is: would it be worth to add the pre-online domain check to these yamls as well? It seems that this check can reduce runtime memory due to rejected OBS and run faster according to tests of Sam. A few arguments of lat/lon range could be added for the workflow for various domain sizes.
> > 
> > 
> > Thanks for the suggestion @HuiLiu-NOAA. I will work on adding this to each yaml here. I have your lat/lon settings from your ztd yaml.
> 
> @HuiLiu-NOAA After some testing, I think I'm not going to include the pre-online domain check in this PR. We have some things to sort out first. We don't have a standard for the range of longitude values in the ioda file. Some ioda files use [-180, 180] and some are [0, 360]. I think once we fix that we can certainly add this functionality into the yamls.

@delippi :  To make things easier, I think I can change  the lon to (0, 360) in the ioda_gnss_ztd.dat.  what do you think?  I think to add the pre-online domain check and the "passice mode" oprion would be of more interest.

delippi:
> > > > Thanks @delippi. These look better to me.
> > > > Another thing is: would it be worth to add the pre-online domain check to these yamls as well? It seems that this check can reduce runtime memory due to rejected OBS and run faster according to tests of Sam. A few arguments of lat/lon range could be added for the workflow for various domain sizes.
> > > 
> > > 
> > > Thanks for the suggestion @HuiLiu-NOAA. I will work on adding this to each yaml here. I have your lat/lon settings from your ztd yaml.
> > 
> > 
> > @HuiLiu-NOAA After some testing, I think I'm not going to include the pre-online domain check in this PR. We have some things to sort out first. We don't have a standard for the range of longitude values in the ioda file. Some ioda files use [-180, 180] and some are [0, 360]. I think once we fix that we can certainly add this functionality into the yamls.
> 
> @delippi : To make things easier, I think I can change the lon to (0, 360) in the ioda_gnss_ztd.dat. what do you think? I think to add the pre-online domain check and the "passice mode" oprion would be of more interest.

Well, I need to do a deep dive and see how it is defined for all the types. Off the top of my head, I'm thinking most (if not all) I had tried to use [0, 360]. I was going to just suggest switch to whichever was most used at this point. I am okay with you changing the ztd to this format if you want.

I think these are great options to have, <ins>and we will get them in!</ins> I'm just thinking it would be better to do this in a few steps/PRs. If it were a handful of yamls it would be fine to do here, but I would like to test each one by one and make sure there are no issues. I think it'll be easier to do in another PR solely focused on such additions; it isn't good practice to have too many things in a single PR.

delippi:
> I updated the wiki page [here](https://github.com/NOAA-EMC/RDASApp/wiki/Updating-the-RDASApp-ctests-(adding-new-obs-or-updating-test-reference-files)) for updating the ctest output for a PR such as this. You can just skip to step 4 since you are not adding new observation types to the ctests. Let me know if there are any issues or if it is confusing.

Thanks @SamuelDegelia-NOAA! The only thing I don't know about is if I should try adding all the observations types into the ctest? You had several issues. Do you think I should just use the types you already confirmed were working right. Then we add the rest later? What is your suggestion?

delippi:
I converted this to a draft since we need the dependency to be completed first. The change to how qsat is requested would be a problem if this is merged before https://github.com/JCSDA-internal/ufo/pull/3645.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_181.yaml
```diff
@@ -108,8 +105,15 @@
...
-                 #errors: [1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056]
+                 errors: [1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056, 1.5056]
+
+         # Reduce observation error and turn off ObsErrorFactorPressureCheck (l_gsd_terrain_match_surfTobs=.true., l_sfcobserror_ramp_t=.true.,)
+         - filter: Perform Action
+           filter variables:
+           - name: airTemperature
+           action:
+             name: inflate error
+             inflation factor: 0.5
```
guoqing-noaa: Line 110-116: This part is turned on when we use the GSL surface operator that @spanNOAA has been working on. It is expected this section should be set along with the GSL surface operator.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_187.yaml
```diff
@@ -107,24 +104,32 @@
...
-             inflation variable:
-               name: ObsFunction/ObsErrorFactorPressureCheck
-               options:
-                 variable: airTemperature
-                 inflation factor: 8.0
-                 # The new feature "surface observation error ramp"
-                 # needs to be added to UFO to align with the ramp
-                 # options for temperature and humidity in GSI.
-                 #surface observation error ramp: true
+             inflation factor: 0.5
```
guoqing-noaa: Similar to the above, Line 107-113 is expected to be turned off (commented out) by default.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_specificHumidity_181.yaml
```diff
@@ -124,7 +121,7 @@
                options:
                  variable: specificHumidity
                  inflation factor: 8.0
-                 request_saturation_specific_humidity_geovals: true
+                 request_saturation_specific_humidity_geovals: false
```
guoqing-noaa: I need to learn more about this. But does this change mean we can do `ObsErrorFactorPressureCheck`  without the qsat?

rrfs-test/validated_yamls/templates/obtype_config/sfcshp_stationPressure_180.yaml
```diff
@@ -171,17 +168,17 @@
            test variables:
            - name:  ObsErrorData/stationPressure
            minvalue: 0.0
-           action:
-             name: reduce obs space
+           #action:
+           #  name: reduce obs space
```
guoqing-noaa: Why do we comment out the `reduce obs space` here?

rrfs-test/validated_yamls/templates/obtype_config/sfcshp_stationPressure_180.yaml
```diff
@@ -171,17 +168,17 @@
...
          # Bounds Check (ObsValue)
          - filter: Bounds Check
            filter variables:
            - name: stationPressure
            minvalue: 20000.0
            maxvalue: 120000.0
-           action:
-             name: reduce obs space
+           #action:
+           #  name: reduce obs space
```
guoqing-noaa: Why do we comment out the `reduce obs space` here?

rrfs-test/validated_yamls/templates/obtype_config/sfcshp_winds_280.yaml
```diff
@@ -278,7 +231,6 @@
            maxvalue: 130
            action:
              name: reject
```
guoqing-noaa, line 233: Line 233 can be removed. We have two `action.name` keys here, and JEDI will only use the last one, i.e. `name: reduce obs space` and it includes an implicit "reject" action. To avoid confusion and misunderstanding, I would suggest we remove the no-longer-needed `name: reject` for all duplicate `action.name` keys.

I created an issue https://github.com/NOAA-EMC/RDASApp/issues/348 and conducted numerous experiments and posted my results there. We can continue further discussions there. Thanks!

rrfs-test/validated_yamls/templates/obtype_config/sfcshp_winds_280.yaml
```diff
@@ -278,7 +231,6 @@
...
-           action:
              name: reduce obs space
```
guoqing-noaa, line 235: Line 233 (right side) can be removed. We have two `action.name` keys here, and JEDI will only use the last one, i.e. `name: reduce obs space` and it includes an implicit "reject" action. To avoid confusion and misunderstanding, I would suggest we remove the no-longer-needed `name: reject` for all duplicate `action.name` keys.

I created an issue https://github.com/NOAA-EMC/RDASApp/issues/348 and conducted some experiments and posted my results in that issue. We can continue further discussions there. Thanks!

In case you may need it, I used
```
sed -i '/name: reject/{N;/\n *name: reduce obs space$/s/.*\n//;}'   <file_name>
```
to remove all duplicate `action.name: reject`

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_181.yaml
delippi: I see, we only want to do this if we're also using that operator at the same time. What is the time estimate on the GSL surface operator getting in to ufo? I will turn this part off for now (and turn back on the OEFPC).

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_specificHumidity_181.yaml
delippi: No, we still need qsat. The `request_saturation_specific_humidity_geovals` is to get qsat from the <ins>geovals</ins> space. Hui's PR simply calculates the qsat within that obsfunction. So for FV3-based runs you can run either way. For MPAS-based only set to false will work properly.

rrfs-test/validated_yamls/templates/obtype_config/sfcshp_stationPressure_180.yaml
delippi: I might've just forgot to uncomment these when I was doing some testing. I will add them back in.

rrfs-test/validated_yamls/templates/obtype_config/sfcshp_stationPressure_180.yaml
delippi: same as above.

rrfs-test/validated_yamls/templates/obtype_config/sfcshp_winds_280.yaml
delippi: Thanks for catching this instance. I already removed the duplicate action key in several of the other yamls in this PR. I will do the same here.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_181.yaml
guoqing-noaa: Sounds good. Thanks!

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_specificHumidity_181.yaml
guoqing-noaa: Thanks for the explanation. So in MPASJEDI, qsat does NOT come from geovals, right (and FV3JEDI can do that as well)?

HuiLiu-NOAA: > Thanks for the explanation. So in MPASJEDI, qsat does NOT come from geovals, right (and FV3JEDI can do that as well)?

You are right!

delippi: Yes, that is correct @guoqing-noaa. This is also precisely the reason why this PR is in draft mode. We are waiting for Hui's PR to be merged which enables both MPAS- or FV3-based runs to run without requesting qsat from geovals.

guoqing-noaa: Thank all for the confirmation!

As we know, PR to the JCSDA repos will take a while to get merged. Also, the new PR only applies to the latest JEDI, which RDASApp lags behind. It will take more time to update the submodules of RDASApp. All these will not happen in a short time frame.

So I would suggest we revert this setting to 'true' (make it work for fv3jedi and we comment them out for MPASJEDI) and get this PR merged. We can change it later once Hui's PR gets merged.

ShunLiu-NOAA: As RRFS DA meeting this morning, the QSAT PR will be merged today or tomorrow.

delippi: I'm hopeful we are almost there on the PR in ufo--it has been under review for some time and there has been a lot of discussion already.

We _could_ revert them back to true, but I would highly advise against commenting (turning off) the OEFPC section for MPAS-JEDI. If we go this route, we should simply not use specificHumidity obs <ins>at all</ins> in the rrfs-workflow. The qsat variable is required to convert ObsError units from RH to q.

However, I don't think it is more urgent to merge this PR than to merge Hui's PR. Hui's is most urgent right now since we absolutely cannot use humidity obs at all until it is merged. Until then, we can make some easy work arounds. One of which would be what you've already done--maintain a useable parm/jedi.yaml based on this PR (excluding humidity obs). We will have to do this anyway to make sure that all the retros start with the exact same jedi.yaml.

HuiLiu-NOAA:
@guoqing-noaa : if you are ok, please approve the UFO QSAT PR.

HuiLiu-NOAA:
> @guoqing-noaa : if you are ok, please approve the UFO QSAT PR.

maybe also @ShunLiu-NOAA .

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_specificHumidity_181.yaml
guoqing-noaa: `The qsat variable is required to convert ObsError units from RH to q`

This is an important message. Thanks!

guoqing-noaa: In terms of the priority, we would hope the q DA capability has already been there last week.

As Ming (@hu5970) showed, assimilating t133 has positive impact on T, neutral impact on wind, negative impact on Q. So we want to assimilate q133 correctly as soon as possible to see whether that will address the issue.

HuiLiu-NOAA: > In terms of the priority, we would hope the q DA capability has already been there last week.
> 
> As Ming (@hu5970) showed, assimilating t133 has positive impact on T, neutral impact on wind, negative impact on Q. So we want to assimilate q133 correctly as soon as possible to see whether that will address the issue.

Agree that q133 (and q120) are important. Anyone has looked at the O-B of these from the retro?

delippi: > The qsat variable is required to convert ObsError units from RH to q

@guoqing-noaa, sorry that I didn't effectively communicate this to everyone earlier.

> In terms of the priority, we would hope the q DA capability has already been there last week.

@guoqing-noaa, we can technically run with a simple work around. Hui's change is only to a single file, so it is pretty easy to replace and recompile. This is what I did for my sample retro test assimilating all prepbufr obs. But we really just need to try to push this to get merged in ufo.

 > Agree that q133 (and q120) are important. Anyone has looked at the O-B of these from the retro?

@HuiLiu-NOAA, I ran a sample retro with all prepbufr observations including humidity obs (using your qsat fix and my obs space yamls in this PR). I have some initial stats on O-B for these obs in the retro here: https://www.emc.ncep.noaa.gov/mmb/dlippi/rrfs-workflow_v2/DA_monitoring/hrly_12km/.

HuiLiu-NOAA:
I think ps120 is also important for the retro workflow. Any plan/progress to get this type in the workflow soon?

delippi:
> I think ps120 is also important for the retro workflow. Any plan/progress to get this type in the workflow soon?

Yes. It is important data. I just need to do it. On the flip side, I don't expect any changes to the yaml from Phase 1. The biggest thing will be to implement the bufr2ioda for this in the workflow. We can only use the python converter for this data.

HuiLiu-NOAA:
> > I think ps120 is also important for the retro workflow. Any plan/progress to get this type in the workflow soon?
> 
> Yes. It is important data. I just need to do it. On the flip side, I don't expect any changes to the yaml from Phase 1. The biggest thing will be to implement the bufr2ioda for this in the workflow. We can only use the python converter for this data.

I think the python tool can be easily added into the workflow, at least method 1, just a few lines of updates?

guoqing-noaa:
> I think ps120 is also important for the retro workflow. Any plan/progress to get this type in the workflow soon?

@HuiLiu-NOAA Are you aware of any research/literature showing the impact of ps120 on regional models? Thanks!

guoqing-noaa:
> > > I think ps120 is also important for the retro workflow. Any plan/progress to get this type in the workflow soon?
> > 
> > 
> > Yes. It is important data. I just need to do it. On the flip side, I don't expect any changes to the yaml from Phase 1. The biggest thing will be to implement the bufr2ioda for this in the workflow. We can only use the python converter for this data.
> 
> I think the python tool can be easily added into the workflow, at least method 1, just a few lines of updates?

@HuiLiu-NOAA If you are familiar with that, could you create a PR to rrfs-workflow to get it added? Thanks!

HuiLiu-NOAA:
> > > > I think ps120 is also important for the retro workflow. Any plan/progress to get this type in the workflow soon?
> > > 
> > > 
> > > Yes. It is important data. I just need to do it. On the flip side, I don't expect any changes to the yaml from Phase 1. The biggest thing will be to implement the bufr2ioda for this in the workflow. We can only use the python converter for this data.
> > 
> > 
> > I think the python tool can be easily added into the workflow, at least method 1, just a few lines of updates?
> 
> @HuiLiu-NOAA If you are familiar with that, could you create a PR to rrfs-workflow to get it added? Thanks!

I think, at least for global models, ps120 is an important obs type. GSL has done some work before. Also, ps obs is important for ZTD assimilation as well (as a anchoring obs).

I will add the ps120 python tool into the workflow and open a PR (similar to what I have done for ZTD python converter) shortly.

guoqing-noaa:
> 
> I think, at least for global models, ps120 is an important obs type. GSL has done some work before. Also, ps obs is important for ZTD assimilation as well.
> 
> I will add the ps120 python tool into the workflow and open a PR (similar to what I have done for ZTD python converter) shortly.

Thanks, @HuiLiu-NOAA.  It is good to have that capability in rrfs-workflow.

Yes, for global models, ps is very important.  But in my limited knowledge, ps is a diagnostic variable in regional models, and hence ps data does not have a large impact as T, Q, and UV. I would be happy to read any publications/presentations investigating the ps120 impact on regional models to update my understanding.

HuiLiu-NOAA:
> > I think, at least for global models, ps120 is an important obs type. GSL has done some work before. Also, ps obs is important for ZTD assimilation as well.
> > I will add the ps120 python tool into the workflow and open a PR (similar to what I have done for ZTD python converter) shortly.
> 
> Thanks, @HuiLiu-NOAA. It is good to have that capability in rrfs-workflow.
> 
> Yes, for global models, ps is very important. But in my limited knowledge, ps is a diagnostic variable in regional models, and hence ps data does not have a large impact as T, Q, and UV. I would be happy to read any publications/presentations investigating the ps120 impact on regional models to update my understanding.

Thanks @guoqing-noaa for the info!  
 ZTD data is a combination of ps + moisture info in the atmosphere. I guess assimilating ps obs together with ZTD data would help to retrieve/separate the moisture info from ZTD data, which would not completely rely on the background error covariance. But this is just a hypothesis at the moment. It would be great results if we can prove this idea.

guoqing-noaa:
> > > I think, at least for global models, ps120 is an important obs type. GSL has done some work before. Also, ps obs is important for ZTD assimilation as well.
> > > I will add the ps120 python tool into the workflow and open a PR (similar to what I have done for ZTD python converter) shortly.
> > 
> > 
> > Thanks, @HuiLiu-NOAA. It is good to have that capability in rrfs-workflow.
> > Yes, for global models, ps is very important. But in my limited knowledge, ps is a diagnostic variable in regional models, and hence ps data does not have a large impact as T, Q, and UV. I would be happy to read any publications/presentations investigating the ps120 impact on regional models to update my understanding.
> 
> Thanks @guoqing-noaa for the info! ZTD data is a combination of ps + moisture info in the atmosphere. I guess assimilating ps obs together with ZTD data would help to retrieve/separate the moisture info from ZTD data, which would not completely rely on the background error covariance. But this is just a hypothesis at the moment. It would be great results if we can prove this idea.

Good idea! 👍

HuiLiu-NOAA:
@delippi : where are the python tool and phase 1 yaml for ps120 you are currently testing?  I would like to take a look of them and try to get them into the workflow. Thanks!

delippi:
> @delippi : where are the python tool and phase 1 yaml for ps120 you are currently testing? I would like to take a look of them and try to get them into the workflow. Thanks!

I haven't done anything with the python converter for the ps120 data which is why I haven't done the phase 2/3 tests yet. I'm not quite as familiar with the python versions. 

Are you talking about the obs space yaml? You can use the yaml in this PR (it is still the Phase 1 with a few extra changes; `adpupa_stationPressure_120.yaml`). What do you mean the obs space yaml to the workflow? Perhaps you're referring to the `GEN_YAML_METHOD` 1 or 3? There wouldn't be anything to add for option 2. For option 2, you just need to add it to the gen yaml list (and it should already be there commented out) and generate the super yaml before running the workflow. Then, just make sure the ioda file that it expects exists.

One thing to consider is how to handle the fact that we will have the yaml converter and python converter making a `ioda_adpupa.nc` file. Perhaps, we can move all adpupa data to completely python based converter??

HuiLiu-NOAA:
Thanks @delippi. 

Let's start with the python converter for ps120 first. The yaml converter did not work for ps120 as I tested weeks ago.

ShunLiu-NOAA:
Let me check with JCSDA to see if they merge Hui's PR ASAP.

delippi:
@HuiLiu-NOAA I was thinking about adding the passivate option in this PR as the very first filter for each obs space. What do you think about this example? This works for everything except mesonet winds since it uses accept lists and such. If we add this to the top of each obs space and set the default to accept it doesn't change anything. Then instead of having to uncomment an entire section to turn on or off, you can just change the action key. Or would you prefer this be at the very end of the obs space? I kind of like it at the beginning of the obs space... That makes the most logical sense. @SamuelDegelia-NOAA do you have any preference?
```
       obs filters:
         # ------------------
         # airTemperature (181)
         # ------------------
         # Accept, reject, or passivate (monitor)
         - filter: Perform Action
           filter variables:
           - name: airTemperature
           action:
             name: accept # accept, reject, passivate

```

HuiLiu-NOAA:
> @HuiLiu-NOAA I was thinking about adding the passivate option in this PR as the very first filter for each obs space. What do you think about this example? This works for everything except mesonet winds since it uses accept lists and such. If we add this to the top of each obs space and set the default to accept it doesn't change anything. Then instead of having to uncomment an entire section to turn on or off, you can just change the action key. Or would you prefer this be at the very end of the obs space? I kind of like it at the beginning of the obs space... That makes the most logical sense. @SamuelDegelia-NOAA do you have any preference?
> 
> ```
>        obs filters:
>          # ------------------
>          # airTemperature (181)
>          # ------------------
>          # Accept, reject, or passivate (monitor)
>          - filter: Perform Action
>            filter variables:
>            - name: airTemperature
>            action:
>              name: accept # accept, reject, passivate
> 
>          # Reject all obs with QualityMarker > 3
> ```

Thanks @delippi. I like this exsample.

guoqing-noaa:
> @HuiLiu-NOAA I was thinking about adding the passivate option in this PR as the very first filter for each obs space. What do you think about this example? This works for everything except mesonet winds since it uses accept lists and such. If we add this to the top of each obs space and set the default to accept it doesn't change anything. Then instead of having to uncomment an entire section to turn on or off, you can just change the action key. Or would you prefer this be at the very end of the obs space? I kind of like it at the beginning of the obs space... That makes the most logical sense. @SamuelDegelia-NOAA do you have any preference?
> 
> ```
>        obs filters:
>          # ------------------
>          # airTemperature (181)
>          # ------------------
>          # Accept, reject, or passivate (monitor)
>          - filter: Perform Action
>            filter variables:
>            - name: airTemperature
>            action:
>              name: accept # accept, reject, passivate
> 
>          # Reject all obs with QualityMarker > 3
> ```

@delippi I would suggest we do that in a separate PR.  It is good that we get this PR merged first and add other capabilities in other PRs.

We should avoid adding unnecessary "empty" filters. We can add the `passivate` part on the fly, or if you want to have a placeholder, let's comment it out by default.

delippi:
> > @HuiLiu-NOAA I was thinking about adding the passivate option in this PR as the very first filter for each obs space. What do you think about this example? This works for everything except mesonet winds since it uses accept lists and such. If we add this to the top of each obs space and set the default to accept it doesn't change anything. Then instead of having to uncomment an entire section to turn on or off, you can just change the action key. Or would you prefer this be at the very end of the obs space? I kind of like it at the beginning of the obs space... That makes the most logical sense. @SamuelDegelia-NOAA do you have any preference?
> > ```
> >        obs filters:
> >          # ------------------
> >          # airTemperature (181)
> >          # ------------------
> >          # Accept, reject, or passivate (monitor)
> >          - filter: Perform Action
> >            filter variables:
> >            - name: airTemperature
> >            action:
> >              name: accept # accept, reject, passivate
> > 
> >          # Reject all obs with QualityMarker > 3
> > ```
> 
> @delippi I would suggest we do that in a separate PR. It is good that we get this PR merged first and add other capabilities in other PRs.
> 
> We should avoid adding unnecessary "empty" filters. We can add the `passivate` part on the fly, or if you want to have a placeholder, let's comment it out by default.

@guoqing-noaa I think the problem with that is we shouldn't be modifying the yamls from cycle-to-cycle except for dates. The yaml should be mostly fixed. I wouldn't say it is an unnecessary filter. I was going to build a jedivar.yaml with all the obs and turn everything to passivate except the ones that we want to actually assimilate. That way the assimilation still creates all the jdiags (or at least I hope it does) so they can be processed by the DA monitoring tools and that makes it easier to compare retro runs when adding more observations in (at least for things like the heatmaps). Furthermore, there will only be a few line changes between jedivar.yaml for baseline1 and baseline2, for example.

guoqing-noaa:
> 
> @guoqing-noaa I think the problem with that is we shouldn't be modifying the yamls from cycle-to-cycle except for dates. The yaml should be mostly fixed. I wouldn't say it is an unnecessary filter. I was going to build a jedivar.yaml with all the obs and turn everything to passivate except the ones that we want to actually assimilate. That way the assimilation still creates all the jdiags (or at least I hope it does) so they can be processed by the DA monitoring tools and that makes it easier to compare retro runs when adding more observations in (at least for things like the heatmaps). Furthermore, there will only be a few line changes between jedivar.yaml for baseline1 and baseline2, for example.


The jedivar.yaml file will not change from cycle to cycle, but is different among different retro/relatime experiments. 

I am fine with you adding this filter block either in the comment or uncomment mode, but to be done in another separate PR.

I have a few minor comments on details:
1. I think we don't need a `reject` action for this filter, right?
2. Let's make the comment for this PR more explicit, something like:    
   `# Data Monitoring or Not`
3. It is good to comment out this filter by default, but it is acceptable to me if you want to uncomment and use the `accept` default action. 

So it may look like this:
```
       obs filters:
         # ------------------
         # airTemperature (181)
         # ------------------
         # Data Monitoring or Not
#         - filter: Perform Action
#           filter variables:
#           - name: airTemperature
#           action:
#             name: passivate # accept

....
```

guoqing-noaa:
Also, we may only want to monitor the data passed the pre-filters (and the regular filter, post filter?)

Will the location of the `passivate` filter affect the final results (such as at the beginning of filters or at the end of filters)? It will be good to do some tests to confirm the behaviors.

delippi:
> Also, we may only want to monitor the data passed the pre-filters (and the regular filter, post filter?)
> 
> Will the location of the `passivate` filter affect the final results (such as at the beginning of filters or at the end of filters)? It will be good to do some tests to confirm the behaviors.

I did some of these tests. Think of it this way. If we have an ob that we want to monitor this is how it will go. The first filter will be to passivate and set everything to monitor. All the filters after that will also activate and only reject some of those obs that were monitored. The remaining obs would have been assimilated but are still marked passive. The only exception is the mesonet winds since they first reject all obs, then use accept lists to accept obs back in. If you were to put passivate at the beginning then it basically wouldn't do anything. If you put it after the accept lists, it will just change all the accepted back to passive. Does that make sense? Basically each filter gets applied on top of the previous filters in the order within the yaml.

delippi:
> > Also, we may only want to monitor the data passed the pre-filters (and the regular filter, post filter?)
> > Will the location of the `passivate` filter affect the final results (such as at the beginning of filters or at the end of filters)? It will be good to do some tests to confirm the behaviors.
> 
> I did some of these tests. Think of it this way. If we have an ob that we want to monitor this is how it will go. The first filter will be to passivate and set everything to monitor. All the filters after that will also activate and only reject some of those obs that were monitored. The remaining obs would have been assimilated but are still marked passive. The only exception is the mesonet winds since they first reject all obs, then use accept lists to accept obs back in. If you were to put passivate at the beginning then it basically wouldn't do anything. If you put it after the accept lists, it will just change all the accepted back to passive. Does that make sense? Basically each filter gets applied on top of the previous filters in the order within the yaml.

So this means to monitor correctly it <ins>MUST</ins> be at the beginning of the filter list. @HuiLiu-NOAA

HuiLiu-NOAA:
> > > Also, we may only want to monitor the data passed the pre-filters (and the regular filter, post filter?)
> > > Will the location of the `passivate` filter affect the final results (such as at the beginning of filters or at the end of filters)? It will be good to do some tests to confirm the behaviors.
> > 
> > 
> > I did some of these tests. Think of it this way. If we have an ob that we want to monitor this is how it will go. The first filter will be to passivate and set everything to monitor. All the filters after that will also activate and only reject some of those obs that were monitored. The remaining obs would have been assimilated but are still marked passive. The only exception is the mesonet winds since they first reject all obs, then use accept lists to accept obs back in. If you were to put passivate at the beginning then it basically wouldn't do anything. If you put it after the accept lists, it will just change all the accepted back to passive. Does that make sense? Basically each filter gets applied on top of the previous filters in the order within the yaml.
> 
> So this means to monitor correctly it MUST be at the beginning of the filter list. @HuiLiu-NOAA

Thanks for this info.  
For the ZTD yaml, it seems ok/working to have the "passive" section in the middle of the filters (double checked the QC flags).

delippi:
@guoqing-noaa 
> The jedivar.yaml file will not change from cycle to cycle, but is different among different retro/relatime experiments.

How does YAML_GEN_METHOD=1 & 3 work when you have missing ioda files?

> I am fine with you adding this filter block either in the comment or uncomment mode, but to be done in another separate PR.

For the retros, I wanted to have a jedi.yaml that contains all obs (at least all prepbufr obs) with everything set to passivate except the obs we are using. For the first retro that will be airTemperature_133. I can add it in another PR, but I'll still likely build it like this for the retros.

delippi:
> Thanks for this info.
> For the ZTD yaml, it seems ok/working to have the "passive" section in the middle of the filters (double checked the QC flags).

You have the following order. You reject anything with QM > 10, then assign some error, then you passivate everything (including the previously rejected QM > 10 what you turn on that section). The passivate should be at the top (or very least before any rejection).
```
       # Reject all obs with QualityMarker > 10
       - filter: Bounds Check
         filter variables:
         - name: zenithTotalDelay
         test variables:
         - name: QualityMarker/zenithTotalDelay
         maxvalue: 10

       # Initial error assignment
       - filter: Perform Action
         filter variables:
         - name: zenithTotalDelay
         action:
           name: assign error
           error parameter: 0.01

       # Monitoring mode (no assimilation, set QCflag=1)
#      - filter: Perform Action
#        filter variables:
#        - name: zenithTotalDelay
#        action:
#          name: passivate
```

guoqing-noaa:
> > Thanks for this info.
> > For the ZTD yaml, it seems ok/working to have the "passive" section in the middle of the filters (double checked the QC flags).
> 
> You have the following order. You reject anything with QM > 10, then assign some error, then you passivate everything (including the previously rejected QM > 10 what you turn on that section). The passivate should be at the top (or very least before any rejection).
> 
> ```
>        # Reject all obs with QualityMarker > 10
>        - filter: Bounds Check
>          filter variables:
>          - name: zenithTotalDelay
>          test variables:
>          - name: QualityMarker/zenithTotalDelay
>          maxvalue: 10
> 
>        # Initial error assignment
>        - filter: Perform Action
>          filter variables:
>          - name: zenithTotalDelay
>          action:
>            name: assign error
>            error parameter: 0.01
> 
>        # Monitoring mode (no assimilation, set QCflag=1)
> #      - filter: Perform Action
> #        filter variables:
> #        - name: zenithTotalDelay
> #        action:
> #          name: passivate
> ```

@delippi Good discussion!

This triggered me to think about what will happen if we `accept` some obs and then `passivate` (all of) them;  or `passivate` first and then `accept` them? My guess is that the latter will replace (or disable, depending on whether to apply to all obs) the first one

delippi:
> > > Thanks for this info.
> > > For the ZTD yaml, it seems ok/working to have the "passive" section in the middle of the filters (double checked the QC flags).
> > 
> > 
> > You have the following order. You reject anything with QM > 10, then assign some error, then you passivate everything (including the previously rejected QM > 10 what you turn on that section). The passivate should be at the top (or very least before any rejection).
> > ```
> >        # Reject all obs with QualityMarker > 10
> >        - filter: Bounds Check
> >          filter variables:
> >          - name: zenithTotalDelay
> >          test variables:
> >          - name: QualityMarker/zenithTotalDelay
> >          maxvalue: 10
> > 
> >        # Initial error assignment
> >        - filter: Perform Action
> >          filter variables:
> >          - name: zenithTotalDelay
> >          action:
> >            name: assign error
> >            error parameter: 0.01
> > 
> >        # Monitoring mode (no assimilation, set QCflag=1)
> > #      - filter: Perform Action
> > #        filter variables:
> > #        - name: zenithTotalDelay
> > #        action:
> > #          name: passivate
> > ```
> 
> @delippi Good discussion!
> 
> This triggered me to think about what will happen if we `accept` some obs and then `passivate` (all of) them; or `passivate` first and then `accept` them? My guess is that the latter will replace (or disable, depending on whether to apply to all obs) the first one

This is the problem I'm having with mesonet winds since it uses accept lists rather than reject lists. the mesonet winds start by rejecting everything, then accepts only those from the use lists. If I put passivate before everything then the passivate will be erased. If I put passivate after the accept lists it erases everything before that and monitors everything. I need a way to access what has already been accepted/rejected. I need the working version of `EffectiveQC` during run time. Using ObsError as the example:

`ObsError`: initial values loaded from the input ioda file.
`ObsErrorData`: up-to-date values (set by one of the preceding filters).
`EffectiveError`: final values obtained after execution of all filters. This group does not exist while filters are running, but is present in the output ioda file.

guoqing-noaa:
> > > > Thanks for this info.
> > > > For the ZTD yaml, it seems ok/working to have the "passive" section in the middle of the filters (double checked the QC flags).
> > > 
> > > 
> > > You have the following order. You reject anything with QM > 10, then assign some error, then you passivate everything (including the previously rejected QM > 10 what you turn on that section). The passivate should be at the top (or very least before any rejection).
> > > ```
> > >        # Reject all obs with QualityMarker > 10
> > >        - filter: Bounds Check
> > >          filter variables:
> > >          - name: zenithTotalDelay
> > >          test variables:
> > >          - name: QualityMarker/zenithTotalDelay
> > >          maxvalue: 10
> > > 
> > >        # Initial error assignment
> > >        - filter: Perform Action
> > >          filter variables:
> > >          - name: zenithTotalDelay
> > >          action:
> > >            name: assign error
> > >            error parameter: 0.01
> > > 
> > >        # Monitoring mode (no assimilation, set QCflag=1)
> > > #      - filter: Perform Action
> > > #        filter variables:
> > > #        - name: zenithTotalDelay
> > > #        action:
> > > #          name: passivate
> > > ```
> > 
> > 
> > @delippi Good discussion!
> > This triggered me to think about what will happen if we `accept` some obs and then `passivate` (all of) them; or `passivate` first and then `accept` them? My guess is that the latter will replace (or disable, depending on whether to apply to all obs) the first one
> 
> This is the problem I'm having with mesonet winds since it uses accept lists rather than reject lists. the mesonet winds start by rejecting everything, then accepts only those from the use lists. If I put passivate before everything then the passivate will be erased. If I put passivate after the accept lists it erases everything before that and monitors everything. I need a way to access what has already been accepted/rejected. I need the working version of `EffectiveQC` during run time. Using ObsError as the example:
> 
> `ObsError`: initial values loaded from the input ioda file. `ObsErrorData`: up-to-date values (set by one of the preceding filters). `EffectiveError`: final values obtained after execution of all filters. This group does not exist while filters are running, but is present in the output ioda file.

We may use `reduce obs space` to remove all those obs not in the mesonet uselist and then apply the "passivate" filter.

delippi:
> > > > > Thanks for this info.
> > > > > For the ZTD yaml, it seems ok/working to have the "passive" section in the middle of the filters (double checked the QC flags).
> > > > 
> > > > 
> > > > You have the following order. You reject anything with QM > 10, then assign some error, then you passivate everything (including the previously rejected QM > 10 what you turn on that section). The passivate should be at the top (or very least before any rejection).
> > > > ```
> > > >        # Reject all obs with QualityMarker > 10
> > > >        - filter: Bounds Check
> > > >          filter variables:
> > > >          - name: zenithTotalDelay
> > > >          test variables:
> > > >          - name: QualityMarker/zenithTotalDelay
> > > >          maxvalue: 10
> > > > 
> > > >        # Initial error assignment
> > > >        - filter: Perform Action
> > > >          filter variables:
> > > >          - name: zenithTotalDelay
> > > >          action:
> > > >            name: assign error
> > > >            error parameter: 0.01
> > > > 
> > > >        # Monitoring mode (no assimilation, set QCflag=1)
> > > > #      - filter: Perform Action
> > > > #        filter variables:
> > > > #        - name: zenithTotalDelay
> > > > #        action:
> > > > #          name: passivate
> > > > ```
> > > 
> > > 
> > > @delippi Good discussion!
> > > This triggered me to think about what will happen if we `accept` some obs and then `passivate` (all of) them; or `passivate` first and then `accept` them? My guess is that the latter will replace (or disable, depending on whether to apply to all obs) the first one
> > 
> > 
> > This is the problem I'm having with mesonet winds since it uses accept lists rather than reject lists. the mesonet winds start by rejecting everything, then accepts only those from the use lists. If I put passivate before everything then the passivate will be erased. If I put passivate after the accept lists it erases everything before that and monitors everything. I need a way to access what has already been accepted/rejected. I need the working version of `EffectiveQC` during run time. Using ObsError as the example:
> > `ObsError`: initial values loaded from the input ioda file. `ObsErrorData`: up-to-date values (set by one of the preceding filters). `EffectiveError`: final values obtained after execution of all filters. This group does not exist while filters are running, but is present in the output ioda file.
> 
> We may use `reduce obs space` to remove all those obs not in the mesonet uselist and then apply the "passivate" filter.

I thought about that. It doesn't work. `reduce obs space` only removes obs from obs space that are selected by the current filter. All other obs previously rejected don't get removed. I think we need QCflagsData or something like that.

delippi:
> > > > > > Thanks for this info.
> > > > > > For the ZTD yaml, it seems ok/working to have the "passive" section in the middle of the filters (double checked the QC flags).
> > > > > 
> > > > > 
> > > > > You have the following order. You reject anything with QM > 10, then assign some error, then you passivate everything (including the previously rejected QM > 10 what you turn on that section). The passivate should be at the top (or very least before any rejection).
> > > > > ```
> > > > >        # Reject all obs with QualityMarker > 10
> > > > >        - filter: Bounds Check
> > > > >          filter variables:
> > > > >          - name: zenithTotalDelay
> > > > >          test variables:
> > > > >          - name: QualityMarker/zenithTotalDelay
> > > > >          maxvalue: 10
> > > > > 
> > > > >        # Initial error assignment
> > > > >        - filter: Perform Action
> > > > >          filter variables:
> > > > >          - name: zenithTotalDelay
> > > > >          action:
> > > > >            name: assign error
> > > > >            error parameter: 0.01
> > > > > 
> > > > >        # Monitoring mode (no assimilation, set QCflag=1)
> > > > > #      - filter: Perform Action
> > > > > #        filter variables:
> > > > > #        - name: zenithTotalDelay
> > > > > #        action:
> > > > > #          name: passivate
> > > > > ```
> > > > 
> > > > 
> > > > @delippi Good discussion!
> > > > This triggered me to think about what will happen if we `accept` some obs and then `passivate` (all of) them; or `passivate` first and then `accept` them? My guess is that the latter will replace (or disable, depending on whether to apply to all obs) the first one
> > > 
> > > 
> > > This is the problem I'm having with mesonet winds since it uses accept lists rather than reject lists. the mesonet winds start by rejecting everything, then accepts only those from the use lists. If I put passivate before everything then the passivate will be erased. If I put passivate after the accept lists it erases everything before that and monitors everything. I need a way to access what has already been accepted/rejected. I need the working version of `EffectiveQC` during run time. Using ObsError as the example:
> > > `ObsError`: initial values loaded from the input ioda file. `ObsErrorData`: up-to-date values (set by one of the preceding filters). `EffectiveError`: final values obtained after execution of all filters. This group does not exist while filters are running, but is present in the output ioda file.
> > 
> > 
> > We may use `reduce obs space` to remove all those obs not in the mesonet uselist and then apply the "passivate" filter.
> 
> I thought about that. It doesn't work. `reduce obs space` only removes obs from obs space that are selected by the current filter. All other obs previously rejected don't get removed. I think we need QCflagsData or something like that.

This works. Thanks @SamuelDegelia-NOAA. I put this after all the mesonet accept lists and it effectively does nothing. Which means it is only accepting obs that had already been accepted. 
```
         # Accept, reject, or passivate (monitor)
         - filter: Perform Action
           filter variables:
           - name: windEastward
           - name: windNorthward
           where:
           - variable: QCflagsData/windEastward
             is_in: 0
           action:
             name: accept # accept, reject, passivate
```

guoqing-noaa:
@delippi I am thinking about something like this:
```
- filter: RejectList
where:
  - variable: MetaData/dataProviderOrigin
    is_not_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
            "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
            "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
            "NOS-NWLON","GoMOOS","HADS"]
action:
   name: reduce obs space
```
It may work.

delippi:
> @delippi I am thinking about something like this:
> 
> ```
> - filter: RejectList
> where:
>   - variable: MetaData/dataProviderOrigin
>     is_not_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
>             "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
>             "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
>             "NOS-NWLON","GoMOOS","HADS"]
> action:
>    name: reduce obs space
> ```
> 
> It may work.

I have a working solution above using `QCflagsData`.

This wouldn't work because it actually needs to be AcceptList (not RejectList). These are data providers that we want to keep. Like this
```
         # Mesonet winds provider/station use lists
         - filter: RejectList  # initially reject all providers
         - filter: AcceptList  # accept back selected providers
           where:
           - variable: MetaData/dataProviderOrigin
             is_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
                     "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
                     "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
                     "NOS-NWLON","GoMOOS","HADS"]
```

guoqing-noaa:
> > @delippi I am thinking about something like this:
> > ```
> > - filter: RejectList
> > where:
> >   - variable: MetaData/dataProviderOrigin
> >     is_not_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
> >             "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
> >             "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
> >             "NOS-NWLON","GoMOOS","HADS"]
> > action:
> >    name: reduce obs space
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It may work.
> 
> I have a working solution above using `QCflagsData`.
> 
> This wouldn't work because it actually needs to be AcceptList (not RejectList). These are data providers that we want to keep. Like this
> 
> ```
>          # Mesonet winds provider/station use lists
>          - filter: RejectList  # initially reject all providers
>          - filter: AcceptList  # accept back selected providers
>            where:
>            - variable: MetaData/dataProviderOrigin
>              is_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
>                      "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
>                      "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
>                      "NOS-NWLON","GoMOOS","HADS"]
> ```

I used `is_not_in`, i.e. reject all obs NOT in the use list and remove them from the obs space.  Do you want to remove all obs outside of the uselist?

delippi:
> > > @delippi I am thinking about something like this:
> > > ```
> > > - filter: RejectList
> > > where:
> > >   - variable: MetaData/dataProviderOrigin
> > >     is_not_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
> > >             "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
> > >             "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
> > >             "NOS-NWLON","GoMOOS","HADS"]
> > > action:
> > >    name: reduce obs space
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > It may work.
> > 
> > 
> > I have a working solution above using `QCflagsData`.
> > This wouldn't work because it actually needs to be AcceptList (not RejectList). These are data providers that we want to keep. Like this
> > ```
> >          # Mesonet winds provider/station use lists
> >          - filter: RejectList  # initially reject all providers
> >          - filter: AcceptList  # accept back selected providers
> >            where:
> >            - variable: MetaData/dataProviderOrigin
> >              is_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
> >                      "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
> >                      "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
> >                      "NOS-NWLON","GoMOOS","HADS"]
> > ```
> 
> I used `is_not_in`, i.e. reject all obs NOT in the use list and remove them from the obs space. Do you want to remove all obs outside of the uselist?

Oh! I missed that detail (`is_in` vs `is_not_in`). I still don't think it would work (with how it is currently configured; with more than 1 accept list) because the result of your suggestion would reject and remove from memory anything not in the list. Then you couldn't try accepting anything else because nothing else exists anymore. It might work if we had a single accept list--I don't think there is a particular reason it couldn't be a single accept list other than it would be too much effort to try to redo right now haha! Thank you for this suggestion though. We can keep it in mind in the future, but I think the other solution works just fine too.

guoqing-noaa:
> > > > @delippi I am thinking about something like this:
> > > > ```
> > > > - filter: RejectList
> > > > where:
> > > >   - variable: MetaData/dataProviderOrigin
> > > >     is_not_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
> > > >             "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
> > > >             "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
> > > >             "NOS-NWLON","GoMOOS","HADS"]
> > > > action:
> > > >    name: reduce obs space
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > It may work.
> > > 
> > > 
> > > I have a working solution above using `QCflagsData`.
> > > This wouldn't work because it actually needs to be AcceptList (not RejectList). These are data providers that we want to keep. Like this
> > > ```
> > >          # Mesonet winds provider/station use lists
> > >          - filter: RejectList  # initially reject all providers
> > >          - filter: AcceptList  # accept back selected providers
> > >            where:
> > >            - variable: MetaData/dataProviderOrigin
> > >              is_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
> > >                      "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
> > >                      "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
> > >                      "NOS-NWLON","GoMOOS","HADS"]
> > > ```
> > 
> > 
> > I used `is_not_in`, i.e. reject all obs NOT in the use list and remove them from the obs space. Do you want to remove all obs outside of the uselist?
> 
> Oh! I missed that detail (`is_in` vs `is_not_in`). I still don't think it would work (with how it is currently configured; with more than 1 accept list) because the result of your suggestion would reject and remove from memory anything not in the list. Then you couldn't try accepting anything else because nothing else exists anymore. It might work if we had a single accept list--I don't think there is a particular reason it couldn't be a single accept list other than it would be too much effort to try to redo right now haha! Thank you for this suggestion though. We can keep it in mind in the future, but I think the other solution works just fine too.


I think we may still want to redo it for two reasons:
1. There are a lot of mesonet stations; we want to remove non-interested mesonet stations from the obs space to reduce the memory footprint. 
2. We update the sfcobs uselist every day (as is done in GSI right now). So we only need to keep a simple template in the template YAML file and replace it with the actual uselist, as is done in GSI.  Two ways to achieve this: (1). Modify the JEDI source codes to read a given file;  (2). update the super yaml files on the fly.

delippi:
> > > > > @delippi I am thinking about something like this:
> > > > > ```
> > > > > - filter: RejectList
> > > > > where:
> > > > >   - variable: MetaData/dataProviderOrigin
> > > > >     is_not_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
> > > > >             "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
> > > > >             "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
> > > > >             "NOS-NWLON","GoMOOS","HADS"]
> > > > > action:
> > > > >    name: reduce obs space
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > It may work.
> > > > 
> > > > 
> > > > I have a working solution above using `QCflagsData`.
> > > > This wouldn't work because it actually needs to be AcceptList (not RejectList). These are data providers that we want to keep. Like this
> > > > ```
> > > >          # Mesonet winds provider/station use lists
> > > >          - filter: RejectList  # initially reject all providers
> > > >          - filter: AcceptList  # accept back selected providers
> > > >            where:
> > > >            - variable: MetaData/dataProviderOrigin
> > > >              is_in: ["RAWS","NEDOR","SFWMD","MOComAgNet","UDFCD","CAIC",
> > > >                      "CODOT","KSDOT","IADOT","MEDOT","MNDOT","NDDOT",
> > > >                      "OK-Meso","MesoWest","WT-Meso","WYDOT","NOS-PORTS",
> > > >                      "NOS-NWLON","GoMOOS","HADS"]
> > > > ```
> > > 
> > > 
> > > I used `is_not_in`, i.e. reject all obs NOT in the use list and remove them from the obs space. Do you want to remove all obs outside of the uselist?
> > 
> > 
> > Oh! I missed that detail (`is_in` vs `is_not_in`). I still don't think it would work (with how it is currently configured; with more than 1 accept list) because the result of your suggestion would reject and remove from memory anything not in the list. Then you couldn't try accepting anything else because nothing else exists anymore. It might work if we had a single accept list--I don't think there is a particular reason it couldn't be a single accept list other than it would be too much effort to try to redo right now haha! Thank you for this suggestion though. We can keep it in mind in the future, but I think the other solution works just fine too.
> 
> I think we may still want to redo it for two reasons:
> 
> 1. There are a lot of mesonet stations; we want to remove non-interested mesonet stations from the obs space to reduce the memory footprint.
> 2. We update the sfcobs uselist every day (as is done in GSI right now). So we only need to keep a simple template in the template YAML file and replace it with the actual uselist, as is done in GSI.  Two ways to achieve this: (1). Modify the JEDI source codes to read a given file;  (2). update the super yaml files on the fly.

Those are good points! Though I wonder if those sort of efforts should be motivation for focusing on JCB?

HuiLiu-NOAA:
JCSDA just commented that there are new namings for all surface observations, like  airTemperatureAt2M, rather than  airTemperature currently. It may be better to change these names accordingly, especially for the new surface OBS operator (C++ version) we are developing with JCSDA.

-------------
 Hi Sijie,

The problem with the suggested approach is in the naming convention airTemperature is representative of a profile of temperatures where-as airTemperatureAt2M is the agreed name for surface temperature.  Therefore, your suggestion goes against the naming conventions.  I would suggest a change in the IODA converter to match the naming standard would be a better approach.

I would suggest discussing this further with Ben and Raji (the new UFO admin).

All the best
Mike

delippi:
> JCSDA just commented that there are new namings for all surface observations, like airTemperatureAt2M, rather than airTemperature currently. It may be better to change these names accordingly, especially for the new surface OBS operator (C++ version) we are developing with JCSDA.
> 
> Hi Sijie,
> 
> The problem with the suggested approach is in the naming convention airTemperature is representative of a profile of temperatures where-as airTemperatureAt2M is the agreed name for surface temperature. Therefore, your suggestion goes against the naming conventions. I would suggest a change in the IODA converter to match the naming standard would be a better approach.
> 
> I would suggest discussing this further with Ben and Raji (the new UFO admin).
> 
> All the best Mike

Thanks @HuiLiu-NOAA! Can you point me to where that comment was made? I'd like to know what the context is.

delippi:
Putting that context here:
> I wanted to discuss an issue we've met with the current unified operator. Currently, when we convert observations from prepbufr to ioda netcdf format, we don't separate temperature readings from surface instruments (e.g., adpsfc) and upper air instruments (e.g., adpupa).
>
> However, I've discovered that using function names like airTemperatureAt2M_WRFDA/UKMO/GSL requires us to separate these observation types during the ioda conversion process.
>
> I'm wondering if it might be better to use function names such as airTemperature_WRFDA/UKMO/GSL in our unified operator instead. This approach would allow us to maintain our current workflow without having to separate the observation types.

So does this mean that we will likely need to change our bufr2ioda tools to output variables like `airTemperatureAt2M` for surface data in order to use the surface operator that sijie is working on?

I don't think a name change is urgent at this point; let's cross that bridge when we get there.

guoqing-noaa:
> JCSDA just commented that there are new namings for all surface observations, like airTemperatureAt2M, rather than airTemperature currently. It may be better to change these names accordingly, especially for the new surface OBS operator (C++ version) we are developing with JCSDA.
> 
> Hi Sijie,
> 
> The problem with the suggested approach is in the naming convention airTemperature is representative of a profile of temperatures where-as airTemperatureAt2M is the agreed name for surface temperature. Therefore, your suggestion goes against the naming conventions. I would suggest a change in the IODA converter to match the naming standard would be a better approach.
> 
> I would suggest discussing this further with Ben and Raji (the new UFO admin).
> 
> All the best Mike

@HuiLiu-NOAA Thanks for the information. I would suggest we keep this PR as is and add/change things in separate new PRs.

delippi:
FYI, I only updated ctest references for ens3dvar ctests since there seems to be some issues with q assimilation in getkf. @SamuelDegelia-NOAA is aware and opened an issue #366 to investigate.

rrfs-test/validated_yamls/templates/obtype_config/adpsfc_specificHumidity_181.yaml
```diff
@@ -125,7 +122,7 @@
...
```
guoqing-noaa: I suggest we remove all `request_saturation_specific_humidity_geovals` lines since we no longer need this after Hui's JEDI PR. Thanks!

delippi: I had left it in to show that it is intentionally set to false. I can remove it if you prefer.

HuiLiu-NOAA: I agree that it is clear to remove this

guoqing-noaa (approved):

SamuelDegelia-NOAA (approved):
Thanks @delippi for the great efforts to improve these yamls.

ShunLiu-NOAA (approved):

Issue #309, opened 2025-03-04
[bug] MPAS crashing after restart from GETKF analyses


SamuelDegelia-NOAA:
## Current behavior (describe the bug)

The rrfs-workflow [PR #665](https://github.com/NOAA-EMC/rrfs-workflow/pull/665) added functionality to cycle ensemble DA using GETKF. The GETKF tasks complete successfully but I am seeing crashes for the fcst step on the 0300 UTC cycle. The error is: 

```
----------------------------------------------------------------------
Beginning MPAS-atmosphere Error Log File for task       1 of      40
    Opened at 2025/03/03 22:26:38
----------------------------------------------------------------------

ERROR: Error in compute_layer_mean: pressure should increase with index
```

## Steps to Reproduce
Running on hera, run the workflow using the default 27 May retro case with `exp.ens_conus12km` and the following settings: 
```
export DO_IODA="true"
export DO_JEDI=true
```

## Expected behavior
The forecast step should not crash after GETKF analysis.

SamuelDegelia-NOAA:
More error information from `/scratch2/BMC/zrtrr/Samuel.Degelia/rrfs-workflow-getkf/experiment/stmp/20240527/rrfs_fcst_03_v2.0.9/enkf/mem001/fcst_03/log.atmosphere.0000.out`: 

```
 --- subroutine MPAS_to_phys - pressure(1) < pressure(2):
 i      =16
 latCell=40.0822
 lonCell=287.934
 1 16 1 17.9503 99877.4 -99876.5 0.945312 0.310608E-03 287.190 10.5356 0.107672E-01
 1 16 2 20.2955 99616.7 -99609.7 7.00781 0.128606E-02 289.898 18.8474 0.119644E-01
 1 16 3 22.9392 99322.8 -99315.1 7.69531 0.137200E-02 290.819 19.4275 0.119640E-01
 1 16 4 25.9173 98991.8 -98984.5 7.24219 0.131206E-02 291.234 19.1206 0.118730E-01
 1 16 5 29.2696 98619.2 -98612.6 6.57031 0.122087E-02 291.550 18.6055 0.117649E-01
 1 16 6 33.0401 98200.1 -98194.1 6.08594 0.115428E-02 291.943 18.2269 0.116568E-01
 1 16 7 37.2760 97729.4 -97724.0 5.47656 0.106815E-02 292.451 17.7125 0.115488E-01
[...]

PR #331, opened 2025-03-18
Update stream name for GETKF in mpas-jedi to prevent forecast crash and improve hofx calculations

SamuelDegelia-NOAA:
## Description
This PR addresses issue #309 which found that the ensemble stream file was missing `rho` (thanks to @spanNOAA for helping with this issue). Additional analysis [here](https://github.com/NOAA-EMC/RDASApp/issues/309#issuecomment-2734216761) and [here](https://docs.google.com/presentation/d/1laEtotJBjZ12qsG4waCdNE905pehN9sEIJiYE2gwevI/edit?slide=id.g338ac2b1ae8_0_12#slide=id.g338ac2b1ae8_0_12) found that additional state variables were missing from the ensemble stream file needed for the surface winds and radiance observation operators.  We now use the background stream for the GETKF ensemble members which includes `rho` and these additional state variables. 

Also included in this PR is an update to `update_ctest_inputs.sh` to also run `gen_yaml_ctest.sh` so that you can update the ctest super yamls with one script if making any edits to the individual yamls. A new utility is also added called `update_ctest_refs.sh` that will update the test references in `RDASApp/rrfs-test/testoutput` for previously run ctests. The wiki page [here](https://github.com/NOAA-EMC/RDASApp/wiki/Updating-the-RDASApp-ctests-(adding-new-obs-or-updating-test-reference-files)) has been updated to provide streamlined instructions for using these tools and updating the ctests. 

Finally, the test references are also updated in this PR. Since #314 was merged after the update to the ctest structure, all references needed to be updated (not just the MPAS GETKF tests). 

## Issue(s) addressed
#309 

None

## Checklist


fix/stream_list/stream_list.atmosphere.ensemble
```diff
@@ -1,7 +1,23 @@
...
+xice
+snowc
+skintemp
+snowh
+vegfra
+u10
+v10
+lai
+smois
+tslb
```
guoqing-noaa: Do we want to include all state variables in `getkf_solver.yaml`?
```
_member: &memberConfig
  date: &analysisDate '@analysisDate@'   
  state variables: [spechum,surface_pressure,temperature,uReconstructMeridional,uReconstructZonal,
theta,rho,u,qv,pressure,landmask,xice,snowc,skintemp,ivgtyp,isltyp,snowh,vegfra,
u10,v10,lai,smois,tslb,pressure_p,qc,qi,qg,qr,qs,cldfrac]
```

guoqing-noaa: check `jedivar.yaml`, we have much less state variables:
```
_member: &memberConfig
  date: &analysisDate '@analysisDate@'
  state variables: &incvars [spechum,surface_pressure,temperature,uReconstructMeridional,uReconstructZonal]
  stream name: ensemble

```

guoqing-noaa: After some examination, I don't think we need to modify `stream_list.atmpshere.ensemble`. 

I think we need to use the background stream for GETKF.

i.e. For ensembles, we use the 'ensemble' stream when doing EnVAR while using the "background" stream when doing GETKF (similar to JEDIVAR where we need to read the background fields).

SamuelDegelia-NOAA: Is there any functional difference to setting the ensemble stream for GETKF to refer to `stream_list.atmosphere.background` instead of updating the list of variables in `stream_list.atmosphere.ensemble`? It seems to me like they are both accomplishing the same thing. To me it seems more straightforward to have the ensemble stream refer to the ensemble stream list file.

guoqing-noaa: I think we need to use the `background` stream so JEDIVAR and GETKF are on the same page to read DA-related state variables. 

We will also want to double-check the background stream as it only lists 24 variables while the yaml file lists 30 variables.

guoqing-noaa: Just checked NCAR runs, they used the `background` stream for GETKF. I think that explains why they don't have crash issues with that 7-variable ensemble stream.

I would think the `ensemble` stream is designed only for the EnVAR process to tell JEDIVAR what variables will take part in providing ensemble BEC.

guoqing-noaa: We may still want to add `rho`, `spechum`, `temperature` into this ensemble stream?

Ming showed this figure:
image

Apparently, the ensemble correlation between T and RH/Wind does not provide a positive impact.

SamuelDegelia-NOAA: I think both solutions, whether we add more variables to the ensemble stream list or change the GETKF yaml to `stream name: background`, are basically doing the same thing. Adding these additional variables to the ensemble stream did not modify the EnVar results (ino change in ensemble BEC). But if we are looking to match the NCAR configuration then I am fine with modifying the yaml instead of the stream list. Could you share their yamls so I can keep them as a reference?

guoqing-noaa: Here is an example from the tutorial LETKF case:
https://www2.mmm.ucar.edu/projects/mpas-jedi/tutorial/202309NCAR/lectures/8-EDA-LETKF.pdf
image

guoqing-noaa: I think it makes sense to use the `background` stream as this configuration is under the `background` section:
```
background:
  members from template:
    template:
      <<: *memberConfig
      filename: ./data/ens/mem%iMember%.nc
    pattern: "%iMember%"
    start: 1
    zero padding: 3
    nmembers: 30
```
```
_member: &memberConfig
  date: &analysisDate '@analysisDate@'
  state variables: [spechum,surface_pressure,temperature,uReconstructMeridional,uReconstructZonal,theta,rho,u,qv,pressure,
landmask,xice,snowc,skintemp,ivgtyp,isltyp,snowh,vegfra,u10,v10,lai,smois,tslb,pressure_p,qc,qi,qg,qr,qs,cldfrac]
  stream name: background
```
Are you okay with this setting?

I agree with you for GETKF,  it does not matter whether this stream is named as `background` or `ensemble` as long as they are the same. 
But on the other hand, if they are the same, we don't need an extra "ensemble" stream, right? We may just use the `background` stream in all these situations.
Also, adding more variables in the ensemble stream may increase the I/O time for JEDIVAR as it will read more variables than expected.

SamuelDegelia-NOAA: Yes that solution is fine with me - I'll update the PR tomorrow morning.

guoqing-noaa: Sounds good. Thanks!

SamuelDegelia-NOAA:
Updated the PR per discussion with @guoqing-noaa above. Instead of adding more variables to the ensemble stream, we now use the background stream for the ensemble members in the GETKF yaml.

SamuelDegelia-NOAA:
@guoqing-noaa Would you mind re-reviewing this PR (whenever you have the time) since this fix works for rrfs-workflow? I included some updated ctest tools in this PR that would be good for others to use.

guoqing-noaa (approved):

SamuelDegelia-NOAA:
@ShunLiu-NOAA It would be great if we could merge this one soon too so that others could have the updated ctest tools.

ShunLiu-NOAA (approved):

PR #336, opened 2025-03-20
Hotfix for offline domain check outputting masked longitude info

SamuelDegelia-NOAA:
## Bug Fix Description
This PR fixes an issue with the offline domain check outputing masked longitude data for ZTD obs. 

I also added a few global attributes to the output file so that we can track the hull shrink factor and which domain (mpas or jedi) was used for the check. 

**Description of Current Behavior:**
- Running `offline_domain_check.py` on ZTD obs outputs all masked longitude data

**Description of Fixed Behavior:**
- We now output the unmasked latitude and longitude data for the offline domain check. We still leave other variables masked if the data are bad. 

## Issue(s) addressed
None

## Dependencies (if applicable)
None

## Checklist
- [x] I have performed a self-review of my own code.

PR #342, opened 2025-03-26
Hotfix for `offline_add_var_to_ioda.py` outputting masked longitude info

SamuelDegelia-NOAA:
## Bug Fix Description
This is a similar hotfix to #336. The `offline_add_var_to_ioda.py` utility also was having issues with masked longitude info for the ZTD obs. This PR applies a similar fix to get the longitude data working correctly. 

Running `offline_add_var_to_ioda.py` on ZTD obs outputs all masked longitude data.

We now output the unmasked latitude and longitude data for the offline tool. We still leave other variables masked if the data are bad.

## Issue(s) addressed
None

None

## Checklist


delippi (approved):
LGTM

HuiLiu-NOAA:
Thanks @SamuelDegelia-NOAA!! This fixes the issue.

@guoqing-noaa : we need to update the offline_ioda_tweak.py and offline_domain_check.py  in the workflow as well.

guoqing-noaa:
> Thanks @SamuelDegelia-NOAA!! This fixes the issue.
> 
> @guoqing-noaa : we need to update the offline_ioda_tweak.py and offline_domain_check.py in the workflow as well.

@HuiLiu-NOAA Yes, can you or @SamuelDegelia-NOAA  make a PR to rrfs-workflow to get it updated?
Thanks!

SamuelDegelia-NOAA:
@guoqing-noaa I can make a PR to rrfs-workflow.

HuiLiu-NOAA:
Thanks @SamuelDegelia-NOAA !! Maybe also the related ztd python and jedi yaml as well.

SamuelDegelia-NOAA:
> Thanks @SamuelDegelia-NOAA !! Maybe also the related ztd python and jedi yaml as well.

Are there ztd python codes or yamls that need to be updated in rrfs-workflow? I had assumed that those were grabbed from RDASApp.

HuiLiu-NOAA:
> > Thanks @SamuelDegelia-NOAA !! Maybe also the related ztd python and jedi yaml as well.
> 
> Are there ztd python codes or yamls that need to be updated in rrfs-workflow? I had assumed that those were grabbed from RDASApp.

Per @guoqing-noaa, the latest RDASApp develop/ is NOT transferred/synced to the workflow automatically.  Maybe need a way to do this.

guoqing-noaa:
> > > Thanks @SamuelDegelia-NOAA !! Maybe also the related ztd python and jedi yaml as well.
> > 
> > 
> > Are there ztd python codes or yamls that need to be updated in rrfs-workflow? I had assumed that those were grabbed from RDASApp.
> 
> Per @guoqing-noaa, the latest RDASApp develop/ is NOT transferred/synced to the workflow automatically. Maybe need a way to do this.

At this point, I still recommend updating RDASApp hash by creating PRs to workflow.
A file works in RDASApp may not work by default in rrfs-workflow, especially in the cycling mode and under different DA settings. To make the workflow stable, we need to do workflow cycling tests before a merge. So it is expected that RDASApp in rrfs-workflow may lag a little bit behind the authoritative RDASApp, similar to the situation that JEDI in RDASApp lags behind the authoritative JEDI.

ShunLiu-NOAA (approved):

ShunLiu-NOAA:
@guoqing-noaa so we can merge this first in RDASApp. Let me know if you have any concern about this PR.

guoqing-noaa (approved):

rrfs-test/IODA/offline_add_var_to_ioda.py
```diff
@@ -75,7 +75,10 @@ def toc(tic=tic, label=""):
             g.createVariable(var, vartype, 'Location', fill_value=fill)
         except:  # String variables
             g.createVariable(var, 'str', 'Location')
-        g.variables[var][:] = invar[:][:]
+        if var in ['latitude', 'longitude']:
+            g.variables[var][:] = invar[:][:].data
+        else:
+            g.variables[var][:] = invar[:][:]
```
guoqing-noaa, line 81: I have approved this PR. But I have some questions:

why lat/lon need to access the `data` object while other variables don't?
why lat/lon works before this change?

SamuelDegelia-NOAA: Something is causing the longitude data to become masked for the ZTD obs when they are converted from (-180,180) to (0,360). I am not sure why it is only occurring for those data. But using `.data` on those variables removes the masking. I do not do this for other variable types since masked values for ObsValues or other variables might be indicative of actual bad data.

guoqing-noaa: @SamuelDegelia-NOAA Ah, thank you for the explanation.  I don't know how you convert from (-180,180) to (0,360).

Here is a possible reason:
Longitude is a coordinate variable, and we may need to rearrange the order of Longitude and the associated data.
For example, `[-180, -90, 0, 90, 180]` becomes `[180, 270, 0, 90, 180]`, which may not be able to act as a coordinate variable now.

Here is a quick example to rearrange the order:
```
lon = np.array([-180, -90, 0, 90, 180])
lon_new = (lon + 360) % 360  # Convert
sort_pos = np.argsort(lon_new)  # Get sorting order
lon_sorted = lon_new[sort_pos]
data_sorted = data[sort_pos]  # Sort data accordingly
```

ShunLiu-NOAA: @guoqing-noaa We can continue to test the potential changes later. But we will merge this PR because the PR resolves the current issue.

HuiLiu-NOAA:
Better to continue to sort this out as @guoqing-noaa commented. I also had a similar puzzle about this.

SamuelDegelia-NOAA:
The masked longitude seems to be coming from the bufr2ioda conversion. [Here](https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L120) is the line where the longitude is transformed to (0,360). The `offline_add_var_to_ioda.py` code does not cause the masked values, but using `.data` on the read in variables does remove the masked values: 

Read in the original IODA file: 
```
(eva) [Samuel.Degelia@hfe05 IODA]$ python
>>> from netCDF4 import Dataset
>>> nc = Dataset('ioda_ztd.nc', 'r')
```

Print the `longitude` variable from the original file: 
```
>>> nc.groups['MetaData'].variables['longitude'][:]
masked_array(data=[4.594950199127197, --, --, ..., --, --,
                   0.4544999897480011],
             mask=[False,  True,  True, ...,  True,  True, False],
       fill_value=3.4028235e+38,
            dtype=float32)
```

Print the `longitude` variable with `.data` attribute to remove masked values: 
```
>>> nc.groups['MetaData'].variables['longitude'][:].data
array([  4.59495, 353.06973, 334.33722, ..., 357.797  , 359.4325 ,
         0.4545 ], dtype=float32)
```

So I think this solution in `offline_add_var_to_ioda.py` is good for now, but if we want to solve the root problem then we will need to investigate why the bufr2ioda code is outputting a masked_array for `longitude`.

guoqing-noaa:
> The masked longitude seems to be coming from the bufr2ioda conversion. [Here](https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L120) is the line where the longitude is transformed to (0,360). The `offline_add_var_to_ioda.py` code does not cause the masked values, but using `.data` on the read in variables does remove the masked values:
> 
> Read in the original IODA file:
> 
> ```
> (eva) [Samuel.Degelia@hfe05 IODA]$ python
> >>> from netCDF4 import Dataset
> >>> nc = Dataset('ioda_ztd.nc', 'r')
> ```
> 
> Print the `longitude` variable from the original file:
> 
> ```
> >>> nc.groups['MetaData'].variables['longitude'][:]
> masked_array(data=[4.594950199127197, --, --, ..., --, --,
>                    0.4544999897480011],
>              mask=[False,  True,  True, ...,  True,  True, False],
>        fill_value=3.4028235e+38,
>             dtype=float32)
> ```
> 
> Print the `longitude` variable with `.data` attribute to remove masked values:
> 
> ```
> >>> nc.groups['MetaData'].variables['longitude'][:].data
> array([  4.59495, 353.06973, 334.33722, ..., 357.797  , 359.4325 ,
>          0.4545 ], dtype=float32)
> ```
> 
> So I think this solution in `offline_add_var_to_ioda.py` is good for now, but if we want to solve the root problem then we will need to investigate why the bufr2ioda code is outputting a masked_array for `longitude`.

@SamuelDegelia-NOAA  Thanks a lot for the detailed information. 
It looks like it is caused by modifying the coordinate variable `longitude` .

@HuiLiu-NOAA Are the conversions from (180,180) to (0, 360) done inside the C++ code or the Python code?  Why only ZTD has this issue?

HuiLiu-NOAA:
> > The masked longitude seems to be coming from the bufr2ioda conversion. [Here](https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L120) is the line where the longitude is transformed to (0,360). The `offline_add_var_to_ioda.py` code does not cause the masked values, but using `.data` on the read in variables does remove the masked values:
> > Read in the original IODA file:
> > ```
> > (eva) [Samuel.Degelia@hfe05 IODA]$ python
> > >>> from netCDF4 import Dataset
> > >>> nc = Dataset('ioda_ztd.nc', 'r')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Print the `longitude` variable from the original file:
> > ```
> > >>> nc.groups['MetaData'].variables['longitude'][:]
> > masked_array(data=[4.594950199127197, --, --, ..., --, --,
> >                    0.4544999897480011],
> >              mask=[False,  True,  True, ...,  True,  True, False],
> >        fill_value=3.4028235e+38,
> >             dtype=float32)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Print the `longitude` variable with `.data` attribute to remove masked values:
> > ```
> > >>> nc.groups['MetaData'].variables['longitude'][:].data
> > array([  4.59495, 353.06973, 334.33722, ..., 357.797  , 359.4325 ,
> >          0.4545 ], dtype=float32)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I think this solution in `offline_add_var_to_ioda.py` is good for now, but if we want to solve the root problem then we will need to investigate why the bufr2ioda code is outputting a masked_array for `longitude`.
> 
> @SamuelDegelia-NOAA Thanks a lot for the detailed information. It looks like it is caused by modifying the coordinate variable `longitude` .
> 
> @HuiLiu-NOAA Are the conversions from (180,180) to (0, 360) done inside the C++ code or the Python code? Why only ZTD has this issue?

The conversion is done in bufr2ioda code.  The other data seems are all in (0, 360) range.

guoqing-noaa:
> The masked longitude seems to be coming from the bufr2ioda conversion. [Here](https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L120) is the line where the longitude is transformed to (0,360). 

Sorry that I missed this part.  Thank @SamuelDegelia-NOAA for the link to the actual code where the conversions are made.

@HuiLiu-NOAA Is it possible that line 281 causes the masked array?
https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L281-L283

`lon` is now in the range of 0~360, but we still set the `valid_range` attribute as -180 ~ 180?

HuiLiu-NOAA:
> > The masked longitude seems to be coming from the bufr2ioda conversion. [Here](https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L120) is the line where the longitude is transformed to (0,360).
> 
> Sorry that I missed this part. Thank @SamuelDegelia-NOAA for the link to the actual code where the conversions are made.
> 
> @HuiLiu-NOAA Is it possible that line 281 causes the masked array?
> 
> https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L281-L283
> 
> `lon` is now in the range of 0~360, but we still set the `valid_range` attribute as -180 ~ 180?

Yes, you are right!  I forgot to change the range.  Thanks!!

guoqing-noaa:
> > > The masked longitude seems to be coming from the bufr2ioda conversion. [Here](https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L120) is the line where the longitude is transformed to (0,360).
> > 
> > 
> > Sorry that I missed this part. Thank @SamuelDegelia-NOAA for the link to the actual code where the conversions are made.
> > @HuiLiu-NOAA Is it possible that line 281 causes the masked array?
> > https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L281-L283
> > 
> > `lon` is now in the range of 0~360, but we still set the `valid_range` attribute as -180 ~ 180?
> 
> Yes, you are right! I forgot to change the range. Thanks!!

@HuiLiu-NOAA Could you test whether a change to line 218 solves the issue?  If yes, I would suggest we revert the change here. Thanks!

HuiLiu-NOAA:
> > > > The masked longitude seems to be coming from the bufr2ioda conversion. [Here](https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L120) is the line where the longitude is transformed to (0,360).
> > > 
> > > 
> > > Sorry that I missed this part. Thank @SamuelDegelia-NOAA for the link to the actual code where the conversions are made.
> > > @HuiLiu-NOAA Is it possible that line 281 causes the masked array?
> > > https://github.com/NOAA-EMC/RDASApp/blob/e21d322490fb7e114cb340e05a5a711fed8544b3/rrfs-test/IODA/python/bufr2ioda_ztd.py#L281-L283
> > > 
> > > `lon` is now in the range of 0~360, but we still set the `valid_range` attribute as -180 ~ 180?
> > 
> > 
> > Yes, you are right! I forgot to change the range. Thanks!!
> 
> @HuiLiu-NOAA Could you test whether a change to line 218 solves the issue? If yes, I would suggest we revert the change here. Thanks!

Yes, fix the range in the bufr2ioda python code solve the problem!!

It is interesting that this range issue did not caused masked longitude in the original ioda_ztd.nc, but did cause the masked longitude in the file after the "offline_domain" check.  I will submit a fix PR for this.

PR #344, opened 2025-03-27
Updating mpas hash from  041ea960 to fba865e9

chunhuazhou:
Updating mpas hash from  041ea960 to fba865e9

PR #329, opened 2025-03-17
Add abi_g16 & abi_g18 YAML files into templates

xyzemc:
This PR is merging the GOES_16 and GOES_18 clear sky ABI YAML file into RDASApp templates. All validation steps have been processed in the following linked issues. Both bias correction and QC filers work properly in JEDI for assimilating the clear sky ABI clear sky (CSR) radiance.
## Issue(s) addressed
Resolves/Results are documented in Issue #
#57 
#292 
#306 
#249 

## Checklist

- [x]  I have performed a self-review of my own code.

- [x] I have run rrfs tests before creating the PR.

`Ctest with  abi_g16`
> 1/1 Test #4: rrfs_mpasjedi_2024052700_Ens3Dvar ...   Passed  440.84 sec
> 
> The following tests passed:
>         rrfs_mpasjedi_2024052700_Ens3Dvar
> 
> 100% tests passed, 0 tests failed out of 1
> 
> Label Time Summary:
> mpi            = 440.84 sec*proc (1 test)
> rdas-bundle    = 440.84 sec*proc (1 test)
> script         = 440.84 sec*proc (1 test)
> 
> Total Test time (real) = 440.89 sec

`Ctest with abi_g18 on top of abi_g16`

> 1/1 Test #4: rrfs_mpasjedi_2024052700_Ens3Dvar ...   Passed  453.52 sec
> 
> The following tests passed:
>         rrfs_mpasjedi_2024052700_Ens3Dvar
> 
> 100% tests passed, 0 tests failed out of 1
> 
> Label Time Summary:
> mpi            = 453.52 sec*proc (1 test)
> rdas-bundle    = 453.52 sec*proc (1 test)
> script         = 453.52 sec*proc (1 test)
> 
> Total Test time (real) = 453.58 sec
> [Xiaoyan.Zhang@hfe02 rrfs-test]$
> 
> 
>

Issue #327, opened 2025-03-14
[bug] Reproducibility issues with updated obs-space yamls used for ctests


SamuelDegelia-NOAA:
## Current behavior (describe the bug)

As discussed in #298, #313, and #321, updating the ctests to use the new obs-space yamls in `rrfs-test/validated_yamls/templates/obtype_configs` is causing reproducibility issues across various machines. The current list of ob types included in the ctests have been manually edited so that the ctests pass on all machines. This issue will consolidate discussion of this topic that was previously split across individual issues and PRs. 

Here is the known list of yamls that, when added to the ctests, causes a failure either when run on the same machine or between Hera/Jet/Hercules. It is possible that more yamls are having similar issues but that the differences are below our assigned ctest tolerance. 

```
msonet_stationPressure_188.yaml # Different result on Hera/Hercules 
adpsfc_stationPressure_187.yaml # Different results on Hera/Jet
adpupa_winds_220 # Repeatability issue on Hera
atms_n20.yaml # Different results on Hera/Jet
```

An example of the repeatability issue for `adpupa_winds_220` wherein the `rrfs_fv3jedi_2024052700_Ens3Dvar` ctest is run five times in a row on Hera: 

```
 Run 1: 
  0: Test Line: 'CostJo   : Nonlinear Jo(adpupa_winds_220) = 3.2593295738966270e+04, nobs = 48714, Jo/n = 6.6907451120758443e-01, err = 1.6247685156989053e+18'
Run 2: 
  0: Test Line: 'CostJo   : Nonlinear Jo(adpupa_winds_220) = 3.2593295738966270e+04, nobs = 48714, Jo/n = 6.6907451120758443e-01, err = 1.6247685156989053e+18'
Run 3: 
  0: Test Line: 'CostJo   : Nonlinear Jo(adpupa_winds_220) = 3.2592904892550017e+04, nobs = 48714, Jo/n = 6.6906648792031076e-01, err = 2.2988790520072753e+18'
Run 4: 
  0: Test Line: 'CostJo   : Nonlinear Jo(adpupa_winds_220) = 3.2596883201982317e+04, nobs = 48714, Jo/n = 6.6914815457532362e-01, err = 2.7916051027288340e+00'
Run 5: 
[...]

PR #346, opened 2025-03-28
Increase ctest tolerance and fix ctest error on Hercules

SamuelDegelia-NOAA:
## Bug Fix Description
Ctests are failing on Hera for fresh clone and builds with a `TestReferenceIntegerMismatchError` (mesonet wind ob count different by 2). It looks like something might have went wrong when I manually resolved the conflicts in #324. This PR updates the test references to work for the latest version of RDASApp. 

Thanks @chunhuazhou for catching this in #344. 

EDIT: The above updates to the test references were actually handled by #329 which was merged before this PR. The below updates are unique to this PR, though. 

Additionally, this PR increases the ctest float relative tolerance to capture some differences occurring across machines that are leading to ctest failure. An integer tolerance is also added in case QC checks are slightly difference on other machines (e.g., +/- 3 obs allowed now). 

1. Ctests fail on some machines due to small differences
2. GETKF solver ctest fails on Hercules due to bad job configuration

1. Ctests pass after updating the tolerance and updating configuration on Hercules

## Issue(s) addressed
None

None

## Checklist


SamuelDegelia-NOAA:
Here is my own ctest result on Hera but we can wait for @rrfsbot to confirm: 

```
+ ctest -j8
Test project /scratch2/BMC/zrtrr/Samuel.Degelia/RDASApp_fresh/RDASApp/build/rrfs-test
1/8 Test #8: rrfs_bufr2ioda_msonet .....................   Passed   31.18 sec
2/8 Test #5: rrfs_mpasjedi_2024052700_getkf_observer ...   Passed  255.08 sec
3/8 Test #2: rrfs_fv3jedi_2024052700_getkf_observer ....   Passed  267.04 sec
4/8 Test #1: rrfs_fv3jedi_2024052700_Ens3Dvar ..........   Passed  425.56 sec
5/8 Test #7: rrfs_mpasjedi_2024052700_bumploc ..........   Passed  455.70 sec
6/8 Test #4: rrfs_mpasjedi_2024052700_Ens3Dvar .........   Passed  516.35 sec
7/8 Test #3: rrfs_fv3jedi_2024052700_getkf_solver ......   Passed  250.53 sec
8/8 Test #6: rrfs_mpasjedi_2024052700_getkf_solver .....   Passed  361.81 sec

Label Time Summary:
mpi            = 2563.26 sec*proc (8 tests)
rdas-bundle    = 2563.26 sec*proc (8 tests)
script         = 2563.26 sec*proc (8 tests)

Total Test time (real) = 616.93 sec
+ exit 0
```

SamuelDegelia-NOAA:
This failure on Jet is a small `TestReferenceFloatMismatchError` and is related to my recent update to the reproducibility issue [here](https://github.com/NOAA-EMC/RDASApp/issues/327#issuecomment-2758414353). I expect the tests running on Hercules to fail for the same reason. 

As we add more observation types like the ZTD and upcoming ABI obs to our ctests, the differences between machines will continue to grow larger. At this point we could increase the `float relative tolerance` (an increase from 1e-3 to 1e-2 would let these tests on Jet pass). But these issues will likely continue as we go forward. Since tracking down machine differences like this could take lots of time and is likely outside of our expertise, we are leaning towards just considering ctests on a single machine. What would you think about this @guoqing-noaa?

guoqing-noaa:
> This failure on Jet is a small `TestReferenceFloatMismatchError` and is related to my recent update to the reproducibility issue [here](https://github.com/NOAA-EMC/RDASApp/issues/327#issuecomment-2758414353). I expect the tests running on Hercules to fail for the same reason.
> 
> As we add more observation types like the ZTD and upcoming ABI obs to our ctests, the differences between machines will continue to grow larger. At this point we could increase the `float relative tolerance` (an increase from 1e-3 to 1e-2 would let these tests on Jet pass). But these issues will likely continue as we go forward. Since tracking down machine differences like this could take lots of time and is likely outside of our expertise, we are leaning towards just considering ctests on a single machine. What would you think about this @guoqing-noaa?

Yes, I am good with this tolerance error.  Is it okay to increase the tolerance to 1e-2? 
It is still good to be able to use different machines to do ctest, such as when Hera is down or has a stability issue, we can use other machines to do ctests. But we can just let it go if it is not a critical error. What do you think?

guoqing-noaa (dismissed):

SamuelDegelia-NOAA:
@guoqing-noaa I think increasing the relative tolerance to 1e-2 should be okay. It will still allow us to capture larger differences that might be problematic. I can go ahead and update the tolerance in this PR (I'll change the title too since this wouldn't really be considered a hotfix anymore). 

And yes, I think that keeping the ctests available on other machines is okay. And we can merge things as long as we check that the failures are expected (small reference mismatch error).

guoqing-noaa:
One thing I am curious about is that Jet and Hera are very similar in the computer CPUS, architectures, but the ctests have different behaviors.

SamuelDegelia-NOAA:
> One thing I am curious about is that Jet and Hera are very similar in the computer CPUS, architectures, but the ctests have different behaviors.

The difference for the failed `rrfs_fv3jedi_2024052700_getkf_solver` ctest was a 0.16% difference in the maximum `water_vapor_mixing_ratio_wrt_moist_air ` for the analysis mean. It seems like a small difference, but I am not sure whether we would expect smaller differences given the similar machine architecture? This is a bit beyond my expertise now. 

It seems that we see these reference mismatch failures more often for GETKF ctests. Maybe small differences in the hofx are amplifying the analysis differences given the very large number of members (13 modulated members x 30 real members = 390 hofx calculations). Just brainstorming here, but maybe this is part of the issue?

rrfs-test/validated_yamls/templates/basic_config/fv3jedi_en3dvar.yaml
```diff
@@ -107,5 +107,6 @@ output:
 test:
   reference filename: testoutput/rrfs-fv3jedi-ens3dvar.ref
   test output filename: rrfs-fv3jedi-ens3dvar.out
-  float relative tolerance: 1.0e-3
+  float relative tolerance: 1.0e-2
```
guoqing-noaa, line 110: What's the difference between "relative tolerance" and "absolute tolerance"?
It looks like we still have a very small "absolute tolerance"? :)

SamuelDegelia-NOAA: Absolute tolerance is the absolute difference between test and reference values needed to pass, while relative tolerance is the percent difference. 

We are often seeing these failures on very small values like for mixing ratio or cloud liquid water. So for those types of failures, relative tolerance is going to be more important. 

Also, the comparison passes if either the relative check **or** absolute check pass. So if the percent difference is within our tolerance but the absolute tolerance is not, then the ctest still passes. Basically, increasing the relative tolerance here should be able to solve our failures.

guoqing-noaa: Great. Thanks a lot for the explanation!

SamuelDegelia-NOAA:
Agh the failures are neverending! That failed getkf test on Hercules is the same error seen in #329 where the test never starts: 

```
srun: error: Unable to allocate resources: Requested node configuration is not available
<end of output>
Test time =   0.02 sec
```

We request `--ntasks-per-node=8` for that test and we don't do that for others. I'll work on solving this.

ShunLiu-NOAA (dismissed):

SamuelDegelia-NOAA:
We can hold off on merging this until I fix the Hercules job configuration issue. Not sure why this just became a problem today.

SamuelDegelia-NOAA:
Okay I fixed the weird issue with the job configuration on Hercules so this PR should be good to go again.

guoqing-noaa (approved):

PR #351, opened 2025-04-01
Add the YAML linter and modify YAML files accordingly in order to pass the linting.

guoqing-noaa:
## Description
This PR adds the standard industry YAML linter to RDASApp. It will help adhere to a consistent style, make YAML files more readable and maintainable, it also help catch some syntax errors and unexpected typos, etc.

As a starting point, we relax the YAML check rules and ignore the inconsistent indentation, empty spaces, etc. 
We will work toward the strict rules incrementally. 

Also, the current `norms.yaml` and `unitests.yaml` look like not working with the current situation, so I temporarily disabled them. Let me know if I understand this incorrectly.

## Issue(s) addressed
None

None

## Checklist


rrfs-test/validated_yamls/templates/obtype_config/adpsfc_airTemperature_183.yaml
```diff
@@ -59,7 +59,6 @@
              value: is_not_valid
            where operator: or
            action:
-             name: reject
```
delippi: This along with all the other updates to the prepbufr obs space yamls are already in #326. The only exception is the proflr trailing white space.

guoqing-noaa: @delippi Thank you for making the updates in PR #326.  I appreciate it. The problem is that I have to make changes in this PR as well, otherwise, the Linting will fail with errors.

delippi: Can it be merged without passing all linting or make this PR dependent on 326?

guoqing-noaa: I would suggest we merge #326 as early as possible. It works for FV3JEDI and is commented out for MPASJEDI by [commentQC.py](https://github.com/NOAA-EMC/RDASApp/blob/develop/rrfs-test/validated_yamls/commentQC.py) (so no troubles with a merge). I don't think we have to hold 326 for longer time.

delippi: > It works for FV3JEDI and is commented out for MPASJEDI (so no troubles with a merge).

What do you mean by this? My update will break it for FV3-JEDI until Hui's PR is actually merged into ufo and then RDASApp synced to that ufo update. Simply making work around code patches in rrfs-workflow doesn't fix the problem. It will fail for RDASApp and I will not be able to make ctest references for RDASApp until then.

I suggest we continue to see if we can help Hui's PR move a bit quicker, then update RDASApp accordingly. This PR isn't critical and could wait until those things fall in line. We already have the work around code patch with Hui's PR so there is no hold on critical tests in that regard. We can also work around to get the correct yamls in my PR326. In fact, all that needs to be done is replace the `rrfs-workflow/sorc/RDASApp` with this build: `/scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2/PRs/RDASApp.qsat2`

With that there would be no need to build or make work around code patches to RDASApp I've also linked my  obspace yamls in that build as well.

ush/linter_yamlcheck.sh
SamuelDegelia-NOAA, line 1: Does this tool allow us to use the yaml lintering without waiting for the automatic checking in Github?

guoqing-noaa: Yes!

But we relax the rules a lot at this stage. Need more efforts to get to the regular YAML code style. I can put a few examples in this page soon.

SamuelDegelia-NOAA:
I'm good to go with this PR once the conflicts are resolved - I just would request that one set of ctests are run anytime the obs space yamls are changed. Thanks @guoqing-noaa!

delippi:
> I'm good to go with this PR once the conflicts are resolved - I just would request that one set of ctests are run anytime the obs space yamls are changed. Thanks @guoqing-noaa!

I agree. I broke the ctests with #326 hence part of my follow up PR #368. I made a cosmetic change (double quotes around ioda.nc filenames) and didn't rerun the ctests. If we can get 368 in before this, that would fix the ctests to run for this PR.

SamuelDegelia-NOAA:
Oh good point @delippi, yeah we should wait on #368.

ShunLiu-NOAA:
@guoqing-noaa Since https://github.com/NOAA-EMC/RDASApp/pull/368 is mergeed, could you please update this PR?

delippi (approved):
All the changes look good to me. I like this linter functionality as it keeps everything a bit tidier.

ShunLiu-NOAA (approved):

PR #352, opened 2025-04-02
bugfix in longitude valid range (0, 360)

HuiLiu-NOAA:
* [Bugfix](?expand=1&template=bugfix_pr.md)
Fixed the longitude valid range from (-180, 180) to (0, 360) after the change of the longitude values.  This bug caused masked longitude after the "offline_domain_check".  Now, this issue is resolved, and the "offline_domain_check" can be reverted to its original form.

### DIFF
--- a/rrfs-test/IODA/python/bufr2ioda_ztd.py
+++ b/rrfs-test/IODA/python/bufr2ioda_ztd.py
@@ -278,7 +278,7 @@ def bufr_to_ioda(config, logger):
 
     obsspace.create_var('MetaData/longitude', dtype=lon.dtype, fillval=lon.fill_value) \
         .write_attr('units', 'degrees_east') \
-        .write_attr('valid_range', np.array([-180, 180])) \
+        .write_attr('valid_range', np.array([0, 360])) \
         .write_attr('long_name', 'Longitude') \
         .write_data(lon)
 


guoqing-noaa (approved):

ShunLiu-NOAA (approved):

ShunLiu-NOAA:
@HuiLiu-NOAA Do we have to revert the changes in "offline_domain_check" ? Will the "offline_domain_check" handle your update without further modification?

SamuelDegelia-NOAA:
It might just be easiest for Hui to change the offline codes back using this PR. I am fine with that since this PR fixes the original problem while the changes to the offline codes were workarounds. I do not want to revert #336 because that PR also added attributes for tracking the settings used in the offline domain check. 

For `offline_domain_check.py` and `offline_add_var_to_ioda.py`, we would change

```
        if var in ['latitude', 'longitude']:
            g.variables[var][:] = invar[:][:].data
        else:
            g.variables[var][:] = invar[:][:]
```

to 

```
        g.variables[var][:] = invar[:][:]
```

HuiLiu-NOAA:
> It might just be easiest for Hui to change the offline codes back using this PR. I am fine with that since this PR fixes the original problem while the changes to the offline codes were workarounds. I do not want to revert #336 because that PR also added attributes for tracking the settings used in the offline domain check.
> 
> For `offline_domain_check.py` and `offline_add_var_to_ioda.py`, we would change
> 
> ```
>         if var in ['latitude', 'longitude']:
>             g.variables[var][:] = invar[:][:].data
>         else:
>             g.variables[var][:] = invar[:][:]
> ```
> 
> to
> 
> ```
>         g.variables[var][:] = invar[:][:]
> ```

Sounds good!

ShunLiu-NOAA:
@HuiLiu-NOAA Could you please take a look at "failing check" in this PR?

HuiLiu-NOAA:
> @HuiLiu-NOAA Could you please take a look at "failing check" in this PR?

@SamuelDegelia-NOAA : could you take a look at the failure related to ecbuild? Thanks.

guoqing-noaa:
> @HuiLiu-NOAA Could you please take a look at "failing check" in this PR?

@ShunLiu-NOAA The failed check is not due to this PR.  But the github action itself does not work.
I suggest we remove those github actions that are no longer needed by RDASApp (we have ctests running on Hera/Jet/Hercules) and set up useful ones as in PR #351

SamuelDegelia-NOAA:
It seems like merging with the upstream fixed the github action issue. But yes I agree that we should disable these existing actions and use those in #351 once it is merged.

Issue #345, opened 2025-03-28
How JEDI handles missing IODA files?


guoqing-noaa:
@HuiLiu-NOAA @spanNOAA and I had a conversation on how to modify the JEDI source codes to handle missing IODA files. We have not reached a solution yet. But we have some clues to move forward.

I will update as we progress on this.

guoqing-noaa:
We found the magic:
```
           engine:
             type: H5File
             obsfile: data/obs/ioda_aircar.nc
             missing file action: "warn"
```
This will let JEDI continue running if the ioda_aircar.nc file is missing

SamuelDegelia-NOAA:
That is great, thanks for finding this @guoqing-noaa. Do you know if using this option still writes out an empty `jdiag` file if the IODA file is missing?

HuiLiu-NOAA:
Per the IODA code, JEDI will create an empty ioda input structure in the memory if the input ioda file is not available (not exist) when the option is set to: 
missing file action: "warn"
So, JEDI may generate an empty jdiag file as if a physical empty ioda file exists. @guoqing-noaa can check if this is the case.

guoqing-noaa:
@SamuelDegelia-NOAA @HuiLiu-NOAA 
No, no output ioda files (i.e jdiag files in our RRFS world)

HuiLiu-NOAA:
> [@SamuelDegelia-NOAA](https://github.com/SamuelDegelia-NOAA) [@HuiLiu-NOAA](https://github.com/HuiLiu-NOAA) No, no output ioda files (i.e jdiag files in our RRFS world)

Good to know.  Thanks @guoqing-noaa.

guoqing-noaa:
@HuiLiu-NOAA @SamuelDegelia-NOAA 
My bad. I don't know why my previous case did not generate empty jdiag files (maybe because I only tested one observer). But when I ran this in rrfs-workflow with other observers, it did generate empty jdiag files.

HuiLiu-NOAA:
Yes, that may be the reason/case. When many obs types are assimilated together, the total obs counts are not zero.  This may be tested/confirmed for one observer only in the workflow?

guoqing-noaa:
> Yes, that may be the reason/case. When many obs types are assimilated together, the total obs counts are not zero. This may be tested/confirmed for one observer only in the workflow?

Thanks for the input. Sorry, I should not use the word "standalone".  I did test one observer in the workflow, but only tried one cycle.

[...]

PR #355, opened 2025-04-06
add the "missing file action" key into yaml files to handle possible missing ioda files in operation

guoqing-noaa:
## Description

As documented in issue https://github.com/NOAA-EMC/RDASApp/issues/345, JEDI provides two `missing file action` options: `warn` and `error`:

`warn`: Issue a warning, construct an empty ObsSpace object, and continue execution. This action is typically applicable to operations where you want the job to forge ahead despite a missing file.

`error`: Issue an error, throw an exception, and quit execution. This action is typically applicable to research and development where you want to be immediately notified when a file is missing.

This PR is to add `missing file action: "warn"` into each `obsdatain.engine` so that we are not bothered by a crash due to missing some ioda observation files.

## Issue(s) addressed
Resolves/Results are documented in Issue #345 

none

## Checklist

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
@@ -9,6 +9,7 @@
            engine:
              type: H5File
              obsfile: data/obs/ioda_abi_g16.nc
+             missing file action: "warn"
          obsdataout:
            engine:
              type: H5File
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g18.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g18.yaml
@@ -9,6 +9,7 @@
            engine:
              type: H5File
              obsfile: data/obs/ioda_abi_g18.nc
+             missing file action: "warn"
          obsdataout:
            engine:
              type: H5File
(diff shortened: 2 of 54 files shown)


SamuelDegelia-NOAA:
Good find, thanks for adding this option. Could we run the ctests on at least one machine to make sure that this doesn't break anything?

guoqing-noaa:
> Good find, thanks for adding this option. Could we run the ctests on at least one machine to make sure that this doesn't break anything?

Sure. Thanks!

SamuelDegelia-NOAA (approved):

hu5970 (approved):
Looks good

delippi (approved):
LGTM. I just hope we can find the solution to not having to comment out the duplicate check! This should be here regardless.

hu5970:
@delippi the duplicate check is not related to this PR, right?

delippi:
> @delippi the duplicate check is not related to this PR, right?

This PR <ins>does not</ins> make any changes to the duplicate check... which is good! 

I noticed that for rrfs-workflow `YAML_GEN_METHOD=1` will be using this option to handle missing ioda files, but the duplicate check currently needs turned off otherwise it will crash. This is because we have a variable `longitude_latitude_pressure` that is added adhoc to the ioda file and it seems this function has trouble with that. It is fine, for now, to comment the duplicate check as long as you are only assimilating sounding or aircraft data. It will affect anything else (including observations in monitor mode).

The bottom line: In order to fully use this option without the need to copy empty ioda files before running bufr2ioda, we need a better solution for the duplicate check and handling that extra variable.

guoqing-noaa:
@delippi Thanks for the discussion!

We will make efforts to make "temporal thinning" work as expected.
I have some conversations with JCSDA folks to address any issues we've got so far. I hope we will have some updates on this not too long.

PR #368, opened 2025-04-16
[feature & bugfix] Add passivate options and fixes associated with ctests 

delippi:
## Description

- This is a follow up PR to #326 in which the filter block to optionally passivate (monitor) observations is created for the prepbufr observation types.
- While making this PR, it was found a bug was unintentionally introduced to the ctests with PR #326 and thus all ioda.nc file names will have double quotes and the gen_yaml_ctest.sh now properly handles that.

All of the added filter blocks are identical which the exception of msonet_winds_288. This needed a special filter block to correctly passivate options which checks `if QCflagsData/windEastward  == 0`. This special case is because of the use of `AcceptLists` rather than only `RejectLists`. All filter blocks are currently set to `accept` by default which does not change results.

## Checklist

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
@@ -41,6 +41,13 @@
          # ------------------
          # stationPressure (188)
          # ------------------
+         # Accept, reject, or passivate (monitor)
+         - filter: Perform Action
+           filter variables:
+           - name: stationPressure
+           action:
+             name: accept # accept, reject, passivate
+
          # Reject all obs with QualityMarker > 3
          - filter: RejectList
            filter variables:
--- a/rrfs-test/validated_yamls/gen_yaml_ctest.sh
+++ b/rrfs-test/validated_yamls/gen_yaml_ctest.sh
@@ -71,7 +71,7 @@ for basic_config in "${!basic_configs[@]}"; do
             new_path=$(echo "$int_path" | sed "s/solver/observer/gI")
             obs_filename_new="obsfile: ${new_path}"
             # Old obs file name to replace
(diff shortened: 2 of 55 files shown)


SamuelDegelia-NOAA (approved):
Thanks for the quick fix with adding the quotes. And the passivate addition looks good to me!

ShunLiu-NOAA (approved):

rrfs-test/validated_yamls/templates/obtype_config/msonet_stationPressure_188.yaml
```diff
@@ -41,6 +41,13 @@
          # ------------------
          # stationPressure (188)
          # ------------------
+         # Accept, reject, or passivate (monitor)
+         - filter: Perform Action
+           filter variables:
+           - name: stationPressure
+           action:
+             name: accept # accept, reject, passivate
+
```
guoqing-noaa, line 50: I would suggest we comment out all monitor block be default.
Although the current form does not affect run results, but it takes extra time to loop through all observations in the obs space to mark the "accept" flag, which can be avoided, especially when there are lots of observations.

delippi: Thanks for the suggestion and it is good to test this. 

I just ran two tests with the workflow (repeating the same 00Z cycle): one with all prepbufr obs set to "accept" and one with each of those "accept blocks" commented out. The run times were basically identical (within 3 seconds), so it doesn't seem to add any extra time to keep them uncommented--this difference is within the usual variance of run times for repeated runs.

Leaving them uncommented does make it easier to toggle on and off, since you only need to change a single line.

I'm not sure how this applies to radiance DA. That might need additional testing.

guoqing-noaa: Thanks for testing the timing.
Although it is not preferred to do it this way, there is no evident harm either. Thanks!

guoqing-noaa (approved):

Issue #350, opened 2025-03-31
ctests creation for rdasapp on wcoss2


TingLei-NOAA:
The RDASapp PR https://github.com/NOAA-EMC/RDASApp/issues/343  added the option for RDASapp building on wcoss2 following setup with GDASapp. 

The remaining issue is how to create (if needed) ctests when, in PBS on wcoss2, there is no simple equivalent command for srun in slurm.    
One way is to add a wrapper mimicking srun.  But I haven't found such treatments in GDASapp. 
This issue is to help document discussion and efforts on this problem . 

@RussTreadon-NOAA and @CoryMartin-NOAA ,
On wcoss2,  how are the ctests dealt with?  Any suggestions/clarifications are welcome!

RussTreadon-NOAA:
We currently run g-w CI to test GDASApp on WCOSS2

TingLei-NOAA:
> We currently run g-w CI to test GDASApp on WCOSS2

@RussTreadon-NOAA  Thanks a lot. I will look at that and discuss with other developers on how to proceed for RDASapp.

SamuelDegelia-NOAA:
When building RDASApp on WCOSS2, `RDASApp/build/rrfs-test/CTestTestfile.cmake` changes `srun` to `mpiexec`. For example, here is the line that defines the `rrfs_mpasjedi_2024052700_Ens3Dvar` ctest: 

```
add_test([=[rrfs_mpasjedi_2024052700_Ens3Dvar]=] "/opt/cray/pals/1.3.2/bin/mpirun" "--time=00:30:00" "-n" "160" "/lfs/h2/emc/da/noscrub/Samuel.Degelia/RDASApp/build/bin/mpasjedi_variational.x" "rrfs_mpasjedi_2024052700_Ens3Dvar.yaml")
```

When I try and run that ctest I get an error about not recognizing the `--time` option. Removing that then leads to the error: 

```
1: No host list provided
1/1 Test #1: rrfs_mpasjedi_2024052700_Ens3Dvar ...***Failed    0.08 sec
```

Is it possible to submit a job on WCOSS2 via `mpirun` directly? Or does everything need to be submitted via a job card? If it is the latter, then maybe the easiest option would be to create job cards in each ctest run directory and then have the `add_test` call in `CTestTestfile.cmake` just submit the jobs.

TingLei-NOAA:
[...]

PR #361, opened 2025-04-09
Sync submodules with upstream repositories (April 2025)

SamuelDegelia-NOAA:
## Overview
This PR updates the RDASApp submodules to align with the latest changes in their respective upstream repositories. Each submodule has been synchronized to the head of their `develop` branches.

#### 1. Submodule Updates (required)
Here are the updated hashes for each submodule:

```
bufr-query:      a4cebcf -> 8d9c0bf
fv3-jedi:        136dfb9 -> d362a1f
fv3-jedi-lm:     3254bb8 -> 5994c38
gsibec:          c8ac58d -> d692989
ioda:            9e6518f -> 9328bdf
iodaconv:        6f87a0f -> 4ee7adc
mpas:            fba865e -> a07354c
mpas-jedi:       a66f8ea -> c8ddad7
oops:            0efe7bd -> 72bdde1
saber:           fd794ae -> 812d0ee
ufo:             0876a4b -> cd2ab0a
vader:           14e8bce -> 30e5a43
```

Per #359, it was found that the mpas-jedi yamls needed to be updated with new state variable names (also see [mpas-jedi PR #1040](https://github.com/JCSDA-internal/mpas-jedi/pull/1040)). Only the yamls needed to be updated to get things working again. No fix files needed to be changed since we copy those from the mpas-jedi `sorc` directory. See #359 for a list of all variable names that were updated.

This PR also incorporates changes to fv3-jedi that remove the D grid winds. Those state variables were removed from the yaml and the field metadata file was correspondingly updated (`data/fieldmetadata/tlei-gfs-restart.yaml`). 

Finally, I also reorganized the ctest job configuration in `rrfs-test/CMakeLists.txt` to try and clean that file up. The updated EnVar tests were failing due to an OOM message so I had to use more nodes for the tests on all machines (something to note when things are updated in rrfs-workflow). After these updates, all rrfs-test ctests are passing on the supported machines. 

**Note: we will still need to update the static B for mpas-jedi due to these variable name changes.**  

#### 2. Staged `ctest-data` Updates (if applicable)
The staged JCSDA test data are also updated here: 

```
fv3-jedi-data:     2085be5 -> e1c444d
ioda-data:         25afc2c -> ad9a013
mpas-jedi-data:    bbe282 -> 0c6dabe
ufo-data:          fffd9fa -> 9c40f67
```

#### 3. Test Reference Updates (if applicable)
The test references for all tests are also updated here (except for the BUMP localization and bufr2ioda ctests). 

[...]

PR #377, opened 2025-04-25
Add wrapper script for running ctests on WCOSS2

SamuelDegelia-NOAA:
## Description

This PR adds a wrapper script `wrapper_ctest_pbs.sh` for running ctests on WCOSS2. The ctests here can not be submitted using `mpirun` through the command line as is normally done with the `ctest` command. Running `mpirun` for PBS on WCOSS2 expects different input than what is defined in our `CMakeLists.txt` file including a host list that is not possible to assign automatically. Instead we introduce this wrapper script that creates and submits a job card for each ctest. 

The `RDASApp/rrfs-test/CMakeLists.txt` script is then modified to call this wrapper script and its required arguments when running on WCOSS2. 

After this change, the ctests can be run on WCOSS2 using the same methods as the other machines (either using `ush/run_rrfs_ctests.sh` or running `ctest -j8` in `build/rrfs-test`). 

## Issue(s) addressed
#350

None

## Checklist


SamuelDegelia-NOAA:
After this PR the ctests can run on WCOSS2. The fv3-jedi tests are currently failing because we need to update the staged RDAS_DATA for changes in #361. We will do that once @TingLei-NOAA returns from leave on Monday. This PR will be left as a draft until then. But all mpas-jedi tests are passing!

```
[samuel.degelia@clogin08 RDASApp] $ pwd
/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp
[samuel.degelia@clogin08 RDASApp] $ cd ush
[samuel.degelia@clogin08 ush] $ ./run_rrfs_tests.sh 

+ cd /lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/ush/../build/rrfs-test
+ pwd
/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/build/rrfs-test
+ ctest -j8
Test project /lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/build/rrfs-test
    Start 6: rrfs_mpasjedi_2024052700_getkf_observer
    Start 4: rrfs_mpasjedi_2024052700_bumploc
    Start 5: rrfs_mpasjedi_2024052700_Ens3Dvar
1/8 Test #8: rrfs_bufr2ioda_msonet .....................   Passed   24.19 sec
2/8 Test #2: rrfs_fv3jedi_2024052700_getkf_observer ....***Failed   70.47 sec
3/8 Test #1: rrfs_fv3jedi_2024052700_Ens3Dvar ..........***Failed   71.49 sec
4/8 Test #3: rrfs_fv3jedi_2024052700_getkf_solver ......***Failed   74.76 sec
5/8 Test #4: rrfs_mpasjedi_2024052700_bumploc ..........   Passed  361.68 sec
6/8 Test #6: rrfs_mpasjedi_2024052700_getkf_observer ...   Passed  542.83 sec
    Start 7: rrfs_mpasjedi_2024052700_getkf_solver
7/8 Test #5: rrfs_mpasjedi_2024052700_Ens3Dvar .........   Passed  549.32 sec
8/8 Test #7: rrfs_mpasjedi_2024052700_getkf_solver .....   Passed  740.43 sec

63% tests passed, 3 tests failed out of 8

Label Time Summary:
rdas-bundle    = 2435.18 sec*proc (8 tests)
script         = 2435.18 sec*proc (8 tests)

Total Test time (real) = 1283.28 sec

The following tests FAILED:
          1 - rrfs_fv3jedi_2024052700_Ens3Dvar (Failed)
          2 - rrfs_fv3jedi_2024052700_getkf_observer (Failed)
          3 - rrfs_fv3jedi_2024052700_getkf_solver (Failed)
Errors while running CTest
Output from these tests are in: /lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/build/rrfs-test/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
+ exit 8


```

SamuelDegelia-NOAA:
Thanks @TingLei-NOAA for updating the staged `RDAS_DATA` on WCOSS2. The ctests are now passing for both fv3-jedi and mpas-jedi. I think things are good to go now, so I will open this PR back up for review. 

Ctest results from WCOSS2: 

```
[samuel.degelia@clogin08 RDASApp] $ cd RDASApp/ush
[samuel.degelia@clogin08 ush] $ ./run_rrfs_tests.sh RRFS-DEV

+ cd /lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/ush/../build/rrfs-test
+ pwd
/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/build/rrfs-test
+ ctest -j8
Test project /lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/build/rrfs-test
    Start 6: rrfs_mpasjedi_2024052700_getkf_observer
    Start 4: rrfs_mpasjedi_2024052700_bumploc
    Start 5: rrfs_mpasjedi_2024052700_Ens3Dvar
1/8 Test #8: rrfs_bufr2ioda_msonet .....................   Passed   58.81 sec
2/8 Test #4: rrfs_mpasjedi_2024052700_bumploc ..........   Passed  360.50 sec
3/8 Test #2: rrfs_fv3jedi_2024052700_getkf_observer ....   Passed  514.96 sec
4/8 Test #1: rrfs_fv3jedi_2024052700_Ens3Dvar ..........   Passed  567.38 sec
5/8 Test #5: rrfs_mpasjedi_2024052700_Ens3Dvar .........   Passed  578.96 sec
6/8 Test #6: rrfs_mpasjedi_2024052700_getkf_observer ...   Passed  685.51 sec
    Start 7: rrfs_mpasjedi_2024052700_getkf_solver
7/8 Test #3: rrfs_fv3jedi_2024052700_getkf_solver ......   Passed  450.25 sec
8/8 Test #7: rrfs_mpasjedi_2024052700_getkf_solver .....   Passed  708.95 sec

Label Time Summary:
rdas-bundle    = 3925.32 sec*proc (8 tests)
script         = 3925.32 sec*proc (8 tests)

Total Test time (real) = 1394.77 sec
+ exit 0
```

rrfs-test/CMakeLists.txt
```diff
@@ -96,8 +109,17 @@ if(MPAS_DYCORE)
               set(MPI_ARGS "${MPI_ARGS} --ntasks-per-node=8")
             endif()
          endif()
+	 if ("${MACHINE_ID}" STREQUAL "wcoss2")
+           configure_file(${CMAKE_SOURCE_DIR}/rrfs-test/ush/wrapper_ctest_pbs.sh ${CMAKE_CURRENT_BINARY_DIR}/wrapper_ctest_pbs.sh COPYONLY)
+	   set(COMMAND_NAME "${CMAKE_CURRENT_BINARY_DIR}/wrapper_ctest_pbs.sh")
```
TingLei-NOAA, line 114: can we have " set(COMMAND_NAME "${CMAKE_CURRENT_BINARY_DIR}/wrapper_ctest_pbs.sh $exe"? 
And, if it could be done that the mpi_tasks and ARGS_NAME don't need to be different while the wrapper_ctest_pbs.sh would take care of all changes inside it?  Namely, only COMMAND_NAME would be different and, maybe more processing inside the wrapper script are needed.

rrfs-test/CMakeLists.txt
```diff
@@ -137,6 +159,13 @@ if(MPAS_DYCORE)
       ### Bufr2ioda ctests ###
       ########################
 
+
+      if ("${MACHINE_ID}" STREQUAL "wcoss2")
+	 set(MPI_TASKS 0)
```
TingLei-NOAA, line 164: in relation to previous thoughts, it is better we could wrap all changes if possible to that wrapper script and don't need to keep this conditioning in other places, like here.   But if it is very difficult to do so for any reasons I am not aware of , the current form is ok to me.

rrfs-test/ush/wrapper_ctest_pbs.sh
```diff
@@ -0,0 +1,67 @@
...
+#PBS -q dev
+cd ${WORKDIR}
+module use ${RDASApp}/modulefiles
+module load RDAS/wcoss2.intel
+ulimit -s unlimited
+ulimit -v unlimited
+ulimit -a
+export OOPS_TRACE=0
+export OMP_NUM_THREADS=1
+export LD_LIBRARY_PATH="\${LD_LIBRARY_PATH}:${RDASApp}/build/lib"
```
TingLei-NOAA: a little confusing, why do we need "escape" for the first "LD_LIBRARY_PATH"?

rrfs-test/ush/wrapper_ctest_pbs.sh
```diff
@@ -0,0 +1,67 @@
...
+/opt/cray/pals/1.3.2/bin/mpirun -n ${NTASKS} -ppn ${PPN} -cpu-bind core "${RDASApp}/build/bin/${EXECUTABLE}" "${CONFIG_FILE}"
```
TingLei-NOAA: since the right modules have been loaded,  can we just run "mpirun" other than using the current full path form? the path could be changed in the future when the different modules could be used/upgraded.

rrfs-test/ush/wrapper_ctest_pbs.sh
```diff
@@ -0,0 +1,67 @@
...
+MPIRUN_EXIT_CODE=\$?
+echo "TEST_FINISHED_WITH_EXIT_CODE \$MPIRUN_EXIT_CODE" >> "${OUTFILE}"
+exit \$MPIRUN_EXIT_CODE
+EOF
+
+# Submit job and get Job ID
```
TingLei-NOAA: a general question, why should we submit the job through another generated script?  can't we just run it inside this wrapper script? 
And, for the current setup, the child job script has the name pbs_jobs.sh.  we might hope it would have an unique name by adding the job id and/or ctest names to it to prevent overwriting especially when the ctests are run in parallel.

rrfs-test/CMakeLists.txt
SamuelDegelia-NOAA: `MPI` in `ecbuild_add_test` needs to be set to 0 for WCOSS2 specifically. If we do not set it to zero, then the ctest tries to run mpirun which will fail. See here in `CTestTestfile.cmake`

`add_test([=[rrfs_fv3jedi_2024052700_Ens3Dvar]=] "/opt/cray/pals/1.3.2/bin/mpirun" "/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/build/rrfs-test/wrapper_ctest_pbs.sh" ...)`

But if we set `${MPI_TASKS}` to zero, then the ctest can just run `COMMAND_NAME` without running it through mpirun: 

`add_test([=[rrfs_fv3jedi_2024052700_Ens3Dvar]=] "/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/build/rrfs-test/wrapper_ctest_pbs.sh" ...)`

I will try using `set(COMMAND_NAME "${CMAKE_CURRENT_BINARY_DIR}/wrapper_ctest_pbs.sh $exe)` though, that should work.

rrfs-test/ush/wrapper_ctest_pbs.sh
SamuelDegelia-NOAA: Yeah the escape character should not be needed, I will change it.

rrfs-test/ush/wrapper_ctest_pbs.sh
SamuelDegelia-NOAA: That should be fine, I will change it.

rrfs-test/CMakeLists.txt
SamuelDegelia-NOAA: We currently set the number of nodes and tasks for the ctests in `CMakeLists.txt`, so it made sense to set 40 ppn and 4 nodes for WCOSS2 here as well (through `ARGS_NAME`). If we move this info to the wrapper, then we would have to change this info in multiple places if we decide to use a different number of tasks for the ctests. Thus I prefer to keep things as they are right now.

rrfs-test/ush/wrapper_ctest_pbs.sh
SamuelDegelia-NOAA: We could submit the job through the ctest (and not using this `wrapper_ctest_pbs.sh` script) as long as we use the `-Wblock=true` argument. Without this, the ctest will consider the test passing as soon as the job as submitted. We could do this: 

`add_test([=[rrfs_mpasjedi_2024052700_Ens3Dvar]=] "qsub" "-Wblock=true" "/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp/build/rrfs-test/pbs_job.sh")`

It doesn't make too much of a difference to me as both function the same. Is there a benefit of submitting it directly versus using a script?

SamuelDegelia-NOAA: Also, I can change the job file from `pbs_job.sh` to a name specific to the ctest. But `pbs_job.sh` is created inside the specific directory for each ctest, so there should be no problem with overwriting.

TingLei-NOAA: Interesting.  The "cmake" will use the exist code from the wrapper script after the latter finish its running, right?  The wrapper is submitting an interactive job,  I think. Otherwise, using a child script to submit the job wouldn't make a difference. Please let me know if I missed something in those processes.  I prefer using the wrapper script as exactly a wrapping script for "srun" rather than a driver .  But this might be some stylish preferences.  So, current practice is ok when the child scripts' names are treated . 
An update: just be reminded that those child scripts would be created in their own dir. So that is fine now.

SamuelDegelia-NOAA: @TingLei-NOAA You mention using the wrapper script exactly as a wrapping script for `srun` rather than a driver. Are you referring to this method I described in my last response where the ctest directly runs qsub on a job file? 

I have tried using the method instead of the ctest running a parent script that builds and submits the job. While it's possible to pass arguments to a PBS job using `qsub -v arg1=val1 arg2=val2`, these variables cannot be used to fill in `#PBS` directive lines, as PBS processes those before the shell interprets variables. Because of this limitation, directly passing arguments won't affect core job configurations like node count, wall time, account name, or output filenames. These settings would have to be hard coded in the pbs job file itself unless we wanted to do some type of `sed` replacement which would make `CMakeLists.txt` even more complex. 

Using a wrapper script to generate the PBS job allows all configuration values — including those in #PBS lines — to be dynamically set. We can thus edit them in one place (`CMakeLists.txt`) where these options are set for the ctests run on other machines. In my opinion, this is a more maintainable approach.

TingLei-NOAA: @SamuelDegelia-NOAA   To directly use qsub to replace srun would be difficult to process the command line options. I think using a wrapping script as you does now is a good way to do this.   I just hope this wrapping script would , on the one hand, do more to allow almost all changes be done in it , and on the other hand, it would  do less that it would not create and run  another child script on the fly.   It is mainly about style preference and you could keep your current design.

TingLei-NOAA: @SamuelDegelia-NOAA   To directly use qsub to replace srun would be difficult to process the command line options. I think using a wrapping script as you does now is a good way to do this.   I just hope this wrapping script would , on the one hand, do more to allow almost all changes be done in it , and on the other hand, it would  do less that it would not create and run  another child script on the fly.   It is mainly about style preference and you could keep your current design.

rrfs-test/CMakeLists.txt
SamuelDegelia-NOAA: @TingLei-NOAA It did not work to move the other arguments under `COMMAND_NAME`. That is because the ctest tries to find an executable that is named exactly what is passed into `COMMAND` under `ecbuild_add_test`. So if we try and set all the arguments there, then the ctest produces this error: 

```
[samuel.degelia@clogin06 rrfs-test] $ ctest -VV -R rrfs_mpasjedi_2024052700_Ens3Dvar 
UpdateCTestConfiguration  from :/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp_rebuild/RDASApp/build/rrfs-test/DartConfiguration.tcl
UpdateCTestConfiguration  from :/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp_rebuild/RDASApp/build/rrfs-test/DartConfiguration.tcl
Test project /lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp_rebuild/RDASApp/build/rrfs-test
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 5
    Start 5: rrfs_mpasjedi_2024052700_Ens3Dvar
Could not find executable /lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp_rebuild/RDASApp/build/rrfs-test/wrapper_ctest_pbs.sh rrfs_mpasjedi_2024052700_Ens3Dvar mpasjedi_variational.x 40 4
Looked in the following places:
/lfs/h2/emc/da/noscrub/samuel.degelia/RDASApp_rebuild/RDASApp/build/rrfs-test/wrapper_ctest_pbs.sh rrfs_mpasjedi_2024052700_Ens3Dvar mpasjedi_variational.x 40 4
...
```

As such we will need to stick with the separate `ARGS_NAME` list passed into `ecbuild_add_test` that needs to differ between WCOSS2 and other machines.

SamuelDegelia-NOAA:
@TingLei-NOAA I think I found a solution that will work for both of us. See the new version of `wrapper_ctest_pbs` from my most recent commit. It now submits the job directly from the wrapper script without creating individual job scripts. This is achieved using `qsub -Wblock=true <<EOF` instead of `cat <<EOF > "$PBS_SCRIPT"`. Let me know what you think. I tested this solution and it is working - all the ctests pass. I

I also addressed your other comments to the best of my ability. But sadly, we need to keep the different `ARGS_NAME` list on WCOSS2 per my comment [here](https://github.com/NOAA-EMC/RDASApp/pull/377#discussion_r2067424638).

TingLei-NOAA:
@SamuelDegelia-NOAA  That is great! Thanks for taking the time to polish the changes.

rrfs-test/CMakeLists.txt
TingLei-NOAA: OK, That is fine.  If we want to dig more on this, that can be done in another PR.

TingLei-NOAA (approved):
Thanks!

ShunLiu-NOAA (approved):

PR #384, opened 2025-05-13
Add `-t` build option to skip linking in and generating ctest data 

SamuelDegelia-NOAA:
## Description
Currently, RDASApp fails to build if the ctest data are not available. To resolve this, we introduce a build option to prevent linking these data and from building the super yamls. 

This option can be used in rrfs-workflow since the ctest data are not needed there. This will also be useful with testing rrfs-workflow on machines where we have not staged this test data. Some developers are currently testing the workflow on Derecho and have run into the problem. 

The default behavior is to link/generate the ctest data, but this can be disabled now with: 

```
./build.sh -t NO
```

## Issue(s) addressed
None

None

## Checklist


ShunLiu-NOAA (approved):

TingLei-NOAA (approved):

guoqing-noaa (approved):

PR #395, opened 2025-05-31
update build.sh - source init.sh; debug compiling; etc

guoqing-noaa:
## Description
update build.sh:
1. source init.sh
Previously init.sh is sourced during the ctest setup step. 
Now the `-t NO` option enables skipping the ctest setup step. In this situation, the fix/ directory is not initialized during the build process and hence the CMake system will download crtm coefficient files from internet (`bundle/test-data-release/crtm_coefficients_2.4.1_skylab_4.0.tar.gz`), which slows the build process.
This is NOT expected.
This PR sources init.sh in the beginning of `build.sh` to skip download crtm fix files.

2. add "-d" to allow compiling in the debug mode
Add the "-d" command line option to facilitate compiling RDASApp in the debug mode

3. modify the usage example to match the latest situation

## Issue(s) addressed
None

None

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -11,12 +11,13 @@ START=$(date +%s)
 dir_root="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 source $dir_root/ush/detect_machine.sh
+source $dir_root/ush/init.sh
 
 # ==============================================================================
 usage() {
   set +x
   echo
-  echo "Usage: $0 -p <prefix> | -t <target> -h"
+  echo "Usage example: $0 -j <num> -m MPAS -t NO | -h"
   echo
   echo "  -p  installation prefix <prefix>       DEFAULT: <none>"
   echo "  -c  additional CMake options           DEFAULT: <none>"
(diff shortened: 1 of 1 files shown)


xyzemc (approved):
Thanks for fixing this.

TingLei-NOAA (approved):

TingLei-NOAA:
@guoqing-noaa  Do we expect all ctests including rrfs tests to pass when rdasapp is built in debug mode ?

SamuelDegelia-NOAA (approved):

guoqing-noaa:
> @guoqing-noaa Do we expect all ctests including rrfs tests to pass when rdasapp is built in debug mode ?

@TingLei-NOAA Good question!
I would guess it will take extra efforts to make sure all rrfs tests to pass for debugging mode rdasapp.

SamuelDegelia-NOAA:
I have tried running the ctests in debug mode before and experienced failures that pointed to obs error inflation routines. Someday it would probably be worth tracking those down.

TingLei-NOAA:
@guoqing-noaa  Thanks. In fact, I am wondering if it is required that those ctests would create identical results in debug mode and release mode. 
@SamuelDegelia-NOAA  Thanks for this info. Seems that is a certain bug to be fixed that was revealed in debug mode.

PR #400, opened 2025-06-07
Updates to bufr2ioda_satwnd.py

HuiLiu-NOAA:
This PR adds the fake variables: "height" and "stnelev" for SATWND, required by the ObsErrorFactorPressureCheck (hard-coded for height-coordinate OBS, but not used for pressure-coordinate OBS) for OBS error inflation filter. The calculation/type of the variable 'MetaData/prepbufrReportType' is also updated, so it can be used in the current RDASApp yaml.
The previous ObsType/ available in prepbufr files is also added in the SATWND Bufr-format data for convenience.  It would be more meaningful and straight forward to use the paramerter "WindCalculationMethod", rather than this ObsType in RDASApp yamls.

### DIFF
--- a/rrfs-test/IODA/python/bufr2ioda_satwnd.py
+++ b/rrfs-test/IODA/python/bufr2ioda_satwnd.py
@@ -129,21 +129,20 @@ def bufr_to_ioda(config, logger):
     # MetaData
     lat = r.get('latitude')
     lon = r.get('longitude')
-    lon[lon > 180] -= 360      # Convert to [-180,180]
+    lon[lon < 0] += 360      # Convert to [0,360]
 
     said = r.get('satelliteId')
     zen  = r.get('satelliteZenithAngle', type='float')
     freq = r.get('frequency', type='float')
     cen  = r.get('processingCenter', type='float')
-    swcm = r.get('windCalculationMethod')
-#   swcm = r.get('windCalculationMethod', type='float')
+    swcm = r.get('windCalculationMethod', type='float')
 
     pob = r.get('pressure', type='float')
     cor = r.get('correlation', type='float')
     vari= r.get('variation', type='float')
 #   surf= r.get('surface', type='float')
 
-#   Fake variables required by ObsFunction/ObsErrorFactorPressureCheck.cc etc (hard-coded for height coordinate OBS) 
+#   Fake variables for SATWND required by ObsFunction/ObsErrorFactorPressureCheck.cc etc (hard-coded for height coordinate OBS) 
     height = np.full_like(pob, fill_value=pob.fill_value, dtype=np.float32)
     stnelev = np.full_like(pob, fill_value=pob.fill_value, dtype=np.float32)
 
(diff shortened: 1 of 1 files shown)


delippi:
@HuiLiu-NOAA, is there any particular reason why we're not using the converters from GDASApp? I think I got `bufr2ioda_satwnd_amv_goes.py` working for Keenan. It had everything except timeOffset (not sure why), but I added that and I think it is working.

Here is my example run script if you want to check it (the .py and .json files are also in same location):
`/scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2/ioda_processing/python/run_satwnd.sh`

HuiLiu-NOAA:
> @HuiLiu-NOAA, is there any particular reason why we're not using the converters from GDASApp? I think I got `bufr2ioda_satwnd_amv_goes.py` working for Keenan. It had everything except timeOffset (not sure why), but I added that and I think it is working.

> 
> Here is my example run script if you want to check it (the .py and .json files are also in same location): `/scratch2/NCEPDEV/fv3-cam/Donald.E.Lippi/RRFSv2/ioda_processing/python/run_satwnd.sh`

I think both the GDASApp and RDASApp versions should work (and similar). I developed the RDASApp version from scratch (without knowledge of the existence of GDASApp version).
One difference is that the RDASApp version uses simpler input/output for RDAS rrfs_workflow.

delippi:
Okay I see.

They have a satwnd converter for each: ahi, avhrr, goes, leogeo, modis, seviri, viirs. I'm not sure if we need those all or if this one convers all those? I have a feeling all/most of our converters weren't just taken from GDASApp--and they should have been. I think we should try to use those if possible and make only minimal changes as necessary. Theoretically, the converters should be identical between systems. As we move toward python based converters, we can take this opportunity to try to unify better. Just my opinion.

HuiLiu-NOAA:
> Okay I see.
> 
> They have a satwnd converter for each: ahi, avhrr, goes, leogeo, modis, seviri, viirs. I'm not sure if we need those all or if this one convers all those? I have a feeling all/most of our converters weren't just taken from GDASApp--and they should have been. I think we should try to use those if possible and make only minimal changes as necessary. Theoretically, the converters should be identical between systems. As we move toward python based converters, we can take this opportunity to try to unify better. Just my opinion.

I agree we can learn from GDASApp for sure, but also be open for new opportunity/information for improvement for RDASApp and RRFS-workflow.

SamuelDegelia-NOAA:
I think it would be beneficial to align the converters used in RDASApp and GDASApp, if possible. Using a common converter would help ensure consistency in the obs generated across both systems, and it would make collaboration with the global and obsproc teams smoother when troubleshooting or refining the process. Would it be possible to apply the improvements you've made, such as the simplified input/output for rrfs-workflow, to the existing GDASApp converter?

HuiLiu-NOAA:
> I think it would be beneficial to align the converters used in RDASApp and GDASApp, if possible. Using a common converter would help ensure consistency in the obs generated across both systems, and it would make collaboration with the global and obsproc teams smoother when troubleshooting or refining the process. Would it be possible to apply the improvements you've made, such as the simplified input/output for rrfs-workflow, to the existing GDASApp converter?

Thanks.  That would be a good way to go. Will do that.

Issue #398, opened 2025-06-04
[bug]hofx missing value on certain point related to complex terrain


xyzemc:
## Current behavior (describe the bug)
> During the 3dEnVar retro experiment with assimilating GOES ABI radiance, there are some random cycle JEDI crashed because of the hofx has missing value at certain point. The error message looks like this:
 `"Exception: Assertion failed: hofx[jobs] != missing in applyFilter, line 130 of /scratch2/NCEPDEV/fv3-cam/Xiaoyan.Zhang/noscrub/JEDI/RRFSV2/workflow/v2.1.0/sorc/RDASApp/bundle/ufo/src/ufo/filters/BackgroundCheck.cc"`
For example, JEDI crashed at (also showing on map):
```
2024-05-05_05Z
obs_lat = 23.2534
obs_lon = -99.4921
and 
2024-05-06_03Z
obs_lat = 25.3558
obs_lon = -100.379
```

![Image](https://github.com/user-attachments/assets/d5906f13-a08a-438e-be87-23a1ce1b6419)

After investigating a few things, I found in crtm, there is 'pressure monotonicity check' for skipping the point to calculate radiance if the pressure profile is non-monotonicity. So at the above point listed, the pressure  increases up to 767.4989hPa  at level 58, but then drops slightly to 767.4648 hPa at level 59. That’s non-monotonic behavior. So the crtm skipped to calculate the radiance on this point, just keep it as missing value. The full pressure value for 59 levels are:
air_pressure =  [22.174291610717773 27.259740829467773 32.978172302246094 39.4063720703125
 46.60930633544922 54.70631408691406 63.900535583496094 74.31890869140625
 85.92999267578125 98.86579895019531 113.1568832397461 128.6026611328125
 145.17239379882812 162.8540802001953 181.5096893310547 200.9765625
 221.1318359375 241.92369079589844 263.54437255859375 285.90130615234375
 308.8238525390625 332.11944580078125 355.659423828125 379.2646179199219
 402.7129821777344 425.8136291503906 448.40472412109375 470.3207702636719
 491.3810119628906 511.426513671875 530.3797607421875 548.2594604492188
 565.1458129882812 581.1422119140625 596.3443603515625 610.8225708007812
 624.6162109375 637.7362060546875 650.172607421875 661.9048461914062
 672.9124755859375 683.1853637695312 692.7286987304688 701.5548706054688
 709.67919921875 717.1239013671875 723.92529296875 730.1093139648438
[...]

PR #404, opened 2025-06-12
Add Domain Check filter to remove the data with missing hofx

xyzemc:
* As described in (https://github.com/NOAA-EMC/RDASApp/issues/398), the JEDI crashed at some cycles caused by the missing hofx when assimilating satellite radiance data. This issue has been investigated and confirmed that it is originated from MPAS model forecast caused the inverse pressure at the lowest two model levels on the high altitude terrain. Before the actual bug being fixed, a QC filter is required to exclude this situation that could cause JEDI crash during the retro experiment cycles.
* This Domain Check filter has been added to abi_g16, abi_g18, atms_npp, atms_n20, and atms_n21 jedi yaml file.
 ```
# Toss data if the hofx has missing value
       - filter: Domain Check
         filter variables:
         - name: brightnessTemperature
           channels: *abi_g16_channels
         where:
         - variable:
             name: brightnessTemperature@HofX
             channels: *abi_g16_channels
           value: is_valid
           minvalue: 50
           maxvalue: 500
         action:
           name: reject


```

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml
@@ -269,6 +269,21 @@
            minvalue: 0.0
            maxvalue: 0.5
 
+       # Toss the data if the hofx has missing value
+       - filter: Domain Check
+         filter variables:
+         - name: brightnessTemperature
+           channels: *atms_n20_channels
+         where:
+         - variable:
+             name: brightnessTemperature@HofX
+             channels: *atms_n20_channels
+           value: is_valid
+           minvalue: 50
+           maxvalue: 500
+         action:
+           name: reject
+
        # Step 0-C: Assign Initial All-Sky Observation Error
        - filter: Perform Action
          filter variables:
--- a/rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml
@@ -248,13 +248,18 @@
                  channels: *atms_n20_channels
                coefs: [1, -1]
 
-       - filter: Bounds Check
+        # Toss the data if the hofx is missing or outside of range
+       - filter: Domain Check
          filter variables:
-         #- name: brightnessTemperature
-         - name: HofX/brightnessTemperature
+         - name: brightnessTemperature
            channels: *atms_n20_channels
-         minvalue: 50.00001
-         maxvalue: 449.99999
+         where:
+         - variable:
+             name: HofX/brightnessTemperature
+             channels: *atms_n20_channels
+           value: is_valid
+           minvalue: 50.00001
+           maxvalue: 449.99999
          action:
            name: reject
 
@@ -268,21 +273,6 @@
            flag all filter variables if any test variable is out of bounds: true
            minvalue: 0.0
            maxvalue: 0.5
-
-       # Toss the data if the hofx has missing value
-       - filter: Domain Check
-         filter variables:
-         - name: brightnessTemperature
-           channels: *atms_n20_channels
-         where:
-         - variable:
-             name: brightnessTemperature@HofX
-             channels: *atms_n20_channels
-           value: is_valid
-           minvalue: 50
-           maxvalue: 500
-         action:
-           name: reject
 
        # Step 0-C: Assign Initial All-Sky Observation Error
        - filter: Perform Action
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
@@ -225,6 +225,21 @@
            minvalue: 0.0
            maxvalue: 0.5
 
+       # Toss data if the hofx has missing value
+       - filter: Domain Check
+         filter variables:
+         - name: brightnessTemperature
+           channels: *abi_g16_channels
+         where:
+         - variable:
+             name: brightnessTemperature@HofX
+             channels: *abi_g16_channels
+           value: is_valid
+           minvalue: 50
+           maxvalue: 500
+         action:
+           name: reject
+
        # Surface type check
        # Toss channels 7,11-16 over land
        - filter: RejectList
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
@@ -225,18 +225,18 @@
            minvalue: 0.0
            maxvalue: 0.5
 
-       # Toss data if the hofx has missing value
+       #Toss data if the hofx is missing or outside of range
        - filter: Domain Check
          filter variables:
          - name: brightnessTemperature
            channels: *abi_g16_channels
          where:
          - variable:
-             name: brightnessTemperature@HofX
+             name: HofX/brightnessTemperature
              channels: *abi_g16_channels
            value: is_valid
-           minvalue: 50
-           maxvalue: 500
+           minvalue: 50.00001
+           maxvalue: 449.99999
          action:
            name: reject
 
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g18.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g18.yaml
@@ -225,7 +225,7 @@
            minvalue: 0.0
            maxvalue: 0.5
 
-       # Toss data if the hofx has missing value
+       # Toss data if the hofx is missing or outside of range
        - filter: Domain Check
          filter variables:
          - name: brightnessTemperature
@@ -235,8 +235,8 @@
              name: brightnessTemperature@HofX
              channels: *abi_g18_channels
            value: is_valid
-           minvalue: 50
-           maxvalue: 500
+           minvalue: 50.00001
+           maxvalue: 449.99999
          action:
            name: reject
 
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g18.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g18.yaml
@@ -225,6 +225,21 @@
            minvalue: 0.0
            maxvalue: 0.5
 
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by SamuelDegelia-NOAA** [comment]
Great change, thanks for fixing such an obscure issue @xyzemc. Do you know if adding this filter will change the results to warrant updating the ctest references?

**Comment by xyzemc** [comment]
> Great change, thanks for fixing such an obscure issue @xyzemc. Do you know if adding this filter will change the results to warrant updating the ctest references?

It should change the ctest reference. I will update the reference file once my cetest passed.

File: rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml [hunk 1]
```diff
@@ -269,6 +269,21 @@
            minvalue: 0.0
            maxvalue: 0.5
 
+       # Toss the data if the hofx has missing value
```
HaidaoLin-NOAA: HI Xiaoyan, nice job! It looks like for all atms yaml files, we already have bounds check for hofx/brightnessTemperature, lines 251-259 here, maybe we don't need another domain check for this?  If I understand correctly, the previous bounds check seems to has the same function as the new domain check.  Or maybe we could remove the previous bounds check and add this domain check? Thanks!  @xyzemc

SamuelDegelia-NOAA: Would adding `value: is_valid` to the existing Bounds Check work?

HuiLiu-NOAA:
For consistency, use:
- name: HofX/brightnessTemperature
instead of
 name: brightnessTemperature@HofX  ?

rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml
xyzemc: @HaidaoLin-NOAA You are right. ATMS already has the Bound check filters. I will retest it to see if it works same as the domain check that I added. I will keep one at the  end. Thank you very much for reminding me to check this.

xyzemc:
> For consistency, use:
> 
> * name: HofX/brightnessTemperature
>   instead of
>   name: brightnessTemperature@HofX  ?

@HuiLiu-NOAA Ok, I will change that.

rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml
xyzemc: @HaidaoLin-NOAA   I conducted a few more tests to verify if the 'Bound Check' in ATMS  yaml file lines 251-259. I used the single abi_g16 obs which has the missing hofx, and put the exact 'Bound Check' in ATMS to ABI yaml file. The test results showed this 'Bound Check' can not QC out this obs. In my understanding, this Bound Check as showing below
```
 - filter: Bounds Check
         filter variables:
         #- name: brightnessTemperature
         - name: HofX/brightnessTemperature
           channels: *atms_n20_channels
         minvalue: 50.00001
         maxvalue: 449.99999
         action:
           name: reject
```
Only work for the value of HofX/brightnessTemperature is valid, but not apply to the missing value. Also  address @SamuelDegelia-NOAA 's question 'value: is_valid' is not working  in Bound Check because the value: is_valid / not_valid construct is only supported in the Domain Check filter, not Bounds Check. So I will revise the ling 251-259 Bound Check in all atms yaml file to Domain Check to exclude both the invalid value and the value outside of 50-450.

xyzemc:
Ctest passed on Hera, no need to update the reference.
```
/scratch1/NCEPDEV/stmp2/Xiaoyan.Zhang/RDASApp_hofx/build/rrfs-test
+ ctest -j8
Test project /scratch1/NCEPDEV/stmp2/Xiaoyan.Zhang/RDASApp_hofx/build/rrfs-test
    Start 6: rrfs_mpasjedi_2024052700_getkf_observer
    Start 4: rrfs_mpasjedi_2024052700_bumploc
    Start 5: rrfs_mpasjedi_2024052700_Ens3Dvar
1/8 Test #8: rrfs_bufr2ioda_msonet .....................   Passed   32.00 sec
2/8 Test #2: rrfs_fv3jedi_2024052700_getkf_observer ....   Passed  175.67 sec
3/8 Test #6: rrfs_mpasjedi_2024052700_getkf_observer ...   Passed  246.19 sec
    Start 7: rrfs_mpasjedi_2024052700_getkf_solver
4/8 Test #5: rrfs_mpasjedi_2024052700_Ens3Dvar .........   Passed  338.15 sec
5/8 Test #4: rrfs_mpasjedi_2024052700_bumploc ..........   Passed  382.19 sec
6/8 Test #1: rrfs_fv3jedi_2024052700_Ens3Dvar ..........   Passed  433.34 sec
7/8 Test #3: rrfs_fv3jedi_2024052700_getkf_solver ......   Passed  262.63 sec
8/8 Test #7: rrfs_mpasjedi_2024052700_getkf_solver .....   Passed  371.16 sec

Label Time Summary:
mpi            = 2241.32 sec*proc (8 tests)
rdas-bundle    = 2241.32 sec*proc (8 tests)
script         = 2241.32 sec*proc (8 tests)

Total Test time (real) = 617.39 sec
+ exit 0

```

SamuelDegelia-NOAA (approved):
LGTM! Thanks again for fixing this bug.

ShunLiu-NOAA (approved):

rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml
guoqing-noaa: Interesting to see that `Domain check` rejects missing values by default while `Bounds check` does not.

HuiLiu-NOAA: yes, maybe the purpose of the filters are slightly different.

guoqing-noaa: > yes, maybe the purpose of the filters are slightly different.

Yes, interesting finding!

Previously I suggest we use `Domain Check` for QualityMarker:
https://github.com/NOAA-EMC/RDASApp/blob/develop/rrfs-test/validated_yamls/templates/obtype_config/aircar_airTemperature_133.yaml#L58-L68
```
         - filter: RejectList
           filter variables:
           - name: airTemperature
           where:
           - variable: QualityMarker/airTemperature
             is_in: 4-15
             value: is_not_valid
           where operator: or
           action:
             name: reduce obs space
```
But I did not get time to test it at that time.

guoqing-noaa: @HuiLiu-NOAA 

I think someone tried this
```
filter: Domain Check
where:
- variable:
    name: QualityMarker/airTemperature
  value: is_valid
  minvalue: 0
  maxvalue: 3
action:
  name: reduce obs space
```
Not sure why it did not work at that time. But it apparently works in @xyzemc's case.
I will revisit this when I get time.

HaidaoLin-NOAA: @xyzemc nice work and thanks for testing!  A curious question, when using the domain check, if we don't include this  line"value: is_valid", can the domain check remove the missing hofx values? Thanks!

HuiLiu-NOAA: How about comment out the 2 lines:
action:
  name: reduce obs space

xyzemc: @HaidaoLin-NOAA No. I have tested and confirmed that the missing hofx value won't be able to removed if we don't include the "value: is_valid"

HaidaoLin-NOAA: Thanks! @xyzemc

guoqing-noaa: @HuiLiu-NOAA 
I think the key may be the order of "value: is_vaild". 
I previously used:
```
  minvalue: 0
  maxvalue: 3
  value: is_valid
```
Based on @xyzemc’s test, putting `value: is_valid` before `minvalue` may be the ky.

@xyzemc if possible, could you try one test by putting `value: is_valid` **after** `maxvalue` to see whether it works or not?  Thanks!

xyzemc: @guoqing-noaa @HuiLiu-NOAA I have tested 'putting value: is_valid after maxvalue', and it works exactly as putting 'value: is_valid' before 'minvalue'. I think the order is doesn't matter, but matter to the 'filter' type, and 'value: is_valid' must under 'where'.

guoqing-noaa: @xyzemc Thanks a lot for the testing!
My previous test might NOT do things right. I will try it again sometime

HaidaoLin-NOAA: @xyzemc This is not important, but just an update.  @guoqing-noaa and I discussed the usage of bounds check and missing data.  I did some tests (using cris-fsr data) for using the bounds check to handle the missing hofx values (and other very high or low values).  You don't have to test it and I am good with your domain check to fix the problem.  Just updating another possible option to remove the missing hofx values using the bounds check.  
      - filter: Bounds Check
         filter variables:
         - name: brightnessTemperature
           channels: *cris-fsr_n20_channels
         test variables:
         - name: HofX/brightnessTemperature
           channels: *cris-fsr_n20_channels
         minvalue: 50.00001
         maxvalue: 449.99999
         action:
           name: reject

HuiLiu-NOAA: This is a very important point to learn about the rejection of missing values, I think, which is definitely worth of further testing/clarification here.

xyzemc: Thanks @HaidaoLin-NOAA for testing another option to remove the missing hofx. We may need to further discuss if we should unify all the satellite jedi yaml file to use one of Bounds Check or Domain Check.

PR #412, opened 2025-06-24
add back sglobs_t120_mpasjedi.yaml

guoqing-noaa:
## Description
We lost the single obs test. This PR brings it back

## Issue(s) addressed
None

None


ShunLiu-NOAA (dismissed):

SamuelDegelia-NOAA:
@guoqing-noaa We moved the single observation test yamls to `rrfs-test/example_yamls` which still exist. But those yamls have not been updated in a while and still use the old state variable names. Glad that they are finally being updated. 

Could you remove those old yamls in `rrfs-test/example_yamls`? And then we could move the new yaml added here into `rrfs-test/example_yamls` (that way we only use `testinput` for the ctests).

guoqing-noaa:
> @guoqing-noaa We moved the single observation test yamls to `rrfs-test/example_yamls` which still exist. But those yamls have not been updated in a while and still use the old state variable names. Glad that they are finally being updated.
> 
> Could you remove those old yamls in `rrfs-test/example_yamls`? And then we could move the new yaml added here into `rrfs-test/example_yamls` (that way we only use `testinput` for the ctests).

Can the whole `rrfs-test/example_yamls` be removed?  I did not follow up with it.

SamuelDegelia-NOAA:
> Can the whole `rrfs-test/example_yamls` be removed? I did not follow up with it.

We could, but I would prefer to use a separate directory for the single observation yaml other than `rrfs-test/testinput`. In my opinion, the `testinput` directory should only be used for the input to the ctests.

SamuelDegelia-NOAA (approved):
LGTM, thanks a ton for updating this yaml.

guoqing-noaa:
@SamuelDegelia-NOAA Thanks a lot for the review and suggestions!

Issue #427, opened 2025-08-13
RDASApp build failing due to incompatible Miniconda environment


keenaneure:
I attempted to build RDASApp on Hera, and I was getting crashes on the ./build.sh step around 50-60% completion. Based on guidance from Sam Degelia and output from the log, it appears that the compiler is trying to link libraries in /scratch3/BMC/wrfruc/hera/Miniforge3, which suggests that there is a different Python/Anaconda environment active when trying to build. The Python installation is taking precedence over the modules/RDAS.intel version.

Here is my log pathway on Hera:
/scratch3/BMC/wrfruc/eure/RDASApp/build.log

Running 'conda deactivate' before running './build.sh' fixes the issue, but I wanted to make note of this for any future issues others may have.

It is possible that this code *might* fix the problem, but I have not tested it:
# deactivate any virtualenv
if [ -n "$VIRTUAL_ENV" ]; then
    deactivate
fi

# deactivate conda environment
if [ -n "$CONDA_PREFIX" ]; then
    conda deactivate
fi

module purge

SamuelDegelia-NOAA:
Thanks for opening this issue @keenaneure. I've had this issue before too where the python associated with a conda environment takes precedence over the python modules loaded through `modulefiles/RDAS/machine.intel.lua`. Running `module purge` normally isn't enough to resolve the environment, so adding these commands to the top of `build.sh` sounds like a good idea.

PR #429, opened 2025-08-15
Remove virtual or conda environments before building

SamuelDegelia-NOAA:
## Description

As found by @keenaneure in #427, building RDASApp can fail if an Anaconda or Python virtual environment is currently active. This happens because environment variables and `PATH` entries from these environments can cause the build to link against incompatible libraries. The build might progress for a while just fine but then fail at a pretty late stage. 

Running module purge alone is not always sufficient to clear out these environments. Here, we add a block to explicitly unset these variables and strip any Conda/Miniforge-related paths from `PATH` and `LD_LIBRARY_PATH` before the build starts, ensuring a clean and consistent build environment.

Note: I originally tried using `deactivate` and `conda deactivate` as discussed in #427, but these commands often fail in non-interactive build shells because the shell functions they depend on are not loaded (e.g., `conda init` or `source conda.sh`). The new approach avoids that problem by directly unsetting the relevant variables and cleaning the environment, which works regardless of whether the deactivate commands are available.

I tested this by loading @keenaneure's Miniforge3 environment and then building RDASApp. Everything ran successfully and all executables were built. 

## Issue(s) addressed
#427 

None

## Checklist

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,6 +5,18 @@
 # 2 - configure; build; install
 # 4 - optional, run unit tests
 
+# Deactivate virtual env or conda env to prevent build issues
+if [[ -n "$VIRTUAL_ENV" || -n "$CONDA_PREFIX" ]]; then
+  unset VIRTUAL_ENV
+  unset CONDA_PREFIX
+  unset CONDA_DEFAULT_ENV
+  unset CONDA_SHLVL
+  export PATH=$(echo "$PATH" | tr ':' '\n' | grep -vi 'conda' | grep -vi 'miniforge' | paste -sd ':' -)
+  if [[ -n "$LD_LIBRARY_PATH" ]]; then
+    export LD_LIBRARY_PATH=$(echo "$LD_LIBRARY_PATH" | tr ':' '\n' | grep -vi 'conda' | grep -vi 'miniforge' | paste -sd ':' -)
+  fi
+fi
+
 module purge
 set -eu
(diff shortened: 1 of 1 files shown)


guoqing-noaa (approved):

guoqing-noaa:
@SamuelDegelia-NOAA This is a nice solution!

For myself, I usually start a new clean shell, :)

guoqing-noaa:
Will this solve a corner case where one adds the Python location to the `PATH` variable manually?

ShunLiu-NOAA (approved):

SamuelDegelia-NOAA:
> Will this solve a corner case where one adds the Python location to the `PATH` variable manually?

@guoqing-noaa This change only unsets Conda/virtualenv environments. It does not handle the case where someone has manually added a custom Python installation to their `$PATH`. Handling that automatically would require hard-coding what we consider a "valid" Python path, which could be brittle.

guoqing-noaa:
> > Will this solve a corner case where one adds the Python location to the `PATH` variable manually?
> 
> @guoqing-noaa This change only unsets Conda/virtualenv environments. It does not handle the case where someone has manually added a custom Python installation to their `$PATH`. Handling that automatically would require hard-coding what we consider a "valid" Python path, which could be brittle.

Got it. Totally agree. In this situation, the best way is to launch a new clean terminal with minimum settings/modules loaded.

Issue #388, opened 2025-05-14
METAR Wind Observation Fitting Ratio Degradation with hofx_scaling


spanNOAA:
Testing adpsfc 287uv observations with different parameter combinations produced these results:
#hofx_scaling False, SetSfcWndObsHeight True
Ratio for adpsfc 287uv: 0.2418249900813161

#hofx_scaling False, SetSfcWndObsHeight False
Ratio for adpsfc 287uv: 0.07746080525048159

#hofx_scaling True, SetSfcWndObsHeight False
Ratio for adpsfc 287uv: 0.01363607281666086

#hofx_scaling True, SetSfcWndObsHeight True
Ratio for adpsfc 287uv: -0.22328631722267617
Issues

Best configuration achieves ~24% fit quality
Disabling SetSfcWndObsHeight reduces ratio from 24% to 8%
Enabling hofx_scaling further degrades performance to near-zero
When both hofx_scaling=True and SetSfcWndObsHeight=True, ratio becomes negative (-22%)

@delippi I recall that wind observations didn't initially include the hofx_scaling option. I'm curious about the reason for enabling this feature later.

HuiLiu-NOAA:
Can you provide the actual yaml and testing case for these options? I would like to take a look of this. 
Thanks.

delippi:
@spanNOAA, thanks for doing these tests. What is the ratio metric you're referring to?

It looks like the hofx_scaling option was initially turned on with PR #218. It is also turned on for mesonet (288) winds. As for why it is on, it likely matched the result from GSI better.

spanNOAA:
> [@spanNOAA](https://github.com/spanNOAA), thanks for doing these tests. What is the ratio metric you're referring to?
> 
> It looks like the hofx_scaling option was initially turned on with PR [#218](https://github.com/NOAA-EMC/RDASApp/pull/218). It is also turned on for mesonet (288) winds. As for why it is on, it likely matched the result from GSI better.

It refers to the observation fitting ratio.

HuiLiu-NOAA:
> > [@spanNOAA](https://github.com/spanNOAA), thanks for doing these tests. What is the ratio metric you're referring to?
> > It looks like the hofx_scaling option was initially turned on with PR [#218](https://github.com/NOAA-EMC/RDASApp/pull/218). It is also turned on for mesonet (288) winds. As for why it is on, it likely matched the result from GSI better.
> 
> It refers to the observation fitting ratio.

[...]

Issue #430, opened 2025-08-16
Re-visiting use of the 10m wind scaling in JEDI and GSI


HuiLiu-NOAA:
A review of GSI code indicates that the 10m wind scaling factor (wind speed reduction due to surface friction) is calculated based on the surface layer physics of either GFS or MM5, and its use in GSI is turned off now. 

  In case of this factor is not available from model background, JEDI uses a optional calculation of this factor for both FV3 and MPAS, i.e., a ratio of 

**_RMS of 10m wind speed / RMS of wind speed at lowest model level_**. 

  It seems that the use of this scaling factor for surface wind obs types in MPAS-JEDI needs further testing in terms of its impact on analysis. See a related issue: https://github.com/NOAA-EMC/RDASApp/issues/388

HuiLiu-NOAA:
The calculation of 10m wind scaling factor in JEDI is similar in both fv3-JEDI and mpas-jedi,  i.e., by a ratio of RMS of 10m wind speed / RMS of wind speed at lowest model level.

Given the fact that the fitting ratio is very close with and w/o the scaling factor applied in the FV3-jedi case, it suggests that turning off the scaling factor seems OK. This seems also the case in GSI right now.

HuiLiu-NOAA:
In Summary:
---------------------------------------------------------------------------------
In JEDI, 10m wind reduction factor is implemented as:
---------------------------------------------------------------------------------

 **_factw = RMS of 10m wind speed / RMS of wind speed at lowest level_**


----------------------
Related  GSI codes
----------------------
GSI/gsimod.F90:
! ifact10 - flag for recomputing 10m wind factor
          = 1 re-compute using GFS surface physics
          = 2 re-compute using MM5 surface physics
          = 0 use value from first-guess

setupw.f90:
           factw = data(iff10,i)     ! use the scaling factor from guess
           ugesin = factw * ugesin
           vgesin = factw * vgesin


--------------------------------------------------------------------
It looks like the 10m wind factor is turned off right now in GSI.
--------------------------------------------------------------------
setuprhsall.f90:
These lines are commented out!!
!    if (sfcmod_gfs .or. sfcmod_mm5) then
!       if (mype==0) write(6,*)'COMPUTE_DERIVED:  call load_fact10'
!       call load_fact10
!    endif

guoqing-noaa:
@HuiLiu-NOAA 
For regional models, it looks like: 
[...]

PR #432, opened 2025-08-18
[Bugfix] Add explicit rejection of non-bufrtypes for given obs space & remove hofx_scaling in all yamls

delippi:
## Bug Fix Description

1. First bug: The obs space yamls are itemized meaning there are yamls like adpsfc_airTemperature_181.yaml, adpsfc_airTemperature_183.yaml, adpsfc_airTemperature_187.yaml, etc. The ioda_adpsfc.nc file contains each of these types. We need to add an explicit filter to remove non 181 types in the 181 yaml, etc. I don't believe this currently causes any harm since there is no ObsError in the IODA files. If we eventually included the ObsError in the IODA, the 181 yaml would assimilate 181, 183, and 187 performing QC on only 181. This would lead to triple assimilation of the same obs (twice without QC and once with QC).
2. Second bug: hofx_scaling is not used in RRFSv1 since we reset the 10m wind factor to 1 (`factw=one`). Testing from others has also shown that using hofx_scaling (at least in MPAS-JEDI) had caused poor fitting ratio scores.

1. Now the obs space yamls reject the non-bufrtypes. For example, the 181 yaml explicitly rejects obs that `is_not_in: 181`.
2. No hofx_scaling is performed following the practices of RRFSv1 and fitting ratios are better without.

## Issue(s) addressed
Resolves Issue #388 

N/A

## Checklist


ShunLiu-NOAA (approved):

SamuelDegelia-NOAA (approved):
Looks good to me!

keenaneure (approved):

spanNOAA (approved):

HuiLiu-NOAA:
Can you show the exact code lines about "hofx_scaling is not used in RRFSv1 since we reset the 10m wind factor to 1 (factw=one)". 
  It looks like in setupw.f90, factw=one is set only for the obs's height above model lowest level height.  See Guoqing's @guoqing-noaa comments in https://github.com/NOAA-EMC/RDASApp/issues/430

guoqing-noaa:
> Can you show the exact code lines about "hofx_scaling is not used in RRFSv1 since we reset the 10m wind factor to 1 (factw=one)". It looks like in setupw.f90, factw=one is set only for the obs's height above model lowest level height. See Guoqing's @guoqing-noaa comments in #430

@HuiLiu-NOAA I agree it is a good choice to remove hofx_scaling, especially for RRFSv2 since we intentionally set the first model level at 9.5 meters, i.e close to 10m as much as possible (see [this original documentation](https://docs.google.com/spreadsheets/d/19bQUmS1fgdXzJ0FBzzIh7P7CUEbYSd6OeoPwKkHnseA/edit?gid=258992312#gid=258992312) when we designed the RRFSv2 vertical levels).
I was trying to understand what GSI does in `setupw.f90` in your issue #430

delippi:
@HuiLiu-NOAA, to check I had added print statements to GSI for both possible paths in the `if` statement. All prints were "branch1" like below:
```
branch1: zob, zges(1), factw   10.0000000000000        9.51069356402639         1.00000000000000
```
There was no case where if followed the `else` part of that block for "branch2".

So it is `factw=one` line doing the reset.
```
        if (zob > zges(1)) then
           factw=one
           write(6,*) 'branch1: zob, zges(1), factw',zob,zges(1),factw
        else
           factw = data(iff10,i)
           write(6,*) 'branch2: zob, zges(1), factw',zob,zges(1),factw
```

PR #444, opened 2025-09-14
add `empty obs space action: "skip output"` for all observers

guoqing-noaa:
## Description
PR https://github.com/JCSDA-internal/ioda/pull/1446 added the capability to skip writing out jdiag files for empty obs space.
Writing out jdiag files for empty obs space causes confusions for lots of end users. NSSL users also shared a similar concern.
This PR adds `empty obs space action: "skip output"` to the `obsdataout` section for all observers.

## Issue(s) addressed
None

None

### DIFF
--- a/rrfs-test/validated_yamls/gen_yaml_ctest.sh
+++ b/rrfs-test/validated_yamls/gen_yaml_ctest.sh
@@ -108,6 +108,7 @@ for basic_config in "${!basic_configs[@]}"; do
     python commentQC.py ${ctest_yaml}
 
     # Move to testinput and remove the old temporary yaml
+    sed -i -e "s/@emptyObsSpaceAction@/create output/"  ./jedi.yaml
     ctest_yaml=${basic_configs[$basic_config]}
     echo "Super YAML created in ../testinput/${ctest_yaml}"
     mv ./jedi.yaml ../testinput/$ctest_yaml
--- a/rrfs-test/validated_yamls/gen_yaml_ctest.sh
+++ b/rrfs-test/validated_yamls/gen_yaml_ctest.sh
@@ -68,7 +68,7 @@ for basic_config in "${!basic_configs[@]}"; do
         cp  "./templates/obtype_config/$config" ./replace.yaml
         if [[ $basic_config == *"solver"* ]]; then
             # New obs filename
-            previous_path=`sed -n '/obsdataout/{n; n; n; s/^[[:space:]]\+//; p;}' ./templates/obtype_config/$config`
+            previous_path=`sed -n '/obsdataout/{n; n; n; n; s/^[[:space:]]\+//; p;}' ./templates/obtype_config/$config`
             int_path=$(echo "$previous_path" | sed "s/obsfile: /..\/rundir-${ctest_yaml::-5}\//gI")
             new_path=$(echo "$int_path" | sed "s/solver/observer/gI")
             obs_filename_new="obsfile: ${new_path}"
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
@@ -16,7 +16,7 @@
              obsfile: "data/obs/ioda_abi_g16.nc"
              missing file action: "warn"
          obsdataout:
-           empty obs space action: "skip output"
+           empty obs space action: "@emptyObsSpaceAction@"
            engine:
              type: H5File
              obsfile: jdiag_abi_g16.nc4
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g16.yaml
@@ -16,6 +16,7 @@
              obsfile: "data/obs/ioda_abi_g16.nc"
              missing file action: "warn"
          obsdataout:
+           empty obs space action: "skip output"
            engine:
              type: H5File
              obsfile: jdiag_abi_g16.nc4
--- a/rrfs-test/validated_yamls/templates/obtype_config/abi_g18.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/abi_g18.yaml
@@ -16,7 +16,7 @@
              obsfile: "data/obs/ioda_abi_g18.nc"
              missing file action: "warn"
          obsdataout:
-           empty obs space action: "skip output"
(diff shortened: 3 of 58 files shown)


guoqing-noaa:
Hercules failure is due to time limit reached.

delippi:
I don't think skipping empty obs space output is a good idea. In fact, I see empty files as less confusing than missing ones. It's straightforward to open a `jdiag` file and confirm there are zero obs inside. In contrast, when no file is written at all, it becomes much harder to trace what happened. Was it filtered out, did the job skip a step, or did something fail silently?

TingLei-NOAA:
@guoqing-noaa Would you please explain more on the "empty obs" ?  Does that mean there are not obs (from the input for jedi analysis app) at all?  Or there are not obs passing the QC ?

guoqing-noaa:
> @guoqing-noaa Would you please explain more on the "empty obs" ? Does that mean there are not obs (from the input for jedi analysis app) at all? Or there are not obs passing the QC ?

`empty obs` usually means no input ioda file. For example, at 03z, we usually don't have radiosonde data.
But it is also possible in rare test cases where if one aggressively removes  obs using QC filters and then not one observation will be left.

guoqing-noaa:
> I don't think skipping empty obs space output is a good idea. In fact, I see empty files as less confusing than missing ones. It's straightforward to open a `jdiag` file and confirm there are zero obs inside. In contrast, when no file is written at all, it becomes much harder to trace what happened. Was it filtered out, did the job skip a step, or did something fail silently?

There are no differences between checking an zero obs diag file and no diag file for empty obs space. Only the former one adds one unnecessary step and causes confusions.

SamuelDegelia-NOAA:
I’d prefer to keep the empty `jdiag` files. For me, it’s much clearer to see which observation types were processed each cycle. If a file is missing entirely, we then have to add extra checks: was the IODA file empty? Did the preprocessing or QC step filter everything out? Or did something fail silently? With an empty file, the answer is clear right off the bat. Another reason to keep the empty `jdiag files` is that it makes downstream scripts and monitoring simpler, since they don't need special logic for sometimes-missing files. 

Isn’t this also how it was done in GSI, where diag files were always written even if they contained zero obs? Just curious where the confusion is coming from. If the concern is confusion from things like adpupa jdiag files at non-00/12Z cycles, maybe that’s better solved with clarification or documentation, rather than skipping the file creation altogether.

guoqing-noaa:
@SamuelDegelia-NOAA GSI will not write out any diag files if the observation files are missing

guoqing-noaa:
@SamuelDegelia-NOAA I don't understand what the differences between checking a zero-obs jdiag file and checking an non-exsiting jdiag file?  Anything you listed are the same for both situations except that writing out an empty file takes an extra unnecessary step. I don't think it will bring any troubles. If you have an good example to demonstrate this, I would love to check it. Thanks!

guoqing-noaa:
I would say this is NOT a scientific-critical problem. We can try this and if we do get any troubles, we can always revert the changes easily.

ShunLiu-NOAA:
@guoqing-noaa, @SamuelDegelia-NOAA and @delippi, can we include this as a configurable option to yaml file? Cab the user specify if he will use this option? Can JCB handle this option?

guoqing-noaa:
@ShunLiu-NOAA Thanks for the good suggestion. Yes, we can make it configurable!
I will modify as follows:
```
empty obs space action: "@emptyObsSpaceAction@"
```
Then it will be replaced at the generation step by replacing `@emptyObsSpaceAction@` with either `skip output` or `create output`

SamuelDegelia-NOAA:
For me it feels cleaner to always have an empty `jdiag` file. It’s deterministic that way in that for each obs space, we always get one output. This makes auditing trivial in that it’s much easier to see an empty file than to mentally scan the directory and notice that one is missing.

I also suspect some of the monitoring tools in `RDASApp/rrfs-test/ush/diagnostics_and_graphics` rely on `jdiag` files being present each cycle (though I’m not 100% sure). On my end, I have code that counts assimilated obs per cycle, and it’s simpler if it can just loop through files. Sure, I could add conditional logic to handle missing files, but then if a file doesn’t exist and my code reports “0,” I still don’t know whether the obs type was processed and filtered out, or if something went wrong upstream.

I’m still not sure why an empty `jdiag` would be confusing, but I’m fine with the configurable option. Thanks for making that change!

guoqing-noaa:
@SamuelDegelia-NOAA Thanks for the discussions! I think it may be more helpful if I can check a concrete case. 
In the situation you mentioned, I still don't see any meaningful differences between reading a non-empty but zero obs diag file and checking if a diag file exists or not. 

For any monitoring, I think it is critical to handle non-existing file situations since in the computer world, anything can happen. Even a non-empty but zero obs diag file may be generated in the JEDI DA step, we may loose it in any subsequent steps. It is always great to make the monitoring process robust especially if running it automatically.

I would share what benefits I see for skipping writing out any empty obs space. I took a few User Interface (or HCI) courses and I agree with one of the basic principles that if a user complains or gets confused at the first try, then it means that user interface can be improved if possible. 

When I first ran radiosonde data assimilation, I was surprised to see that I got radiosonde diag file at 03z, 04z, etc where I don't have any radiosonde ioda input files!! Although I later figured out that those diag files contain zero obs after "ncdump" or checking using a python code, I had thought I might do something wrong initially (to make thing complicated, we do have very limited radiosonde obs at 01z, 07z, etc). Yunheng from NSSL and a few other users got similar situation as well. 

So this is not an intuitive way to do things. If possible, we would expect to do things straightly, ie., if we don't have observations for a cycle, we would intuitively NOT expect any diag files for those observations. If we expect a diag file from a given obs type but we don't see that, that also intuitively indicates something is wrong in the DA process. All these can be done without doing an extra step (such as running ncdump or a python script). 

Back to the logic to check non-existing diag files, I think that would be easily handled in any monitor scripts, not much different from checking whether we have zero observations. Thanks!

SamuelDegelia-NOAA:
@guoqing-noaa Thanks for laying out your perspective in detail. I think this comes down to different priorities: for me, the determinism of always having a `jdiag` file (even if empty) makes auditing and downstream automation cleaner. Maybe that is just what I am more used to at his point. For you and others, not seeing a file at all feels more intuitive.

I’m glad we now have a configurable option, that seems like the best option. Thanks again for making the change and for the thoughtful discussion.

guoqing-noaa:
@SamuelDegelia-NOAA  Yeah, thanks for the discussions!

SamuelDegelia-NOAA (approved):

Issue #447, opened 2025-09-19
Propose adding "filter name" to each filter in each ob space yaml


delippi:
## Description
I am proposing to add an unused (by JEDI) key `filter name` to every filter in each ob space yaml and will match the comment that is already in place for each block. The reason for this is two fold:

1. When moving to JCB, all comments will be stripped from the yamls. Having this `filter name`, while potentially unused otherwise (JCB doesn't know it is unused), will keep some level of documentation about each filter and add ease of navigating through the yaml.
2. Another reason for this is that it potentially makes parsing and mass edits easier. (Hopefully with JCB this part will be easier or even unneeded with the observation chronicle--though the chronicle doesn't yet support anything for conventional data). The idea is that we can have any number of `Bounds Check`, for example, in each yaml but no way to really distinguish which Bounds Check when parsing or mass editing.

I've already tested such functionality with the following example in the ctest case and it ran without any issue:

```
         # Online regional domain check
         - filter: Bounds Check
           filter name: "Online regional domain check"
           filter variables:
           - name: airTemperature
           test variables:
           - name:  GeoVaLs/observable_domain_mask
           minvalue: 0.0
           maxvalue: 0.5
```

I wanted to get feedback and support for this change before I edit the 50+ files and every filter within those files.

delippi:
@guoqing-noaa @SamuelDegelia-NOAA @SamuelDegelia-NOAA I wanted to get feedback and support for this change before I edit the 50+ files and every filter within those files. I think this could be really useful, but there could be things I'm overlooking.

Thanks

guoqing-noaa:
@delippi This is a good idea! 

I have some minor comments:
1. It may be better to use `uname` (unique name) instead of `filter name`.  No matter what, each filter already has a "name",  i.e. "filter: Bounds Check". Using `uname` may help reduce any confusions.
2. I suggest we use either camelCase or concatenation_using_underscore for the unique name, i.e. `onlineRegionalDomainCheck` or `online_regional_domain_check`. We may update JEDI codes in the future to use this extra information, for example, we can include them in the jdiag files. camelCase or concatenation_using_underscore is more friendly for downstream processing.
3. I suggest we make unique names for each filter, not just copy comments (although in lots of situations comments are already good options for unique names).
Thanks!

delippi:
[...]

Issue #254, opened 2024-12-17
Add JCB to RDASApp


delippi:
## Description
While our current method of templating yamls to generate very large jedi.yamls is simple and effective, we still would like to migrate to using JCB. For a couple of reasons:

1. JCB is more powerful in configuring the jedi yamls especially with the observation_chronicle function.
2. GDASApp already uses JCB and we should unify methods and tools between systems as much as possible. This also means that global and regional teams can work together more effectively vs having two completely different systems.

## Requirements

- https://github.com/NOAA-EMC/jcb must be a submodule under sorc/
- https://github.com/NOAA-EMC/jcb-algorithms is a submodule under parm/
- https://github.com/NOAA-EMC/jcb-rdas needs created and to be a submodule under parm/

## Acceptance Criteria (Definition of Done)

- All new submodules added to RDASApp
- RDAS added to client list in other JCB repos
- Provide a simple example and wiki instructions
- Can build a useable jedi.yaml

## Dependencies
None

PR #448, opened 2025-09-19
Rplace jcb-rdas submodule with regular directory

delippi:
## Description
The purpose of this PR is mainly to replace the jcb-rdas submodule with a regular directory instead of having to continuously update the jcb-rdas submodule hash. GDASApp plans to also go this approach.

Other changes include:

- Updated each obs space filter with a udescription (unique descriptor) to describe each filter. This is two fold 1) JCB removes comments and this will help keep that information if inspecting a post-JCB yaml 2) it makes mass editing a bit easier.
- Mass edited convinfo settings that were mis configured due to using an incorrect convinfo.rrfs (gross, ermin, ermax)
- Added tools to quickly compare the convinfo.rrfs with each yaml. Also added the correct convinfo.rrfs and errtable.rrfs currently being used by rrfsv1 for future reference.

## Issue(s) addressed
https://github.com/NOAA-EMC/RDASApp/issues/254
https://github.com/NOAA-EMC/RDASApp/issues/447

None


delippi:
Sorry for another huge PR. Mainly moving the files I just put in jcb-rdas with a few mods.

ShunLiu-NOAA (dismissed):
Looks good to me.

parm/jcb-rdas/observations/atmosphere/adpsfc_airTemperature_181.yaml.j2
```diff
@@ -0,0 +1,320 @@
...
+         - localization method: Horizontal Gaspari-Cohn
+           lengthscale: 200e3 # orig
+
+       obs filters:
+         # ------------------
+         # airTemperature (181)
+         # ------------------
+         # Accept, reject, or passivate (monitor)
+         - filter: Perform Action
+           udescription: "iuse_check"
```
guoqing-noaa: I think the word `descriptor` in your PR's description is better than the word `udescription`.

SamuelDegelia-NOAA:
Does running `python jcb_client_init.py` still clone the original `jcb-rdas` repo even after the submodule is removed?

delippi:
> Does running `python jcb_client_init.py` still clone the original `jcb-rdas` repo even after the submodule is removed?

Yes, it should since we didn't delete the actual jcb-rdas repo.

SamuelDegelia-NOAA:
@delippi Thanks, just checking. That means we could have multiple different versions of these templates unless we push updates to both RDASApp and `jcb-rdas`. I think that is fine for now but just something we should be aware of while we are still figuring out to handle the separate `jcb-rdas` repo.

SamuelDegelia-NOAA (approved):

delippi:
> @delippi Thanks, just checking. That means we could have multiple different versions of these templates unless we push updates to both RDASApp and `jcb-rdas`. I think that is fine for now but just something we should be aware of while we are still figuring out to handle the separate `jcb-rdas` repo.

@SamuelDegelia-NOAA forgot to reply. The templates in jcb-rdas will eventually by synced. I'm not sure how it is planned to do that yet or even in jcb-rdas/gdas will even remain as a repo. For now, those templates are working for the testing. The actual ones we'll want to use will always be under parm/jcb-rdas. Yes, it'll be confusing to have potentially different templates floating around, but that won't be forever.

ShunLiu-NOAA (approved):

PR #467, opened 2025-10-09
Update offline domain checks for better efficiency when run on larger domains

SamuelDegelia-NOAA:
## Description

This PR introduces new versions of the offline domain check (`offline_domain_check.py` and `offline_domain_check_satrad.py`). When testing these codes on the 3-km NA domain, it was found that they became extremely slow and could not finish within a reasonable amount of time. For example, running the domain check on `ioda_msonet.nc` could take nearly an hour. 

This update replaces the old alpha-shape triangulation (which scaled poorly with large MPAS or FV3 grids) with a fast method that generates the true model boundary directly. By avoiding Delaunay triangulation over millions of points, the runtime drops from minutes or hour to a few seconds while producing an equally accurate domain mask. 

Through testing, I found that using the domain check on either the FV3 or MPAS domains yields very similar results to the original version. On the conus 12-km domain, the final obs counts are equivalent between versions. On the 3-km NA domain, the final obs counts are within just a few obs of the original version. 

Here is the output and timing from the original domain check run on the NA 3-km domain. These tests use `ioda_aircar.nc` for simplicity:  

```
[Samuel.Degelia@hfe10 IODA]$ python offline_domain_check_orig.py -g fv3_grid_spec_conus3km -o ioda_aircar.nc -f -s 0.005
Obs file: ioda_aircar.nc
Grid file: fv3_grid_spec_conus3km
Hull shrink factor: 0.005
Ob counts:
  Excluded: 750
  Included: 81102
  Total:    81852
Time to find obs within domain: (751.41s), 00:12:31
Time to create new obs file: (0.34s), 00:00:00
```

And here is the output and timing from the new version: 

```
[Samuel.Degelia@hfe10 IODA]$ python offline_domain_check.py -g fv3_grid_spec_conus3km -o ioda_aircar.nc -f -s 0.005
Obs file: ioda_aircar.nc
Grid file: fv3_grid_spec_conus3km
Hull shrink factor: 0.005
Ob counts:
  Excluded: 748
  Included: 81104
  Total:    81852
Time to find obs within domain: (7.66s), 00:00:07
Time to create new obs file: (0.45s), 00:00:00
```

The new domain check takes only 7 seconds to generate the domain mask whereas the previous version took 751 seconds. 

## Issue(s) addressed
None

None

## Checklist


ShunLiu-NOAA (approved):

HuiLiu-NOAA:
Great update! Thanks.

ShunLiu-NOAA:
Thanks @SamuelDegelia-NOAA. I got a copy of your updated offline domain check from Donnie and tried it with rrfs-workflow. It works wonderfully and greatly improves efficiency.

SamuelDegelia-NOAA:
@ShunLiu-NOAA Thank you for testing it. I should note that the version I sent you and Donnie yesterday had some issues when applied to the MPAS domain. Those problems are resolved in this PR. But since you are testing the FV3 version in rrfs-workflow, it should have no problem.

xyzemc:
That is great. Thanks!

PR #502, opened 2025-12-08
Enable "surface observation error ramp: true" for all conv t/q obs near surface

delippi:
## Description
This PR enables `surface observation error ramp: true` for all temperature and humidity conventional obs in the obs YAMLs and updates the ctest refs for the expected failed tests.
 
For RRFSv1 the following gsiparm namelist settings are used:
- `l_sfcobserror_ramp_t=.true.,`
- `l_sfcobserror_ramp_q=.true.,`

In GSI's setup t/q routines, there is no kx type matching for `l_sfcobserror_ramp_t/q` option and therefore applies to all types near surface.

```
     if (.not. (hofx_2m_sfcfile  .and. landsfctype) ) then
         rlow=max(sfcchk-dpres,zero)
!        linear variation of observation ramp [between grid points 1(~3mb) and 15(~45mb) below the surface]
         if(l_sfcobserror_ramp_t) then
            ramp=min(max(((rlow-1.0_r_kind)/(15.0_r_kind-1.0_r_kind)),0.0_r_kind),1.0_r_kind)
         else
            ramp=rlow
         endif  
     else
        rlow  = zero
        ramp  = zero
     endif   
```

## Checklist


delippi:
Here are some validation plots confirming this result for aircar (as an example):

This first plot is with `surface observation error ramp: false` (the default setting)
control

This second plot is with `surface observation error ramp: true`. While still not perfect (I don't think we can expect perfect match), JEDI's ob errors more closely align with that of GSI. Please notice the scale of the axis differences between this and the previous plot for the delta obs error vs pressure plot and the JEDI/GSI obs error comparison plot.
hofx_compare_2024050601_aircar_airTemperature_133_v1 2 0_det_c13_3dvar_vad_8panel

SamuelDegelia-NOAA (approved):
Nice improvement in the obs error compared to GSI. Thanks for the quick fix!

ShunLiu-NOAA (approved):

PR #505, opened 2025-12-09
Addition of 2 observation-space YAMLs for GOES AMVs ob type 247 (deep layer water vapor)

keenaneure:
This PR contains 2 observation-space YAMLs for GOES AMV satwinds. Type 247: deep layer water vapor. There is one YAML for each of the GOES-16 (satID: 270) and GOES-18 (satID: 272) satellites, and 2 total in this PR.

These have been successfully compared against equivalent GSI tests with the same observations.


delippi:
I am currently reviewing by running through our dev-sci retro and looking at ob counts, hofx, and obs error with the new validation tools. There was some concern that I wanted to just double check.

keenaneure:
No problem. Thanks for the update.

HuiLiu-NOAA:
A quick scan of the yamls seem fine to me. 
The next lines may be not necessary, though:

       linear obs operator:
         name: VertInterp

keenaneure:
> A quick scan of the yamls seem fine to me. The next lines may be not necessary, though:
> 
> ```
>    linear obs operator:
>      name: VertInterp
> ```

Perhaps this can be removed. Should all 3 ob types have this removed? Currently, the YAMLs for 245 and 246 have this included.

HuiLiu-NOAA:
> > A quick scan of the yamls seem fine to me. The next lines may be not necessary, though:
> > ```
> >    linear obs operator:
> >      name: VertInterp
> > ```
> 
> Perhaps this can be removed. Should all 3 ob types have this removed? Currently, the YAMLs for 245 and 246 have this included.

I would agree with this removal. The yamls seem more clean.

delippi:
> I am currently reviewing by running through our dev-sci retro and looking at ob counts, hofx, and obs error with the new validation tools. There was some concern that I wanted to just double check.

I ran these (and the 245 and 246 satwnds) through 01Z cycle in our simplified retro (after changing to .j2 templates) so that I could check the results with the new validation tool. Honestly, the use of these obs looks pretty nice with the exception of some number of observation counts for 245 and 246 (you can find those counts in the lower left panel text annotations; sorry for the small text).

hofx_compare_2024050601_satwnd_windEastward_245_270_v1 2 0_det_c13_3dvar_satwnd_8panel
hofx_compare_2024050601_satwnd_windEastward_245_272_v1 2 0_det_c13_3dvar_satwnd_8panel
hofx_compare_2024050601_satwnd_windEastward_246_270_v1 2 0_det_c13_3dvar_satwnd_8panel
hofx_compare_2024050601_satwnd_windEastward_246_272_v1 2 0_det_c13_3dvar_satwnd_8panel
hofx_compare_2024050601_satwnd_windEastward_247_270_v1 2 0_det_c13_3dvar_satwnd_8panel
hofx_compare_2024050601_satwnd_windEastward_247_272_v1 2 0_det_c13_3dvar_satwnd_8panel

delippi:
> > > A quick scan of the yamls seem fine to me. The next lines may be not necessary, though:
> > > ```
> > >    linear obs operator:
> > >      name: VertInterp
> > > ```
> > 
> > 
> > Perhaps this can be removed. Should all 3 ob types have this removed? Currently, the YAMLs for 245 and 246 have this included.
> 
> I would agree with this removal. The yamls seem more clean.

Technically, 3DVar still uses the linear obs operator, though I tested in our ctest and without that block 
1. It does still seem to work.
2. It seemed to run faster. (needs more testing)

It must fall back on the Nonlinear observation operator name? And even is we were using a truly nonlinear obs operator, why would we even need to set this assuming the fallback properly maps to what is needed?

Final thought: I prefer to explicitly define our options so that we know exactly what we are using (similar to my position for always defining the obs error in the YAMLs even if they might be in the IODA file), but in this case, I've mixed feelings if it truly does run faster.

HuiLiu-NOAA:
> > > > A quick scan of the yamls seem fine to me. The next lines may be not necessary, though:
> > > > ```
> > > >    linear obs operator:
> > > >      name: VertInterp
> > > > ```
> > > 
> > > 
> > > Perhaps this can be removed. Should all 3 ob types have this removed? Currently, the YAMLs for 245 and 246 have this included.
> > 
> > 
> > I would agree with this removal. The yamls seem more clean.
> 
> Technically, 3DVar still uses the linear obs operator, though I tested in our ctest and without that block
> 
>     1. It does still seem to work.
> 
>     2. It seemed to run faster. (needs more testing)
> 
> 
> It must fall back on the Nonlinear observation operator name? And even is we were using a truly nonlinear obs operator, why would we even need to set this assuming the fallback properly maps to what is needed?
> 
> Final thought: I prefer to explicitly define our options so that we know exactly what we are using (similar to my position for always defining the obs error in the YAMLs even if they might be in the IODA file), but in this case, I've mixed feelings if it truly does run faster.

In default, the linear operators (TL/AD) will use the same parameters defined in the nonlinear operator space, as the case "VertInterp" operator.  If any linear operators use extra parameters than the nonlinear ones, then it will be needed to turn on this section to read in these parameters, which might be slower.  So, the purpose of this section is to read in extra parameters for the linear operators.

  In short, for the "VertInterp" operator, no need to read in additional parameters. So, skipping this section would make the yaml more clear (and clean/neat and even faster).

HuiLiu-NOAA:
> > I am currently reviewing by running through our dev-sci retro and looking at ob counts, hofx, and obs error with the new validation tools. There was some concern that I wanted to just double check.
> 
> I ran these (and the 245 and 246 satwnds) through 01Z cycle in our simplified retro (after changing to .j2 templates) so that I could check the results with the new validation tool. Honestly, the use of these obs looks pretty nice with the exception of some number of observation counts for 245 and 246 (you can find those counts in the lower left panel text annotations; sorry for the small text).
> hofx_compare_2024050601_satwnd_windEastward_245_270_v1 2 0_det_c13_3dvar_satwnd_8panel hofx_compare_2024050601_satwnd_windEastward_245_272_v1 2 0_det_c13_3dvar_satwnd_8panel hofx_compare_2024050601_satwnd_windEastward_246_270_v1 2 0_det_c13_3dvar_satwnd_8panel hofx_compare_2024050601_satwnd_windEastward_246_272_v1 2 0_det_c13_3dvar_satwnd_8panel hofx_compare_2024050601_satwnd_windEastward_247_270_v1 2 0_det_c13_3dvar_satwnd_8panel hofx_compare_2024050601_satwnd_windEastward_247_272_v1 2 0_det_c13_3dvar_satwnd_8panel

Could this be related to potential issue of the rdas_ua2u tool?

delippi:
@HuiLiu-NOAA no, any differences in these plots wouldn't be from the rdas_ua2u tool. These plots are only showing observation space things from identical backgrounds. In each DA cycle, we run GSI and JEDI from an identical background. The rdas_ua2u tool comes into play when we computes D-grid winds from A-grid winds needed to restart the model. Its irrelevant which we start from in this case because at the next cycle we then just run GSI and JEDI from that same background. This eliminates any differences in previous DA or forecast cycle that may propagate through the forecast and lets us consider only the current DA cycle differences.

Edit: This is also the 01z cycle (the first DA cycle after cold start) which wouldn't have seen differences, if there were any, anyway.

HuiLiu-NOAA:
> > > > A quick scan of the yamls seem fine to me. The next lines may be not necessary, though:
> > > > ```
> > > >    linear obs operator:
> > > >      name: VertInterp
> > > > ```
> > > 
> > > 
> > > Perhaps this can be removed. Should all 3 ob types have this removed? Currently, the YAMLs for 245 and 246 have this included.
> > 
> > 
> > I would agree with this removal. The yamls seem more clean.
> 
> Technically, 3DVar still uses the linear obs operator, though I tested in our ctest and without that block
> 
>     1. It does still seem to work.
> 
>     2. It seemed to run faster. (needs more testing)
> 
> 
> It must fall back on the Nonlinear observation operator name? And even is we were using a truly nonlinear obs operator, why would we even need to set this assuming the fallback properly maps to what is needed?
> 
> Final thought: I prefer to explicitly define our options so that we know exactly what we are using (similar to my position for always defining the obs error in the YAMLs even if they might be in the IODA file), but in this case, I've mixed feelings if it truly does run faster.

"(similar to my position for always defining the obs error in the YAMLs even if they might be in the IODA file),"

Could this re-defining obs error also slow the JEDI DA process vs. not doing so in the yamls?  I guess the slowing could be substantial given the large amount of observations affected.

HuiLiu-NOAA:
@keenaneure: 
 It would be interesting if you could do a simple test: just turn off all obs err definition sections of the yamls and run a few cycles, to get an idea how faster of JEDI DA could be.  Given the large amount of AMV data, I guess, the time saving could be substantial.

delippi:
> > > > > A quick scan of the yamls seem fine to me. The next lines may be not necessary, though:
> > > > > ```
> > > > >    linear obs operator:
> > > > >      name: VertInterp
> > > > > ```
> > > > 
> > > > 
> > > > Perhaps this can be removed. Should all 3 ob types have this removed? Currently, the YAMLs for 245 and 246 have this included.
> > > 
> > > 
> > > I would agree with this removal. The yamls seem more clean.
> > 
> > 
> > Technically, 3DVar still uses the linear obs operator, though I tested in our ctest and without that block
> > ```
> > 1. It does still seem to work.
> > 
> > 2. It seemed to run faster. (needs more testing)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It must fall back on the Nonlinear observation operator name? And even is we were using a truly nonlinear obs operator, why would we even need to set this assuming the fallback properly maps to what is needed?
> > Final thought: I prefer to explicitly define our options so that we know exactly what we are using (similar to my position for always defining the obs error in the YAMLs even if they might be in the IODA file), but in this case, I've mixed feelings if it truly does run faster.
> 
> "(similar to my position for always defining the obs error in the YAMLs even if they might be in the IODA file),"
> 
> Could this re-defining obs error also slow the JEDI DA process vs. not doing so in the yamls? I guess the slowing could be substantial given the large amount of observations affected.

I suppose it could be different. I would suspect possibly the opposite (slower when reading per ob's ObsError) since what we currently do is just apply the same obs function across all obs in the obs space. Even still, its not always about "saving time" per se, but doing things right. The ob errors aren't always defined in the IODA file, and if they are, how do we know if they are what we intended to use? I always advocate for clarity and transparency.

If you're curious about potential speed up, you're more than welcome to to test, but, again, I don't think it would be the right move.

keenaneure:
Could a single-time test tell us about the speed as well?

delippi:
> Could a single-time test tell us about the speed as well?

It should be at least 5-10 different runs of each configuration because run-to-run machine variance is real. That's why I said removing the linear obs operator block needed more testing to confirm it actually did speed up because I only tested with and without (1 experiment each).

I think it is all getting out of the scope of this PR.

delippi (approved):

keenaneure:
Got it. I will add 246 and 247 to the rrfs-workflow super YAML after this is merged.

ShunLiu-NOAA (approved):

Issue #295, opened 2025-02-26
Test N20 CrIS-FSR data in MPAS/JEDI with 12km CONUS case


HaidaoLin-NOAA:
Test CrIS-FSR radiance data from N20, work will include the following steps:

1. convert budr data to ioda format with bufr2ioda tools:
   #234
2. convert cris-fsr bias file to ioda format with satbias2ioda tools:
https://github.com/NOAA-EMC/RDASApp/issues/52
3. creat and test N20 CrIS-FSR yaml file based on these reference yaml files(cris/iasi/atms) from GDASApp/RDASApp:
https://github.com/NOAA-EMC/jcb-gdas/blob/de1fd083ccdf637470f23157b3d694dc331f0ef8/observations/atmosphere/cris-fsr_n20.yaml.j2
https://github.com/NOAA-EMC/jcb-gdas/blob/de1fd083ccdf637470f23157b3d694dc331f0ef8/observations/atmosphere/iasi_metop-a.yaml.j2
https://github.com/NOAA-EMC/RDASApp/blob/develop/rrfs-test/validated_yamls/templates/obtype_config/atms_n20.yaml
4. make sure the yaml file is working properly, the first thing is the conversion form radiance to brightness temperature, and then different filters/QC etc.
5. check the assimilation results
6. ctest

HaidaoLin-NOAA:
Updates on CrIS-FSR radiance to BT conversion:
1) assign channel wavenumber in m-1 through the variable assignment filter (for three different CrIS-FSR bands, each band has a different intercept number)
2) unit conversion: radiance unit conversion from W M**-2 SR**-1 CM to W M**-2 SR**-1 M by multiplying 0.01 through a new group variable and the variable assignment filter, the new group variable (radiance) is used in the filter of SatBrightnessTempFromRad 
3) The calculated BTs from JEDI are verified against the BT values calculated outside of JEDI with accuracy up to three decimal places, this is the [PPT](https://docs.google.com/presentation/d/15_En5ogoLRVfwPJylFvD3nlshrdHXB_J/edit#slide=id.p1) for the comparison

HaidaoLin-NOAA:
https://github.com/NOAA-EMC/GDASApp/issues/1493#event-17159985472

PR #512, opened 2025-12-17
Add bufr2netcdf and bufr2ioda yaml files for cris-fsr data

HaidaoLin-NOAA:
Fixes the issue(s) mentioned in
https://github.com/NOAA-EMC/RDASApp/issues/234
https://github.com/NOAA-EMC/RDASApp/issues/295
Related PRs:
https://github.com/NOAA-EMC/rrfs-workflow/pull/842
https://github.com/NOAA-EMC/rrfs-workflow/pull/1201
https://github.com/NOAA-EMC/rrfs-workflow/pull/1214

PR #513, opened 2025-12-17
Add n20/n21 cris-fsr obsSpace yaml files

HaidaoLin-NOAA:
fix issue: https://github.com/NOAA-EMC/RDASApp/issues/295
related PRs:
https://github.com/NOAA-EMC/rrfs-workflow/pull/1201
https://github.com/NOAA-EMC/rrfs-workflow/pull/842
https://github.com/NOAA-EMC/rrfs-workflow/pull/1214
https://github.com/NOAA-EMC/RDASApp/pull/512

### DIFF
--- a/rrfs-test/validated_yamls/templates/obtype_config/cris-fsr_n20.yaml
+++ b/rrfs-test/validated_yamls/templates/obtype_config/cris-fsr_n20.yaml
@@ -1,0 +1,548 @@
+     - obs space:
+         name: cris-fsr_n20
+         _anchor_channels: &cris-fsr_n20_channels 1
+         _anchor_use_flag: &cris-fsr_n20_use_flag [1]
+         _anchor_use_flag_clddet: &cris-fsr_n20_use_flag_clddet [1]
+         _anchor_error: &cris-fsr_n20_error [1]
+         _anchor_obserr_bound_max: &cris-fsr_n20_obserr_bound_max [1]
+         distribution:
+           name: "RoundRobin"
+           halo size: 100e3
+         obsdatain:
+           engine:
+             type: H5File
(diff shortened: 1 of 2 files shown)


guoqing-noaa (approved):

delippi (approved):
Seems fine to me. 

I appreciate all the comments around each filter block. Just one comment about that: when this gets loaded up in JCB, all of those comments will get removed (as does all the extra line spacing) in the final super yaml, so it becomes quite difficult to read that final yaml if you ever needed to do so. So one thing that I have been doing for the conventional types is adding a fake, unused key (`udescriptor`) in each filter to act as a post-JCB comment which helps find the filter block if you ever need to look through that giant yaml. It may also help to update filter blocks programmatically. Here is an example

https://github.com/NOAA-EMC/RDASApp/blob/5103845fe411a3b025c23330ba29576ee86a78c7/parm/jcb-rdas/observations/atmosphere/adpsfc_airTemperature_181.yaml.j2#L51

At some point, we might consider to add a similar `iuse_check` to the radiance yamls like this example:
https://github.com/NOAA-EMC/RDASApp/blob/5103845fe411a3b025c23330ba29576ee86a78c7/parm/jcb-rdas/observations/atmosphere/adpsfc_airTemperature_181.yaml.j2#L49-L58
This is used in JCB to turn on/off each data type while keeping the obs space in the final yaml (in monitor mode). This isn't urgent because we can just as easily exclude it from the final yaml as well. 

I am not suggesting that we need to add either at this time. This is just for awareness. In fact, I don't think any radiance yamls currently have either of these optional functions.

HaidaoLin-NOAA:
@delippi Thank you for the detailed/useful information!  I will try to learn more about this. Thanks.

Issue #518, opened 2026-01-07
the setup in fv3jedi for pressure analysis (delp and surface pressure)


TingLei-NOAA:
This issue is to facilitate further discussion/investigation of this issue as the title. 

 @delippi has found https://github.com/NOAA-EMC/RDASApp/issues/507#issuecomment-3715467328 that the delp in the analysis is not updated in his rrfs runs.  It is further confirmed this occurs for some standard rrfs fv3-jedi setup as in the rdasapp ctest .
Further tracing this in fv3-jedi, my current understanding is that : not like in gsi, in which, after the analysis , the surface pressure increment would be used to update delp increment (by the hydrostatic relationship),  fv3jedi does a more general ps to delp (and vice versa) in the variables changes in the control2analysis block.   That means we have to use both delp and ps as the analysis variables and explicitly specify ps (no delp) as the active variables in the background error session.
The above setup (using both ps and delp as the analysis variables but only ps as the control variable) will not work for enkf .  If we still want to only use ps (not delp aggressively directly analyzed), we have to add an gsi-similar like ps to delp conversion in fv3-jedi.

We had some discussion before as https://github.com/NOAA-EMC/RDASApp/issues/456, in which I thought fv3-jedi was doing similarly on this as GSI does ,misleadingly.  
Also bring this to @XuLu-NOAA  'a attention. 
An clarification :  "assignment" of the issues by me means the attention is hoped for and doesn't mean any responsibilities. :). 
Update on Jan.12,2026:  the "Fv3-JEDI-compatible " treatment of delp and ps in the fv3jedi hybrid analysis is done by : (1) both of ps and delp could (need) be used as analysis variables; in the static B part, only ps is used as the control variable (go through the B ) and the above mentioned linear relation is used between ps and delp; (2) for ensemble error part, ps and delp are analyzed by the localized ensemble sampled covariance.  Hence, finally, this "F3-jedi compatible" analysis would give such a self-consistent hybrid analysis, in which, the ensemble component of ps increment would be just be discarded after the analysis. This is a major difference from what the current fv3reg GSI hybrid run does

ShunLiu-NOAA:
Following our discussion today, the first step is to compare the Psfc-derived delp increments in GSI against the delp increments in JEDI. We can likely implement a Psfc assimilation strategy in JEDI that mirrors the GSI approach.

TingLei-NOAA:
For discussion convenience,  some relevant  subroutines are listed here:
[...]

PR #520, opened 2026-01-12
Add offline tool to compute delp increments

delippi:
## Description
This PR adds an offline python tool to produce delp increments from increments of ps. It also adds the infrastructure to add delp as an analysis variable to the fv3-jedi ctests, but that is currently turned off since we want to use an offline tool to do this. There are also various other consistency and formatting changes. It was previously found that using ps as control variable and analysis variable did not diagnostically update delp. Delp increments are needed for model restart.

Running the new tool:
```
python offline_compute_delp_inc.py --sfc_inc inc_jedi.sfc_data.nc --core_inc inc_jedi.fv_core.res.nc --akbk fv3_akbk
```

## Issue(s) addressed
- https://github.com/NOAA-EMC/RDASApp/issues/518

## Checklist
- [x] Unit tests added/updated (if applicable).

Acknowledgements:
@TingLei-NOAA @HuiLiu-NOAA @Masanori-NOAA @SamuelDegelia-NOAA @ShunLiu-NOAA This was a whole team effort.


ShunLiu-NOAA (dismissed):

TingLei-NOAA (commented):
So a general quesiton first:  for those pure ens3dvar runs,  in this PR, ps and delp are used simultaneously, right?  In that case, the ps updating wouldn't have any impacts (unless indirectly affect the results by changing of the values of the background terms in the cost function in the optimization step).  In that case, should we just use only delp for those pure ens3dvar?

HuiLiu-NOAA:
> So a general quesiton first: for those pure ens3dvar runs, in this PR, ps and delp are used simultaneously, right? In that case, the ps updating wouldn't have any impacts (unless indirectly affect the results by changing of the values of the background terms in the cost function in the optimization step). In that case, should we just use only delp for those pure ens3dvar?

You mean "ps and delp are used simultaneously" as analysis variables or control variables?

TingLei-NOAA:
> > So a general quesiton first: for those pure ens3dvar runs, in this PR, ps and delp are used simultaneously, right? In that case, the ps updating wouldn't have any impacts (unless indirectly affect the results by changing of the values of the background terms in the cost function in the optimization step). In that case, should we just use only delp for those pure ens3dvar?
> 
> You mean "ps and delp are used simultaneously" as analysis variables or control variables?


@HuiLiu-NOAA , in pure ens3dvar, delp and ps are used both as analysis variables and control variables , simultaneously.  The subtle differences between analysis variables and control variables exist for the static B (with those linear variables change in effect).

HuiLiu-NOAA:
> > > So a general quesiton first: for those pure ens3dvar runs, in this PR, ps and delp are used simultaneously, right? In that case, the ps updating wouldn't have any impacts (unless indirectly affect the results by changing of the values of the background terms in the cost function in the optimization step). In that case, should we just use only delp for those pure ens3dvar?
> > 
> > 
> > You mean "ps and delp are used simultaneously" as analysis variables or control variables?
> 
> @HuiLiu-NOAA , in pure ens3dvar, delp and ps are used both as analysis variables and control variables , simultaneously. The subtle differences between analysis variables and control variables exist for the static B (with those linear variables change in effect).

Thanks. If only "delp" is used in the pure end3dvar, how to update "ps"?

TingLei-NOAA:
" If only "delp" is used in the pure end3dvar, how to update "ps"?"
@HuiLiu-NOAA  the point is that, for pure ens3dvar,  even using both ps and delp as analysis variables, the ps wouldn't be updated in the current fv3-jedi ( ps is absent from the restart files).

HuiLiu-NOAA:
> " If only "delp" is used in the pure end3dvar, how to update "ps"?" @HuiLiu-NOAA the point is that, for pure ens3dvar, even using both ps and delp as analysis variables, the ps wouldn't be updated in the current fv3-jedi ( ps is absent from the restart files).

OK. then, the "delp" only option should work.

delippi:
@TingLei-NOAA as you say, there might not be any impact. I can double check this. At the same time, the mgbf nml files expect a count of the analysis variables, so differing the counts between pure envar and hybrid envar could be more difficult than need be. It might also add confusion for outside users when to use what. If it makes zero impact, would you be okay with keeping the variable list (and thus count) the same between all configurations?

This is also in line with my preference for use of an offline tool. It would be great if we can make it work consistently among all variations. If we must use offline tool for one, I would like to use offline tool for all.

TingLei-NOAA:
@delippi  Ok, I agree with the current choice (keeping ps even it won't have any direct impact on the analysis). 
Sure, further discussion is needed on which could be the unified treatment for pressure analysis (both variational analysis and enkf).

HuiLiu-NOAA:
I agree that keeping both "ps" and "delp" in the analysis list of pure ens3dvar would be better or more clear, since the "ps" is indeed one variable of the analysis output (although ps is not used in the restart file).

delippi:
I just tested `rrfs_fv3jedi_2024052700_3denvar_mgbf` ctest with and without `air_pressure_at_surface` in the analysis_variables list. At first, I forgot to change the km3=7 to 6 (after removing the variable), but it still seemed to run. I ran it again changing km3=6. The ctest passed and comparing the final analysis increments were identical in both cases. It still seems logical to me to keep it in the list to be consistent with the other variations.

HuiLiu-NOAA:
> I just tested `rrfs_fv3jedi_2024052700_3denvar_mgbf` ctest with and without `air_pressure_at_surface` in the analysis_variables list. At first, I forgot to change the km3=7 to 6 (after removing the variable), but it still seemed to run. I ran it again changing km3=6. The ctest passed and comparing the final analysis increments were identical in both cases. It still seems logical to me to keep it in the list to be consistent with the other variations.

This sounds good to me.

delippi:
@TingLei-NOAA @HuiLiu-NOAA @Masanori-NOAA is there anything that you are concerned with here?

I'm starting to lean toward the offline tool like GDAS is using, but that might take awhile longer to get going. My understanding is that tool will both 

1. do the variable transforms we need like ps -> delp
2. add increments to the background which will allow us to remove all/most of the add increments via netcdf tools that I put in the ex jedi script. It should clean that script up quite a bit I hope. 
 
Once I can get that going I can switch all these options back, but also it would be nice to have these in place to compare this vs offline method.

TingLei-NOAA:
@delippi I am still trying to run your hybrid case to make fv3-jedi does as we expected. Because I have been using jedi-bundle, hence, I needed to do some update/work-around that rdasapp does for gsibec part. 
If the fv3-jedi hybrid does work for delp and ps in a correct way,  this PR would be fine to me.

HuiLiu-NOAA:
It may be worth to do a comparison of the off-line and online options, just a double check.

delippi:
@TingLei-NOAA If you're having too much trouble, it's probably easiest to do these tests by cloning rrfs-workflow [dev-sci] and running the simplified workflow. If you start building that today, we can have a one-on-one and I can walk you through it tomorrow to expedite your tests.

@HuiLiu-NOAA Yes, I think it would be good to do that comparison with the offline tool too. The only thing is I'm not sure how long it will take to get that working. It will take even longer to build that into RDASApp because it will require moving that code from GDASApp into da-utils and then adding da-utils as a submodule etc. This PR seems good, at least for 3DVar.

@ShunLiu-NOAA do you want to wait until I can get a simple test with the offline tool? Or until Ting has a chance to look into the hybrid variation in more detail?

Another option is I can edit this PR to turn off the delp as analysis variable by just commenting out those parts? That way it is there ready to be used. My leaning is toward this option as I think we ought to use the offline tool going forward.

Masanori-NOAA:
No concerns from my side. Thanks again for your great work.

ShunLiu-NOAA:
@delippi Thank you for listing these options. Could you share the path to your test directory on WCOSS2. @TingLei-NOAA, let's check Donnie's results first.

TingLei-NOAA:
@ShunLiu-NOAA  The problem is the impact of  missing adjoint as found by @Masanori-NOAA and @HuiLiu-NOAA is not easy for me to identify from the results. That is why i need to trace the exact variable change steps in the hybrid runs. 
@delippi  I guess I would ask for your help later on runing the workflow for 3km NA runs. But for this hybrid run, I am running it to confirm/clarify on the above steps .  I hope  this pending PR is not blocking any works you are planning on.

SamuelDegelia-NOAA:
PASSED on wcoss2

started build_and_test on wcoss2 at UTC time: Fri Jan 16 04:37:48 UTC 2026
finished at UTC time: Fri Jan 16 05:37:08 UTC 2026
```
Test project /lfs/h2/emc/da/noscrub/samuel.degelia/rrfsbot/PRs_RDASApp/520/build/rrfs-test
      Start  6: rrfs_fv3jedi_2024052700_getkf_observer
      Start 15: rrfs_mpasjedi_2024052700_getkf_observer
      Start  1: rrfs_fv3jedi_2024052700_3dvar
      Start  2: rrfs_fv3jedi_2024052700_3denvar
      Start  3: rrfs_fv3jedi_2024052700_3denvar_mgbf
      Start  4: rrfs_fv3jedi_2024052700_hybrid3denvar
      Start  5: rrfs_fv3jedi_2024052700_hybrid3denvar_mgbf
      Start  8: rrfs_fv3jedi_2024052700_3dvar_conv_surface
      Start  9: rrfs_fv3jedi_2024052700_3dvar_conv_upperair
      Start 10: rrfs_fv3jedi_2024052700_3dvar_remote
 1/17 Test #10: rrfs_fv3jedi_2024052700_3dvar_remote ..........   Passed   72.99 sec
      Start 11: rrfs_fv3jedi_2024052700_3dvar_satrad
 2/17 Test  #1: rrfs_fv3jedi_2024052700_3dvar .................   Passed   86.15 sec
      Start 12: rrfs_fv3jedi_2024052700_3denvar_refl
 3/17 Test  #9: rrfs_fv3jedi_2024052700_3dvar_conv_upperair ...   Passed   94.98 sec
      Start 13: rrfs_mpasjedi_2024052700_bumploc
 4/17 Test  #8: rrfs_fv3jedi_2024052700_3dvar_conv_surface ....   Passed  103.98 sec
      Start 14: rrfs_mpasjedi_2024052700_3denvar
 5/17 Test  #6: rrfs_fv3jedi_2024052700_getkf_observer ........   Passed  160.03 sec
      Start  7: rrfs_fv3jedi_2024052700_getkf_solver
 6/17 Test #11: rrfs_fv3jedi_2024052700_3dvar_satrad ..........   Passed  136.00 sec
      Start 17: rrfs_bufr2ioda_msonet
 7/17 Test #17: rrfs_bufr2ioda_msonet .........................   Passed   33.16 sec
 8/17 Test  #2: rrfs_fv3jedi_2024052700_3denvar ...............   Passed  260.00 sec
 9/17 Test  #4: rrfs_fv3jedi_2024052700_hybrid3denvar .........   Passed  262.01 sec
10/17 Test  #7: rrfs_fv3jedi_2024052700_getkf_solver ..........   Passed  180.02 sec
11/17 Test  #3: rrfs_fv3jedi_2024052700_3denvar_mgbf ..........   Passed  411.10 sec
12/17 Test  #5: rrfs_fv3jedi_2024052700_hybrid3denvar_mgbf ....   Passed  433.00 sec
13/17 Test #13: rrfs_mpasjedi_2024052700_bumploc ..............   Passed  349.01 sec
14/17 Test #15: rrfs_mpasjedi_2024052700_getkf_observer .......   Passed  459.04 sec
      Start 16: rrfs_mpasjedi_2024052700_getkf_solver
15/17 Test #14: rrfs_mpasjedi_2024052700_3denvar ..............   Passed  472.00 sec
16/17 Test #16: rrfs_mpasjedi_2024052700_getkf_solver .........   Passed  331.00 sec
17/17 Test #12: rrfs_fv3jedi_2024052700_3denvar_refl ..........   Passed  1141.89 sec

100% tests passed, 0 tests failed out of 17

Label Time Summary:
rdas-bundle    = 4986.37 sec*proc (17 tests)
script         = 4986.37 sec*proc (17 tests)

Total Test time (real) = 1228.06 sec
```
workdir: /lfs/h2/emc/da/noscrub/samuel.degelia/rrfsbot/PRs_RDASApp/520

delippi:
I just tested the new offline python tool `offline_compute_delp_inc.py` with some NA3km output. It seems to produce reasonable results. It was pretty quick too (~10s).

delippi:
I just updated the name and description of this PR. I compared the output of the offline tool against the 3dvar online method. The results were basically identical (rounding error e-8 in a few spots). I also compared it against the hybrid online method and those were very different as others had suggested that it might be.

I kept the settings for using delp as analysis variable, but turned off. So all the yaml changes are basically cosmetic and didn't change ctest results. I think everything is ready to go.

ShunLiu-NOAA:
@TingLei-NOAA Let me know if there are any final comments. Otherwise, I'll merge this now.

SamuelDegelia-NOAA (approved):

TingLei-NOAA:
@delippi   Great there is an off-line too now. 
 What did you mean by "I kept the settings for using delp as analysis variable, but turned off."

delippi:
> @delippi Great there is an off-line too now. What did you mean by "I kept the settings for using delp as analysis variable, but turned off."

Sorry, that was unclear. I meant the changes like:

```diff
+      active variables: {{control_variables}}
...
+ #- air_pressure_thickness
```
So I have air_pressure_thickness in the list but commented out ("#") and I have the active variables set. Basically all the things we would need to turn it back on except for the mgbf fix files km3 settings.

TingLei-NOAA (approved):
@delippi  Thanks. That is great. For clarification/confirmation of what you added.   Now we stick to the strategy as the global side is doing as we discussed . Use only surface pressure as the analysis variable and do off-line update of delp. 
The update method is the same as in GSI.

Issue #496, opened 2025-11-26
Time-window logic between GSI and JEDI is inconsistent for sondes


delippi:
## Description
For sonde observations, GSI applies special logic in `read_prepbufr.F90` that overrides the twindow specified in the convinfo table. Even when twindow = 1.5 hours (+/-5400 s), the 01z and 13z cycles will force late/early sondes into the window by resetting their time_drift to -1 hour and assimilate the full profile.

In practice, this means:
- All sonde obs older than the convinfo window are shifted to -3600 s
- They are still assimilated, even though they fall outside the nominal +/-5400 s
- A large spike appears at -3600 s in the GSI time distribution
- The same sonde profile is (likely) assimilated in two consecutive cycles (00z + 01z, and 12z + 13z)

JEDI does not implement this fallback logic, and simply applies the user-defined time window via MetaData/timeOffset. Thus, JEDI keeps only the obs that truly fall within +/-twindow, while GSI forcibly pulls older profiles back in.

### Long-term question(s):
- Should RRFSv1 GSI be updated to respect the convinfo time window without forcing late/early soundings back in?
- Should JEDI eventually replicate this behavior, or should we align both systems to the simpler and more transparent JEDI time window?

delippi:
This is the relevant code snippet:
https://github.com/NOAA-EMC/GSI/blob/13b754cc5edbf3fd276a75d9de91c9f354a534fc/src/gsi/read_prepbufr.F90#L1970-L1972
```
                    if(abs(time_drift) > ctwind(nc) .or. abs(time_drift) > twindin)then
                       time_drift=timex
                       if(abs(timex) > ctwind(nc) .or. abs(timex) > twindin) CYCLE LOOP_K_LEVS
```

This checks if the `time_drift` is outside the bounds set in convinfo table and then resets it to `timex` which is -1 h. This is why we see a huge spike at -1 h (-3600 s) in the figure:

<img width="1800" height="1080" alt="Image" src="https://github.com/user-attachments/assets/5b5d9817-aeaf-499b-8974-3494aff93b6d" />

I tested changing the code to: 
```diff
                    if(abs(time_drift) > ctwind(nc) .or. abs(time_drift) > twindin)then
-                       time_drift=timex
+                      time_drift=time_drift
-                       if(abs(timex) > ctwind(nc) .or. abs(timex) > twindin) CYCLE LOOP_K_LEVS
[...]

PR #538, opened 2026-02-18
Offline patch for sonde timeOffset

delippi:
## Description
In https://github.com/NOAA-EMC/RDASApp/issues/496, it was shown that GSI applies special logic in `read_prepbufr.F90` that overrides the twindow specified in the convinfo table for obs with balloon drift information. Essentially, GSI will attempt a fallback time (-3600s) if the original time falls outside the allowed window (5400s), and rejects the observation if the fallback still fails the new window check. This doesn't usually impact the 00/12z cycles since those are launch times. The impact from this change is usually noted in 01/13z cycles where the timeOffset can be -7500s (outside the -5400s time window) in which case, those obs gets snapped to -3600s.

This PR adds this capability to change the timeOffset of the data via the `offline_ioda_patch.py` utility. Here is an example call to use this new function:
```
./offline_ioda_patch.py -o ioda_adpupa.nc --patch-timeoffset
```
There are hard coded kx values so it is technically safe to run on any ioda file and should only change types 120/220 timeOffsets, but it will add a new metaData variable called origtimeOffset.

## Issue(s) addressed
Resolves/Results are documented in Issue https://github.com/NOAA-EMC/RDASApp/issues/496

## Checklist

### DIFF
--- a/rrfs-test/IODA/offline_ioda_patch.py
+++ b/rrfs-test/IODA/offline_ioda_patch.py
@@ -36,6 +36,13 @@ tic1 = tic()
 
 parser = argparse.ArgumentParser()
 parser.add_argument('-o', '--obs', type=str, help='ioda observation file', required=True)
+parser.add_argument('--patch-timeoffset', action='store_true',
+                    help='Patch MetaData/timeOffset for soundings (ObsType 120/220)')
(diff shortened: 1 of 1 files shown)


SamuelDegelia-NOAA (approved):
LGTM!

ShunLiu-NOAA (approved):

Issue #552, opened 2026-04-13
Bridging the Gap Between RDASApp Needs and JEDI Development


guoqing-noaa:
PR reviews and merges in RDASApp are typically completed within a few days, whereas merging those changes into the official JEDI repositories often takes several months (and in extreme cases, over a year).

To bridge this practical gap, we propose several possible approaches:

#### 1. Expansion of JCSDA Review Capacity
JCSDA could allocate additional resources to accelerate the review and integration of PRs, with a target average turnaround time of less than one month. However, given current resource constraints, the feasibility of this option appears limited.

#### 2. Make all JEDI private repositories public with dedicated NOAA branches
JCSDA could make all JEDI private repositories public, allowing us to create dedicated branches (e.g., `rrfs` or `noaa`) in needed submodules. These branches would be primarily reviewed and maintained by NOAA developers, with support from JCSDA. This would enable faster progress while upstream PRs are still pending. While doable, this approach may not be feasible in the next few months.

#### 3. Use temporary forks of some JCSDA repositories (recommended)
We can create temporary forks of relevant JCSDA repositories (e.g., `SABER`, `GSIBEC`, `MPAS-JEDI`, `FV3-JEDI`), with branches such as `rrfs` or `noaa` to support our specific needs. Changes can be developed and reviewed through standard PR review workflows within these forks. When syncing with upstream JCSDA updates, we can reapply changes via new PRs as needed. Once the relevant PRs are merged, those fork will be removed from RDASApp.

This approach appears to be the most practical in the near term. It eliminates fragile `workaround` practices and makes changes easier to review, track, and collaborate on. A similar practice has already been successfully adopted in [`RRFSx/MPAS-Model`](https://github.com/RRFSx/MPAS-Model/tree/rrfs). I can provide an example PR in RDASApp to demonstrate this process.

Method 3 reflects a standard and widely adopted industry practice for working with large, slower-moving upstream projects that typically have more rigorous review processes. Representative examples include the `Android` Open Source Project, `Red Hat` Enterprise Linux, `React` developed by Meta Platforms, and the `Chromium` / `Chrome` ecosystem developed by Google, etc.

ShunLiu-NOAA:
@guoqing-noaa Thank you for bringing this inquiry to the RDAS team. I will coordinate with DA management to ensure this is addressed properly.

guoqing-noaa:
Sorry that I under-communicated the method 3 or 2 listed here and hence it might cause confusions.

[...]

PR #555, opened 2026-04-15
use the **`temporary fork`**  method to bridge the gap between RDASApp needs and JCSDA development

guoqing-noaa:
## Description
This PR adopts the method 3 as discussed in Issue #552, i.e. the fork method, to bridge the gap between RDASApp needs and JCSDA development.

Five repos are forked at RRFSx and the associated PRs are at:
| Pull Request | Description |
|--------------|-------------|
| https://github.com/RRFSx/GSIbec/pull/1 | patch for handling missing values |
| https://github.com/RRFSx/saber/pull/1 | patch for GSIBEC |
| https://github.com/RRFSx/saber/pull/2 | add MGBF |
| https://github.com/RRFSx/ufo/pull/1 | patch for GSL surface DA |
| https://github.com/RRFSx/mpas-jedi/pull/1 | update 2m t,q and 10m u,v between outer loops |
| https://github.com/RRFSx/fv3-jedi/pull/1 | fv3 parallel IO |

All changes are copied from either the `sorc/_workaround_` directory or the pending related PRs.

In this process, I found some files were out of sync with the upstream changes, such as saber had included `torchbalance` but the related files under mgbf have not reflected this change yet.
Also, @rtodling has replaced SP with IP in https://github.com/JCSDA-internal/saber/pull/1194 and https://github.com/JCSDA-internal/fv3-jedi/pull/1460, but we have not updated them accordingly.

This demonstrated that it is critical to get each POC to review each patch through respective PR whenever we do a RDASApp submodule sync.

This will enable us safely sunset the current fragile `_workaround_` practices.

## Issue(s) addressed
Resolves/Results are documented in Issue #552 

none

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -15,7 +15,7 @@
 	ignore = all
 [submodule "sorc/fv3-jedi"]
 	path = sorc/fv3-jedi
-	url = https://github.com/jcsda/fv3-jedi
+	url = https://github.com/RRFSx/fv3-jedi
 	branch = develop
 	ignore = all
 [submodule "sorc/fv3-jedi-km"]
@@ -25,7 +25,7 @@
 	ignore = all
 [submodule "sorc/gsibec"]
 	path = sorc/gsibec
-	url = https://github.com/GEOS-ESM/GSIbec
+	url = https://github.com/RRFSx/GSIbec
 	branch = develop
 	ignore = all
 [submodule "sorc/ioda"]
@@ -50,12 +50,12 @@
 	ignore = all
 [submodule "sorc/saber"]
 	path = sorc/saber
-	url = https://github.com/jcsda/saber
+	url = https://github.com/RRFSx/saber
 	branch = develop
 	ignore = all
 [submodule "sorc/ufo"]
 	path = sorc/ufo
-	url = https://github.com/jcsda/ufo
+	url = https://github.com/RRFSx/ufo
 	branch = develop
 	ignore = all
 [submodule "sorc/vader"]
@@ -70,7 +70,7 @@
 	ignore = all
 [submodule "sorc/mpas-jedi"]
 	path = sorc/mpas-jedi
-	url = https://github.com/JCSDA/mpas-jedi
+	url = https://github.com/RRFSx/mpas-jedi
 	branch = develop
 	ignore = all
 [submodule "sorc/bufr-query"]
--- a/build.sh
+++ b/build.sh
@@ -263,7 +263,7 @@ if [[ $BUILD_RRFS_TEST == 'YES' ]]; then
 fi
 
 # Copy workaround codes (remove these as soon as PRs are merged)
-if [[ $BUILD_WORKAROUND == 'YES' ]]; then
+if [[ $BUILD_WORKAROUND == 'COPY_WORKAROUND' ]]; then
 
   # Apply workaround patches (git commit-base workaround method)
   # Need to convert all the copy-based workaround methods to commit-based method


guoqing-noaa:
https://github.com/RRFSx/saber/pull/2 and https://github.com/RRFSx/fv3-jedi/pull/1 are still pending.
It will be good that we have an agreement on the replacement of SP with IP.
Also, it will be good thta @TingLei-NOAA could do a review of https://github.com/RRFSx/saber/pull/2. Thanks!

guoqing-noaa:
@Junjun-NOAA FYI, MPAS-JEDI will inherit the use of `SP` in the MGBF workaround instead of the `IP` in the latest SABER if we don't update further:
https://github.com/RRFSx/saber/pull/2#discussion_r3084653526
However, not sure whether this will affect the results or not.

hu5970:
@guoqing-noaa Let's discuss this method further before to take actions.

guoqing-noaa:
> @guoqing-noaa Let's discuss this method further before to take actions.

@hu5970 Sure, you guys take time to think about this and make the decisions. And this is a Draft PR.

I would like to bring up that the current `_workaround_` practice is too risky. I demonstrated in the above that just "copying old codes around" may lost sync with upstream easily and silently (such as the above sp vs ip change). This may bring unexpected results to MPAS-JEDI without awareness.

I strongly suggest we go through the regular rigorous PR review process for each patch instead of just "copy codes around", especially now we have so many patches. Thanks!

EricJames-NOAA:
> > @guoqing-noaa Let's discuss this method further before to take actions.
> 
> @hu5970 Sure, you guys take time to think about this and make the decisions. And this is a Draft PR.
> 
> I would like to bring up that the current `_workaround_` practice is too risky. I demonstrated in the above that just "copying old codes around" may lost sync with upstream easily and silently (such as the above sp vs ip change). This may bring unexpected results to MPAS-JEDI without awareness.
> 
> I strongly suggest we go through the regular rigorous PR review process for each patch instead of just "copy codes around", especially now we have so many patches. Thanks!

@guoqing-noaa I agree with all your comments...I think this method is the best approach moving forward.  So in this new paradigm, we would no longer have _workaround_ files, but instead a complete fork of each repository that needs to be modified for RRFSv2 applications.  I guess in this case, syncing submodules would become a process of adding our workaround modifications on top of the most recent authoritative code, right?

guoqing-noaa:
> > > @guoqing-noaa Let's discuss this method further before to take actions.
> > 
> > 
> > @hu5970 Sure, you guys take time to think about this and make the decisions. And this is a Draft PR.
> > I would like to bring up that the current `_workaround_` practice is too risky. I demonstrated in the above that just "copying old codes around" may lost sync with upstream easily and silently (such as the above sp vs ip change). This may bring unexpected results to MPAS-JEDI without awareness.
> > I strongly suggest we go through the regular rigorous PR review process for each patch instead of just "copy codes around", especially now we have so many patches. Thanks!
> 
> @guoqing-noaa I agree with all your comments...I think this method is the best approach moving forward. So in this new paradigm, we would no longer have _workaround_ files, but instead a complete fork of each repository that needs to be modified for RRFSv2 applications. I guess in this case, syncing submodules would become a process of adding our workaround modifications on top of the most recent authoritative code, right?

@EricJames-NOAA Yes and thanks for the input!

Since we have many patches now and are not modifying one or two lines, one or two files. So it is important to get the author or POC of each patch to explicitly review it again in the new context (if that patch has not been merged in official JEDI yet) whenever we do submodule sync. This will help avoid any silent risks.

guoqing-noaa:
Copy the following comments from the associated issue:

Sorry that I under-communicated the method 3 and hence it might cause confusions.

What I mean is that we created **`temporary`** forks/branches to hold our pending changes so that we may continue to run realtime and retro experiments while the upstream PRs are still under review.

Once those PRs are merged, we will immediately remove those temporary fork/branches from RDASApp and point to the official repos again. 

The lifetime of those temporary forks/branches depends on the review speed. Some may be just a few weeks, some may be a few months. These **`temporary`** forks/branches also remind us to actively push pending PRs move forward.

We would expect that most time, all JEDI submodules point to the official ones and in the late almost mature stage, we will have no need of those temporary forks/branches.

ShunLiu-NOAA:
@guoqing-noaa Thank you for the update. Let's continue this discussion during tomorrow's RRFS-DA meeting.

guoqing-noaa:
The "commit-based" patching brings an unnecessary extra effort when working on the development of JEDI submodules.
For example, I need to modify `sorc/mpas-jedi/src/mpasjedi/State/mpas_state_mod.F90`,  but now I cannot make edits directly.
I have to run `build.rdas` first and then manually stop it once the cmake step is done. This is to get the patching applied first.
Then I can make needed edits. After that, I go back to do the build.

guoqing-noaa:
close as it is outdated and only serves as an example.

